PR #1132, opened 2014-09-25
Log 404s

mariansollmann:
To easily see changes in volume of 404 errors.

Sometimes we have 404 errors which actually are programming bugs (eg `CM_Exception_Nonexistent` is thrown because some model cannot be loaded and accordingly a 404 page is rendered). It would be good to log/track them so we can monitor these and inspect if necessary.


mariansollmann:
@njam I changed everything from your review. But now the exception get catch twice because first is "not existed" and secound is "invalide param". any suggest how to prevent this?

mariansollmann:
@alexispeter maybe you can answer this as well

njam:
- CM_Http_Response_Page: You can use `CM_ExceptionHandling_Handler_Abstract::logException()` instead of manually formatting and adding exceptions to the log
- CM_Http_Response_Page: To avoid logging an entry for each configured exception we should match the exception's class to the config key! You can probably combine this with the existing check: `array_key_exists(get_class($e), $this->_getConfig()->catch)`?
- CM_Paging_Log_NotFound: Formatting

njam:
Let's also add a test for the ajax navigation case (instead of normal page load). Have a look at: `CM_Http_Response_View_AbstractTest::testLoadPage`.

alexispeter:
Just remove the foreach and use `$catch[get_class($e)]`

mariansollmann:
> CM_Http_Response_Page: You can use CM_ExceptionHandling_Handler_Abstract::logException() instead of manually formatting and adding exceptions to the log

@njam how does it work if I want to add the exception to the `NotFound` log. If I use `CM_Bootloader::getInstance()->getExceptionHandler()->logException($e);` it will add it to the `Error` Log because of `Severity=>error`

@alexispeter in the new `testNotFoundErrorLogging` the response get redirect to the error page and then needs the `default` layout. Not sure if to add the class `CM_Layout_Default` to the test is the right approach. 
please review.

njam:
- CM/Http/Response/Page.php: Make $path assignement up to where the $options are read?
- CM/Http/Response/View/AbstractTest.php: Redefining `CM_Layout_Default` is a bit dangerous (what if both are loaded?).  I worked around this in another PR by setting a different "not-found" page (a mock one, with a custom layout): [example](https://github.com/njam/CM/blob/ga/tests/library/CM/Http/Response/View/AbstractTest.php#L83)

mariansollmann:
@njam please review.
what about my question in https://github.com/cargomedia/CM/pull/1132#issuecomment-72012608

njam:
Yup, I write an answer to that but it got lost :/ You're right, adding to the log directly is the best way then!

njam:
Changes looks good to me

mariansollmann:
@njam please have a look on the last commits. thx

njam:
- CM/Http/Response/Page.php: Inline $path?

mariansollmann:
@njam please last review.

njam:
lgtm

PR #1502, opened 2014-11-30
FB-295 Make CM_App central application instance

tomaszdurka:


njam:
ticket is closed, so closing review

PR #1557, opened 2014-12-16
Simplify resource Responses and Assets

tomaszdurka:


tomaszdurka:
As discussed with @njam I won't overcomplicate it too much.

One of the problem which arised is having several problems solved by `CM_Asset_Css*`. It could be separated into 3 separate concepts - `CM_Css`, `CM_Asset_Css_Vendor/Library` and `CM_Frontend_ViewCss`...

njam:
@tomaszdurka what about this? Some tickets depend on it now. Can we finish the not-overcomplicated version quickly?

njam:
@tomaszdurka I was looking at this with Alexis. The changes on the branch look reasonable. But what's discussed on the ticket here is something else. Not sure what's the importance of this. What do you think - how should we proceed with this?

njam:
Closing for now.
@tomaszdurka if you have thoughts on this one, please share and reopen

PR #1560, opened 2014-12-16
Nonblocking Clockwork execution

### LINKED ISSUE #1220: Maintenance: Long-running tasks shouldn't delay other tasks
Ideas:
- Fork when running task in Clockwork manager (but making sure they don't run twice?)
- Fork one clockwork per maintenance task?

cc @tomaszdurka

alexispeter:
closes #1220


alexispeter:
please review @njam

alexispeter:
After discussing this topic again with @njam, we decided to use forking for all events instead of using gearman for some.

fyi @tomaszdurka

njam:
- I'm not fully convinced of passing the `$pid` to the `CM_Process_WorkloadResult`. I think it would be better to keep the *workload-result* as simple as possible, and introduce this dependency between the pid (from the *fork-handler*) and the *workload-result*.
 - I understand we need a mechanism to identify a forked process from `fork()` across `listenForChildren()`. We already have such a mechanism - the *fork-handler-sequence* which is unique per `fork()` call. It will be passed back as the key of the *workload-result* array. Question remains how to retrieve this `$sequence`; how about returning it in `fork()` instead of the `$pid` (which is never used)? Alternatively we could allow to pass a second argument as a reference to `fork()` which would be set to the sequence?
- CM_Clockwork_Manager::_handleWorkloadResult: Docu

alexispeter:
please review @njam

njam:
- CM_Clockwork_Manager::_handleResult: Docu
- CM_Clockwork_Manager::_handleResult: How about rather introducing `_getRunningEvent($identifier)` and then inlining the rest of the logic into `runEvents()`?
- **General**: How about now instead of using `CM_Clockwork_Manager::_eventsRunning` directly use `CM_Process::_forkHandlerList`? I.e. `CM_Process::getForkHandler($sequence)`?

alexispeter:
> General: How about now instead of using CM_Clockwork_Manager::_eventsRunning directly use CM_Process::_forkHandlerList? I.e. CM_Process::getForkHandler($sequence)?

Cannot be done, events need to be linked somehow to the forkhandler/result.

alexispeter:
please review @njam

njam:
- `CM_Clockwork_ManagerTest::testRunEvents` failed on php 5.5
- Streamline `$identifier` docu - always `int`?

alexispeter:
please review @njam

njam:
- CM/Clockwork/Manager.php: Formatting

Do you know why the tests failed? Is it flaky? Can it be made more robust?

alexispeter:
There is a slight execution time dependence for forked processes (like we have in the CM_Process tests). 
The test will fail if the fork takes too long to execute.
We can either increase the time the parent process wait for the fork, or remove the test altogether.  testRunNonBlocking() does basically the same as testRunEvents() with the difference that forking functionality is mocked.

alexispeter:
please review @njam

PR #1564, opened 2014-12-18
Introduce a flag 'avoidPasswordManager' for forms.

vogdb:


vogdb:
@fauvel please review.

njam:
- CM/Form/Abstract.php: Add var-docu for `$_avoidPasswordManager`
- CM/Form/Abstract.php::isAvoidPasswordManager: Rename to `getX` (methods [should begin with verb](https://github.com/cargomedia/guides/tree/master/style#naming))
- CM/Form/Abstract.php::getAvoidPasswordManager: Add type casting
- block.form.php: Add the html with two concatenations, so we don't have unnecessary whitespace in the output?
- block.form.php: Don't add `/` when closing tag (just `<input>`)

vogdb:
@njam please review.

njam:
- Add type-casting to setter, but not getter
- block.form.php: Apply formatting - How about writing this in two lines?
```php
            $html .= '<input style="display:none" type="text" name="fakeusernameremembered">';
            $html .= '<input style="display:none" type="password" name="fakepasswordremembered">';
```

vogdb:
Yes! @njam please review.

njam:
Will these two fields affect `CM_Form_Abstract.getData()` (that's sent to the server)?
If yes, could we make them `disabled`?

vogdb:
> Will these two fields affect CM_Form_Abstract.getData() (that's sent to the server)?

No because they are not registered in the form. Should I make them disabled on a fire case?

njam:
> No because they are not registered in the form.

True

> Should I make them disabled on a fire case?

Can we make them disabled from the start? Or does it prevent the workaround to work?

vogdb:
> Can we make them disabled from the start? Or does it prevent the workaround to work?

It prevents the workaround to work =(

njam:
Ok - lgtm!

Issue #1476, opened 2014-11-19
Remove unneeded javascript polyfills


njam:
Now that we're [not supporting IE9 anymore](https://github.com/cargomedia/CM/pull/1474) there might be other polyfills that are not needed anymore.

Please check all javascript polyfills and compare with caniuse.com and [supported browsers](https://github.com/cargomedia/sk/wiki/Supported-Browsers)

vogdb:
All of the libraries are finally updated.

PR #1565, opened 2014-12-29
Update jquery.cookie.js from 1.1 to 1.4.1.

vogdb:
The child of #1476.

- jquery.cookie. from 1.1 (Jul 2012) to 1.4.1 (Apr 2014). Changelog is here: https://github.com/carhartl/jquery-cookie/blob/master/CHANGELOG.md. Quick research showed that we don't rely on these changes but still it'd be better to make this update thoroughly.
  - 1.4.0. `$.cookie('name')` now returns `undefined` in case such cookie does not exist (was `null`). Because the return value is still falsy, testing for existence of a cookie like `if ( $.cookie('foo') )` keeps working without change.
  - 1.4.0. Badly encoded cookies no longer throw exception upon reading but do return `undefined` (similar to how we handle JSON parse errors with `json = true`).
  - 1.4.0. Dropped `$.cookie('name', null)`. Replaced with `$.removeCookie('name')`.

### DIFF
--- a/client-vendor/after-body/jquery.cookie/jquery.cookie.js
+++ b/client-vendor/after-body/jquery.cookie/jquery.cookie.js
@@ -1,42 +1,71 @@
-/*jshint eqnull:true */
 /*!
- * jQuery Cookie Plugin v1.1
+ * jQuery Cookie Plugin v1.4.1
  * https://github.com/carhartl/jquery-cookie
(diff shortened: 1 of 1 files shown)


vogdb:

christopheschwyzer:
lgtm (afaict)

PR #1573, opened 2014-12-31
Change emoticons setup.

vogdb:


njam:

alexispeter:
code looks ok

njam:
Should we use an approach without db tables instead? (see https://github.com/cargomedia/sk/pull/1010#issuecomment-68594442)

alexispeter:
I think so.

vogdb:
@alexispeter @njam I need a small guidance.
- What should I use instead of `$output->writeln`
- Is it ok to move `_checkEmoticonValidity` inside `CM_Paging_Emoticon_All` and use it only after `parent::__construct($source);`? I did it with a purpose of max preserving the existing source code.

njam:
When moving this code to a paging it will not run in an environment with terminal output anymore. So writing to standard out is not a good idea anymore.

Instead I would suggest to throw an exception.
Agree, we should still run `_checkEmoticonValidity`. Maybe run it when we load the data from the filesystem.

@alexispeter thoughts?

alexispeter:
It might be a smart idea to introduce a simple class to represent emoticons. Would also give easy access to configuration.

Constructor would take the original code as argument and load additional codes and images. 
Factory method could take original and additional codes as argument, return the appropriate object using data stored in the config and report if the specified code has duplicates.

vogdb:
> Constructor would take the original code as argument and load additional codes and images.

How it can load additional codes and images while Paging holds all the info? Paging scans the filesystem and takes all the info. After that it creates a list of CM_Emoticon. Shouldn't a constructor accept all the arguments at ones like this:
```php
  __construct($code, $file, $additionalCode) {...}
```
or like this:
```php
  __construct($dataArray) {...}
```
?

alexispeter:
I would prefer it if CM_Emoticon was responsible for loading data. The situation would be similar to models, where pagings load ids and the model class loads data associated with this id.

vogdb:
Ok. Should CM_Emoticon locate config file by code on its own or accept it
as a constructor argument?

njam:
One concern - with removing the mysql table we're moving some heavy, resource intensive code from the *setup script* to normal *business logic code*.
We need to be careful because we shouldn't run that heavy code too often, and also we should avoid it throwing lots of exceptions (due to validation?). We want to accomplish that by caching all the data in memcache.
This approach has one downside: we **restart and flush memcache** sometimes while the application is running. Usually what we do in other cases is we cache the data in a file cache (e.g. for class-children cache, composer modules cache, css cache etc).
Do you think the filesystem based data loading approach is fast enough so that we can live with cache invalidation even during normal application lifetime? Or do we need an additional database storage?

alexispeter:
I think it should be ok.

vogdb:
Closed in favor of #1593.

PR #1577, opened 2015-01-06
Pass URL parameters when redirecting from a localized page

fauvel:
Fixes page not found issues due to missing parameters after a redirect from a page with a localization specified in the URL. Mainly affects users who are logged in and navigate through external links (i.e. via search engines).


fauvel:
@njam please review

njam:
Hmm I don't think "decoded" is the right thing to use here.
Maybe @vogdb can enlighten us about this.
Or you could try passing an instance of a "location" as a parameter (and add it to the unit test)?

fauvel:
It should be right code-wise, but I'm adding a test :)

fauvel:
@njam it's working fine, please review!

njam:
Ah true .. `CM_Util::link` encodes the parameters again.
But wouldn't it be *safer* to use `getParamsEncoded()` here - this way we would skip the *decoding* and *encoding* of parameters just for the redirect (which can lead to problems if the object doesn't exist)?

fauvel:
Actually it doesn't make a big deal of a difference – the user experience would be exactly the same. If a non existing model is being used as a parameter, you normally get an embedded 404 page, and in case of a language redirect, you are being redirected to `/error/not-found` (depending on the application config), which is the exactly same page.

But I agree we actually just want to pop the language part out of the URL, and I'm not sure why the logic is so complex here, and why we are first instantiating the page object. I think we could handle this redirect before, I'll try it out.

Also the same piece of code is being used for embedded pages, which should probably break if the request happened to have a language part (which is not the case normally).

njam:
I think instantiating the `$page` is okay here - we use its logic, plus instantiation is very cheap.

I agree it doesn't make much of a difference whether we use `getParamsEncoded` or `getParamsDecoded`. I'd go with *encoded* just because it's less code involved and simpler. But both is okay with me.

fauvel:
Ah ok, there's a bit of automagic in `CM_Params::encode()` I wasn't aware of :) Please review!

vogdb:
I'm kind of late here but you've got the point of `CM_Params`. `encode` is fine.

njam:
lgtm then

fauvel:
I've moved the language redirect logic on top of the page loop processing (it doesn't make sense anyway to handle it for every page in the loop). This should also solve our GA event tracking issue. @njam wdyt?

njam:
- _process: `preprocessPageResponse` can potentially also set `redirectUrl`. So I think that `if (!$this->getRedirectUrl())` should stay (instead of the `else`)?
- _process: I don't think `setPathParts(getPathParts())` should be necessary, because if the language is popped, it will be already removed from the `_pathParts` of that request?

General: A side-effect will be that the language will now not be popped anymore in "embed" page renderings, right?
I would rather suggest to keep it where it is (or move it to `_processPageLoop`?). What about introducing this `postprocessPageResponse()` - IMO this would solve the GA problem in a clearer way - would it work?

fauvel:
- Ok for keeping `if (!$this->getRedirectUrl())` – I've made an `elseif` out of it to make the control flow more explicit.
- I think `setPathParts()` is needed, because `_path` is not being updated by `popPathPart()`.
- I agree there is a side-effect, but it is almost unnoticeable and quite difficult to reproduce:
  - To reproduce it, you have to load a page with a specific language while you are logged out, then log in from another tab, and finally, in the initial tab, navigate to a new page via the ajax navigation.
  - The side-effect is that the URL in the browser navigation bar (and in the navigation history) still has the language part in it. The page itself is rendered exactly the same (i.e. the links it contains don't have any language part), so that the language part will be removed from the URL in the navigation bar as soon as you navigate further.

fauvel:
I'm not fond of keeping the language redirect in the page processing loop, because as I already said, this is a one-time operation that has nothing to do with the page processing in itself – exactly like the host redirect.

fauvel:
I've implemented the language redirection on `CM_Http_Response_Page_Embed::_process()` as well, I think this is the best solution. @njam please review!

njam:
- _process: setPathParts(): `popPathPart` calls `setPathParts` which sets both `_pathParts` and `_path`? - the idea is that all these "state" values on the request should always be consistent (except for the `_uri` which is always the original value)
 - But `_query` does *not* get set - then doesn't the generated link contain the orignal query?
- _process: I found the previous `if (!$this->getRedirectUrl()) {` clearer than the `elseif`. It's correct that control flow will never reach an earlier `if` *and* this block - but I think the two parts are quite independent are easier to read if kept apart also in code?
 - If done, then this code would become compatible with `CM_Http_Response_Page_Embed::_process`. How about introducing a helper function on `CM_Http_Response_Page` and use it in both places?

fauvel:
@njam I've introduced `CM_Http_Response_Page::_processPageEmbed()` – naming suggestions welcome :)

This enables host redirects for embed pages, so that I had to make sure we have a `host` header in our unit tests (small adjustments made to the test helpers). I've added unit test as well as "integration tests" covering possible endless loops in `CM_Http_Response_View_Abstract::loadPage()`.

Wdyt?

njam:
I think it's good now.

- createRequest: breaking change (just for the records)
- _processPageEmbed: How about `_processContentOrRedirect`
 - I would also suggest to move `_processPageLoop` down, inbetween `_processContentOrRedirect` and `_processPage` for readability - wdyt?

fauvel:
@njam could you please review a last time?

njam:
lgtm!

fauvel:
Tagged 1.92.0

PR #1578, opened 2015-01-06
gearman remove non-blocking comment

alexispeter:


alexispeter:
Bug is fixed but, as discussed with @njam, graceful shutdown of worker processes is not desired due to the way we deploy the app.

please review @njam

njam:
Looks good.

BTW here some good notes on the topic: http://dpb587.me/blog/2013/01/14/terminating-gearman-workers-in-php.html

PR #1575, opened 2015-01-08
Add link tag: hreflang for language and regional URLs

njam:
See https://support.google.com/webmasters/answer/189077

Meta information for Google to understand that a page has multiple versions in different languages. E.g.:
```html
<link rel="alternate" href="http://example.com/en" hreflang="en" />
```

cc @fauvel @moby86 @NicolasSchmutz


njam:
@fauvel please review

As according to Google we should always render all available *hreflang link-tags*, including the current version's I've introduced a new render instance `$renderDefault` which represents the "global" view of the current site (without viewer, language etc).
With this render instance I'm then rendering all the different language URLs.

fauvel:
Great, let's see whether it improves our page ranking! The code looks good to me.

PR #1579, opened 2015-01-08
Introduce CM_File_Image::freeMemory

njam:
How about introducing a `CM_File_Image::freeMemory()`?
Then we'd call this instead of the quirky `clone` approach which is not as straight-forward. Also we could potentially use it in different places.

@alexispeter please review


alexispeter:
good idea, lgtm

njam:
@alexispeter can you change the `clone` to `freeMemory()` in SK accordingly?

PR #1582, opened 2015-01-08
Remove usertext IE9 word-break hack

christopheschwyzer:


christopheschwyzer:
@mariansollmann @njam what was `oneline` intended to do anyways? I think `oneline` is a bit misleading:
```php
 /**
 * Supported modes:
 * =====================================================
 * escape = escape, remove badwords
 * oneline = escape, remove badwords, add emoticons
 * simple = escape, remove badwords, nl2br, add emoticons
 * markdown = escape, remove badwords, create html markdown, add emoticons
 * markdownPlain = escape, remove badwords, create html markdown, strip all tags, add emoticons
 */
```

njam:
`online` means the output should span only one line (hah). It's used for headings.
Did I understand your question correctly?

christopheschwyzer:
It's just that the styles were doing the opposite ;) In your described use-case `oneline` should have `.nowrap` or something. For me it seems we mix presentation and logic. _Oneline_ mode only adds emoticons to the _escape_ mode. 
Thinking... Now I see - it means that it doesn't add line-breaks. Ok than. Still a bit confusing though.

njam:
You're right, more exactly I think it does the following:
- Characters: One line of text (without newline characters)
- Display: Can wrap onto multiple lines

So I assume the styles should stay?

christopheschwyzer:
> So I assume the styles should stay?

They are not needed. Usertext already comes with
```css
  word-wrap: break-word;
  overflow-wrap: break-word;
```
This prevents content from overflowing. I just removed this hack (necessary as the above lines didn't work in IE9) that a) was for IE9 only, b) probably wasn't supposed to affact `online` only anyways

christopheschwyzer:
@mariansollmann please review
@njam fyi

mariansollmann:
looks good

mariansollmann:
you checked all usages because of `display: inline-block;`?

christopheschwyzer:
I'm positive :)

PR #1580, opened 2015-01-09
Introduce a storage adapter for MongoDB

fauvel:
As discussed with @njam, I've introduced a `CM_Model_StorageAdapter_MongoDb`, which also allows for storing different types of models in the same collection. The use case would be to list by creation date different models extending the same abstract class. These models would share some common data fields and have their own data bag on top of that.

This is the rationale behind the introduction of `CM_Model_Abstract::getTableName()` and the storage of the class type `_type` together with the object data.


njam:
- StorageAdapter/Database.php::_getTableName: I think you can also do `$className::getTableName()`?
- CMTest_TH::getMongoDb(): Docu
- CMTest_TestCase: I'm not too fond of `assert(Not)RowMongoDb`. It's global magic that's difficult to work with if we want to change the place where the serviceManager resides. How about just inlining these usages - seems pretty straigh-forward here?

@tomaszdurka please note Sébastien is introducing a `CMTest_TH::getMongoDb()` here. Let's discuss how to handle serviceManager in tests in https://github.com/cargomedia/CM/pull/1502 !

@alexispeter I think it's best if you review the actual mongodb code!

alexispeter:
* I would restrict the internal id returned by create() to only contain the document id and not the type. Document Ids are always unique. 
* If you insert data into mongodb without specifying an id, the generated id will be a ObjectId (`MongoId` object). Mongodb uses strict comparison when looking up data, meaning that a criteria like `['_id' =>  new MongoId(<idString>)]` would work but `['_id' => <idString>] would not. Since we probably don't want to pass ObjectIds to the model constructor instead of strings, we will need to handle this either in the storage adapter or on the model level. Doing it in the adapter will restrict us to only using ObjectIds (which is ok imo), while doing it in the model could have unforeseen consequences since we generally expect ids to be strings.
* loadMultiple() Since we allow multiple types in the same collection, it might be most efficient to group types by their collections, so we perform one query per collection instead of one per type.

njam:
Just one note on `loadMultiple`: We've seen no real impact of this approach on mysql. IMO we can also just do a loop here until we see a real need for a multi-get?

fauvel:
@alexispeter please review

alexispeter:
* testSave() add test for nonexistent id? Basically what you had with `$adapter->save($type, $id1 + ['foo' => '9999'], ['foo' => 'world', 'bar' => 66]);` but since ids are scalars in mongoDb just use a nonexistent id.
* in loadMultiple() you added capability to handle multiple references to the same data in the input-array. Is there a particular use-case for that? Just curious. Anyway we should probably add the same capability to CM_Model_StorageAdapter_Database. Wdyt? At the very least, testLoadMultiple() should test this functionality.

fauvel:
@alexispeter please review

alexispeter:
lgtm

fauvel:
Tagged 1.95.0

PR #1585, opened 2015-01-12
Update jquery.noUiSlider.js from 5.0.0  to 7.0.9.

vogdb:
The child of #1476.

### DIFF
--- a/client-vendor/after-body/jquery.noUiSlider/jquery.noUiSlider.css
+++ b/client-vendor/after-body/jquery.noUiSlider/jquery.noUiSlider.css
@@ -5,13 +5,17 @@
  */
 .noUi-target,
 .noUi-target * {
-	-webkit-touch-callout: none;
-	-webkit-user-select: none;
-	-ms-touch-action: none;
-	-ms-user-select: none;
-	-moz-user-select: none;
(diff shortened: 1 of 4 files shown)


vogdb:

christopheschwyzer:
This should be tested on various devices. I will do it next Tuesday when I'm back in the office - or somebody else needs to take over.

vogdb:
I think @NicolasSchmutz will be able to test it on all devices!

NicolasSchmutz:
@vogdb I've tested it and it works very good with all devices.
Only one thing: The slider should be disabled if the location field is empty.
Like here: https://www.fuckbook.com/users/search

vogdb:
perfect!

vogdb:
Your notice made me to find a bug in CM_Field_Distance. It's kind of private. So, see an email from me.

vogdb:
Ah. It's not a bug. I understood it incorrectly!

NicolasSchmutz:
@vogdb deactivation of the slider (while the location field is empty or a country is selected) works perfect with all devices!
The dropdown of the location field is a bit faulty/slow with some devices - but thats another problem.

So this is good to go!

PR #1591, opened 2015-01-14
Remove some code smells

mariansollmann:


mariansollmann:
@njam shouldn't we use `javascript:void(0)` instead of `javascript:;`?  http://stackoverflow.com/questions/134845/href-attribute-for-javascript-links-or-javascriptvoid0

@christopheschwyzer 
- /CM/layout/default/css/form.less line 278: shouldn't @sizeSwitch be assigned with `:` instead of `=`?
- we never introduce LESS variable `@breakpointSmall` is this on purpose?

njam:
> shouldn't we use `javascript:void(0)` instead of `javascript:;` in html-tag-a-href?  http://stackoverflow.com/questions/134845/href-attribute-for-javascript-links-or-javascriptvoid0

On that stackoverflow page somebody says `javascript:;` might not work in old IE version. And some people propose to prefer `javascript:;` over `javascript:void(0)`.
I don't think we need to worry about those old IE versions. Is there another reason why you want to change it

mariansollmann:
@njam no real other reason. Just the IDE dont like `javascript:;`

njam:
True it doesn't like the "unnecessary ;". Not sure whether it would work without.
I would suggest to keep it as is to avoid problems in some browsers?

mariansollmann:
agree.

christopheschwyzer:
- in `layout/default/FormField/File/default.less` is the `font-familly` even necessary? Can be default font AFAIK.

- `form.less`: `@sizeSwitch =  @sizeCheckbox + 4px;` assignment with ":" instead "="

mariansollmann:
@njam please review.

njam:

alexispeter:
* CM_Model_StorageAdapter_Cache::_generateId() is this method needed?I can't find it anywhere in the class hierarchy, maybe legacy code? Could you investigate and remove it completely if appropriate?

rest looks good
I assume you had css checked

alexispeter:
talked about it quickly with @njam, you can remove _generateId() and throw the exception in create()

christopheschwyzer:
`less / css` is fine

mariansollmann:
@alexispeter please review.

alexispeter:
lgtm

PR #1588, opened 2015-01-14
Allow to find by abbreviation (or name) in CM_Model_Location

dlondero:
It would be great to be able to `findByAbbreviation()` countries and states as right now that could be done only by `id` and most of the times that differs from development to production.

One idea could be to add `CM_Model_Location_Abstract::findByAttribute($key, $value)` and use it directly or call it from `CM_Model_Location_Country::findByAbbreviation()`.

As countries and states are in different tables it would be good to pull `CM_Model_Abstract:: getTableName()` changes from #1580.

@njam FYI

### DIFF
--- a/library/CM/CacheConst.php
+++ b/library/CM/CacheConst.php
@@ -10,6 +10,9 @@ class CM_CacheConst {
     // _ip:X
     const Location_ByIp = 'Location.ByIp';
 
+    // _level:X_name:X_value:X
+    const Location_ByAttribute = 'Location.ByAttribute';
+
     // _class:X_userId:X
     const Page = 'Page';
 
--- a/library/CM/Model/Abstract.php
+++ b/library/CM/Model/Abstract.php
@@ -501,13 +501,6 @@ abstract class CM_Model_Abstract extends CM_Class_Abstract implements CM_Compara
     }
 
     /**
-     * @return string
-     */
-    public static function getTableName() {
-        return strtolower(get_called_class());
-    }
-
-    /**
      * @param array $idRaw
      * @return array
      */
--- a/library/CM/Model/Abstract.php
+++ b/library/CM/Model/Abstract.php
@@ -501,6 +501,13 @@ abstract class CM_Model_Abstract extends CM_Class_Abstract implements CM_Compara
     }
 
     /**
+     * @return string
+     */
+    public static function getTableName() {
+        return strtolower(get_called_class());
+    }
+
+    /**
      * @param array $idRaw
      * @return array
      */
--- a/library/CM/Model/Location.php
+++ b/library/CM/Model/Location.php
@@ -179,6 +179,61 @@ class CM_Model_Location extends CM_Model_Abstract {
     }
 
     /**
+     * @param int    $level A LEVEL_*-const
+     * @param string $fieldName
+     * @param string $fieldValue
+     * @return CM_Model_Location|null
+     * @throws CM_Exception_Invalid
+     */
+    public static function findByAttribute($level, $fieldName, $fieldValue) {
+        switch ($level) {
+            case self::LEVEL_COUNTRY:
+                $className = 'CM_Model_Location_Country';
+                $tableName = CM_Model_Location_Country::getTableName();
+                $tableSchema = CM_Model_Location_Country::getSchema();
+                break;
+            case self::LEVEL_STATE:
+                $className = 'CM_Model_Location_State';
+                $tableName = CM_Model_Location_State::getTableName();
+                $tableSchema = CM_Model_Location_State::getSchema();
+                break;
+            case self::LEVEL_CITY:
+                $className = 'CM_Model_Location_City';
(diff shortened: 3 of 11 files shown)


dlondero:
@njam @tomaszdurka was thinking to something like this. Still missing helper methods in `CM_Model_Country` and other levels where needed.

njam:
Generally I like the direction.
I don't like that we need to make `getSchema` static. I would rather not do that check and let it fail on the persistence layer.
But I would introduce `CM_Model_StorageAdapter_Database::findByData(array $data)` (with an interface `CM_Model_StorageAdapter_FindableInterface` (?)). wdyt?

dlondero:
So, just to understand better your idea, you would have `CM_Model_StorageAdapter_FindableInterface` with `findByData` and `CM_Model_StorageAdapter_Database` should implement the interface. Here `findByData` should receive a `type` and a `$data` array in order to do the select we now have in `CM_Model_Location` (but with multiple values so we can find by `name` and `abbreviation` for example).

In `CM_Model_Location` we would need to get persistence layer, get the adapter and call `findByData` to do the work...is that correct?

njam:
Exactly! (see `CM_Model_Abstract::_getPersistence()`)

dlondero:
And am I supposed to implement the interface also on `CM_Model_StorageAdapter_CacheLocal` to check against the cache if we already have a value there before querying the database? Or is it ok to create the key in `CM_Model_Location` and try to get it from the cache?

njam:
I would implement the caching in `CM_Model_Location`. Reason is that we don't want to invalidate the cache (because locations rarely change), so it's a special case.

tomaszdurka:
Hmm, but if they rarely change they would be invalidated as rarely within `CM_Model_StorageAdapter_CacheLocal`, no?

dlondero:
@njam please review

dlondero:
@tomaszdurka isn't 24 hours lifetime for CacheLocal? 

```php
$config->CM_Cache_Local->lifetime = 86400;
```

Hitting db once a day I guess is okay.

njam:
> Hmm, but if they rarely change they would be invalidated as rarely within CM_Model_StorageAdapter_CacheLocal, no?

@tomaszdurka yes, but cache invalidation would be *very* difficult. How to find out which caches contain a certain model? We would need to have a reverse lookup map of which caches exist?

tomaszdurka:
Ah so you mean opposite. That we actually cannot easily invalidate those caches via `CM_Model_StorageAdapter_CacheLocal` and it is easier to invalidate them within `CM_Model_Location_Abstract`. I agree then.

njam:
Oh I only now see that Location is not using the model-persistence :| Let's use the workaround you proposed for now.

- CM_Model_StorageAdapter_FindableInterface::findByData: While it will make the code a bit more complex I think this function should return an array (which is the internal ID-representation). In most cases we this looks like `['id => 12]` and we just use the `12` as the ID. But there are exceptions (like the location (which is not using the persistence though))
- CM_Model_StorageAdapter_Database: I would drop the `ORDER BY`, I assume it's not soo useful, but adds some complexity and potential performance impact.
- CM_Model_LocationTest: You can use `Location::equals(Location)` to compare locations more easily (provided by `interface CM_Comparable`).
- CM_Model_StorageAdapter_Database::findByData: Add test
- CM_Model_Location::findByAttribute: I think this should actually return `CM_Model_Location` instances (and not `CM_Model_Location_Country` etc), because these are the ones we use. @tomaszdurka please confirm? (function docu already states this)
- CM_Model_Location::findByAttribute: There's already a "level to type" map in `_getLocation` - make it reusable? (`getTypeByLevel`?)
- CM_Model_Location::findByAttribute: Use `CM_Cache_Abstract::key()` to get a cache key from an array of data? -> will not to extend that function to include array keys!

@tomaszdurka
> and it is easier to invalidate them within `CM_Model_Location_Abstract`. I agree then.

I don't see how we could invalidate them in "Location" either? But I think the 24h expiration is okay as Daniel says - because these objects change rarely, and up-to-date data is not crucial.

dlondero:
@njam I don't understand the last point about using `CM_Cache_Abstract::key()`. Should we or we should not? Also do we really want to use an md5 as key? Probably we'll never try to search a key to check its content but if needed the current version is more readable. Wdyt?

njam:
- LocationTest: Instead of `new CM_Model_Location(CM_Model_Location::LEVEL_STATE, self::$_baselStadt)` you could use `self::$_baselStadt->get(CM_Model_Location::LEVEL_STATE)`

Rest looks good, agree on keeping the cache key code.

dlondero:
@njam afaik `self::$_baselStadt` is only a string with the `id` returned by `CM_Db_Db::insert`

njam:
Ah true - maybe rename them to "switzerlandId" etc?

Rest looks good

dlondero:
@njam makes a lot of sense. Please review. Should be good to go.

njam:
Looks good!

PR #1592, opened 2015-01-14
Update icons

christopheschwyzer:


christopheschwyzer:
@njam maybe you want to review my second commit (Cli change)

njam:
Code change of f5ef55e7c5903ea8bb86d411620226b93e527133 looks good.
I understand this will remove *any* horizontal whitespace? So it's not possible anymore to have an icon that doesn't use the full width? -> We don't have any such icons so it's okay!

christopheschwyzer:
Yes. I think that's ok.

christopheschwyzer:
I will go on and merge this

PR #1593, opened 2015-01-15
Emoticon refactoring

alexispeter:
As discussed in https://github.com/cargomedia/CM/pull/1573, implementing a file-based emoticon approach as opposed to keeping data in a mysql table.
This basically skips the emoticonRefresh operation from CM_Css_Cli, as emoticon data is now computed on-the-fly and cached in apc, making addition and removal of emoticons easier.
In addition to the pre-existing image-files in `<namespace>/layout/default/resource/img/emoticon/`, emoticons may have additional json files which define additional codes.
Furthermore, a new class `CM_Emoticon` is introduced to allow easier handling of emoticons.


alexispeter:
please review @njam

njam:
- JSON files: Add newlines at end of files
- CM_Emoticon::_buildEmoticonData: Rename to `_readEmoticonData`?
- CM_Emoticon::getEmoticonData: You'r passing `0` as ttl to `CM_Cache_Abstract::set`. As a result the default will be used. I assume that's not what you wanted?
- CM_Emoticon::findCode: Rename to findByCode
- CM_Emoticon::findCode: Use strict version of `array_search` - third arg
- CM_Emoticon::findCode: Rename $emoticon to $data?
 - Let's keep these names consistent in general - In `_load` we have $dataList (multiple) and $data (single). In getEmoticonData it's $emoticonData (single). And in findCode it's $emoticonData (multiple) and $emoticon (single). I'd suggest $dataList and $data.
- CM_Emoticon::findName: Rename to findByName
- CM_Emoticon::_buildEmoticonData: How about using `CM_File::listFiles` instead of `glob`?
- CM_Emoticon::_buildEmoticonData: Prefer single quotes (exception msg)
- CM/Usertext/Filter/EmoticonTest.php: Why did you replace normal emoticon names with *additionalCodes*? (`_getEmoticonImg` usage)

alexispeter:
> CM_Emoticon::_buildEmoticonData: How about using CM_File::listFiles instead of glob?

Using CM_File::listFiles() would usually take ~ 110 ms to read the filesystem, sporadically increasing to ~400 ms.
Using glob consistently takes 70-80ms.

njam:
listFiles performance: In such cases I would suggest to add a comment to the source code explaining why a more complex implementation was chosen (for performance considerations).

Did you consider that `listFiles()` includes the `CM_File` instantiation?
Comparing 100 runs of both implementations I get:
```
listFiles: 2781.89 ms
glob: 1819.71 ms
```
After [this optimization](https://github.com/cargomedia/CM/pull/1594) I get:
```
listFiles: 1728.98 ms
glob: 2081.58 ms
```

Code:
```php
$dir = CM_File::createTmpDir();
for ($i = 0; $i < 1000; $i++) {
    $dir->joinPath("file-{$i}")->write('hello');
}

CM_Util::benchmark('listFiles');
for ($i = 0; $i < 100; $i++) {
    $fileList = $dir->listFiles();
}
echo 'listFiles: ' . CM_Util::benchmark('listFiles') . PHP_EOL;

CM_Util::benchmark('glob');
for ($i = 0; $i < 100; $i++) {
    $pathList = glob($dir->getPathOnLocalFilesystem() . '/*');
    $fileList = \Functional\map($pathList, function ($path) {
        return new CM_File($path);
    });
}
echo 'glob: ' . CM_Util::benchmark('glob') . PHP_EOL;
```

alexispeter:
As discussed, the periodic hiccups in performance in FilesystemAdapter::listFiles() is caused by the nfs-mount and can be ignored for production.

alexispeter:
please review @njam

njam:
- CM_Emoticon:_readEmoticonData: Inline $emoticonPath?

alexispeter:
please review @njam

njam:
looks good!

alexispeter:
tagged https://github.com/cargomedia/CM/releases/tag/1.94.0

PR #1596, opened 2015-01-16
Adjust floatbox height in fullscreen mode.

vogdb:
To fix http://stackoverflow.com/questions/8468066/child-inside-parent-with-min-height-100-not-inheriting-height/8468131#8468131


vogdb:
@njam FYI.

christopheschwyzer:
Sorry I don't understand ;). We set a `min-height` (the actual height of the screen) to the floatbox using javascript. So `height` becomes superfluous? (We use `min-height` because on smaller screens it can extend the height of the screen) 

What is `height: 1px` for?

vogdb:
I took this solution  from http://stackoverflow.com/a/21836870/495656

njam:
@christopheschwyzer the problem was that if the floatbox content doesn't fit the full height of the viewport (let's say only half of it), then even though we use floatbox' `fullscreen` mode the content will not span the whole viewport.

@vogdb @christopheschwyzer it's good if we can get that fixed. Please make sure the overflow still works *if* the content is higher than 100% though (like on the photo view), on all browsers!

christopheschwyzer:
I see, thanks for noting. I discussed it with @vogdb as well. I tested it on iOS simulator only - works fine. Maybe we should test this on other devices as well. I will check it tomorrow!

vogdb:

christopheschwyzer:
I tested this briefly on Chrome, FF, Safari, iOS Safari, Android Chrome, Android Stock, Android FF, Windows Phone IE. There seems to be an issue in IE10 - Will try to investigate tomorrow.

vogdb:
Hm. It worked ok in IE10. Don't forget that currently `toggleShareMode` is enabled only for large screens!

christopheschwyzer:
Ah my mistake. Works in IE10 as well. I restarted tests. Lgtm!

PR #1599, opened 2015-01-20
Configure redis database=2 in tests

njam:
Addressing https://github.com/cargomedia/sk/issues/1131

@dlondero @tomaszdurka please review

### DIFF
--- a/library/CM/Elasticsearch/Index/Cli.php
+++ b/library/CM/Elasticsearch/Index/Cli.php
@@ -40,7 +40,7 @@ class CM_Elasticsearch_Index_Cli extends CM_Cli_Runnable_Abstract {
             $indexName = $index->getIndex()->getName();
             $key = 'Search.Updates_' . $index->getType()->getName();
             try {
-                $ids = CM_Redis_Client::getInstance()->sFlush($key);
+                $ids = $this->_getRedis()->sFlush($key);
                 $ids = array_filter(array_unique($ids));
                 $index->update($ids);
                 $index->getIndex()->refresh();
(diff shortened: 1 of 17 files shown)


dlondero:
@njam lgtm!

tomaszdurka:
Just wanted to note down (for some time already) that I like this change a lot.
Small thing is that instantiating redis client is quite complicated and requires dev to look into code and understand config array. Maybe some extended docblock would help with that.

njam:
I googled a bit for such a docu annotation, but couldn't find anything.

tomaszdurka:
Yes unfortunately there is no supported (by intellij) docu syntax. But maybe use something on our own aka:
```php
    /**
     * @param array $config {
     *     @type string $host
     *     @type int $port
     * }
     */
```
Unfortunately this doesn't work well with intellij formatting.

alexispeter:
We have a few methods where we use docs to describe the structure of an array.

```php
/**
* @param array              $chatMessage ['id' => int, 'content' => string, 'createStamp' => int, 'chat' => SK_Entity_Chat 'user' => SK_User]
* @param CM_Frontend_Render $render
* @param boolean|null       $skipOpenChatForSender
*/
```
###### CM_Model_Abstract::factoryGenericMultiple #####
```php
/**
* @param array    $idTypeList [['type' => int, 'id' => int|array],...] | [int|array,...] Pass an array of ids if $modelType is used
* @param int|null $modelType
* @return CM_Model_Abstract[] Can contain null-entries when model doesn't exist
*/
```
```php
/**
* @param string $channel
* @return array ['key' => string, 'type' => int]
*/
```

njam:
True -> https://github.com/cargomedia/CM/pull/1612

PR #1602, opened 2015-01-21
Add text-overflow: ellipsis; to buttons

christopheschwyzer:


christopheschwyzer:
@dlondero please review

dlondero:
I'm not totally sure about this: shouldn't the form adapt to have a wider button?  If users can't read the label totally they can miss the purpose of the button.

christopheschwyzer:
Yes, I agree. This rule only applies when there is an explicit `width` or `max-width` set - which we try to avoid of course. But there might be some edge cases where this can come in handy.

dlondero:
Ok, thanks for the clarification. Looks good then!

christopheschwyzer:
Build keeps failing for php 5.5. 
screen shot 2015-01-21 at 19 12 31

PR #1600, opened 2015-01-21
Create CM_FormField_Search.

vogdb:

PR #1601, opened 2015-01-21
Add empty action to block.form.php.

vogdb:
Connects to #1600.


vogdb:
@njam please look.

njam:
lgtm!

PR #1603, opened 2015-01-22
Create our own custom Modernizr test 'iOS' to detect iOS devices.

vogdb:
Currently we have already a places where we rely on this user agent detection. Those places are:
 - `jquery.iosKeyboardActive.js`
 - `SK/Component/Comments.js::ready` L32.

Now we need to add another such place at SK_Component_HeaderBar to decide whether we need to popup or toggle the Share component. So, my proposition is to add our custom `Modernizr.iOS`, reuse it at the all existing places and use in the all future cases where we need this user agent detection.

The implementation will be easy:
```js
Modernizr.addTest('iOS', function () {
  return navigator.userAgent.match(/(iPad|iPhone|iPod)/i);
});
```

@njam @christopheschwyzer ?

### DIFF
--- a/client-vendor/after-body/modernizr/modernizr.custom.js
+++ b/client-vendor/after-body/modernizr/modernizr.custom.js
@@ -4,7 +4,7 @@
  */
 
 (function() {
-  Modernizr.addTest('inputfixed', function() {
+  Modernizr.addTest('inputinfixed', function() {
     return !navigator.userAgent.match(/(iPad|iPhone|iPod)/i);
   });
 })();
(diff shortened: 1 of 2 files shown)


vogdb:
@njam FYI.

christopheschwyzer:
Aren't we messing with modernizr conventions now? _Sniffing_ for user agent is not a _feature detection_. Also, this is now addressing a very similar problem as the `jquery. iosKeyboardActive` plugin. However, I don't have a better idea. :/

vogdb:
I do not have a better idea either. So, lets stick to this decision for a while. At least it will be encapsulated in one place. Now it is starting to spread =(.

vogdb:
@christopheschwyzer please review,

christopheschwyzer:
As for the class naming: @njam wrote `inputinfixed`, @vogdb you used `inputfixed`. I like @njam's a bit better as it describes more a functioning feature.

vogdb:

christopheschwyzer:
lgtm!

PR #1608, opened 2015-01-26
Add global event bus for App

vogdb:
In `CM_App`:
```
cm.event.trigger('notification.conversation.change', 12)
```

In view:
```
  globalEvents: {
    'notification.conversation.change': function(count) {
    }
  },
```

### DIFF
--- a/library/CM/App.js
+++ b/library/CM/App.js
@@ -996,7 +996,7 @@ var CM_App = CM_Class_Abstract.extend({
     }
   },
 
-  appEvents: {
+  events: {
     /**
      * @type {Backbone.Events}
      */
--- a/library/CM/App.js
+++ b/library/CM/App.js
@@ -996,15 +996,15 @@ var CM_App = CM_Class_Abstract.extend({
     }
   },
 
-  events: {
+  appEvents: {
     /**
-     * @type {Object}
+     * @type {Backbone.Events}
      */
     _dispatcher: _.clone(Backbone.Events),
 
     /**
      * @param {String} eventName
-     * @param {Function} callback fn(CM_View_Abstract view, array data)
+     * @param {Function} callback fn(array data)
      * @param {Object} [context]
      */
     bind: function(eventName, callback, context) {
@@ -1013,7 +1013,7 @@ var CM_App = CM_Class_Abstract.extend({
 
     /**
      * @param {String} eventName
-     * @param {Function} callback fn(CM_View_Abstract view, array data)
+     * @param {Function} callback fn(array data)
      * @param {Object} [context]
      */
     unbind: function(eventName, callback, context) {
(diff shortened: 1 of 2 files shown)


vogdb:
@tomaszdurka please review =)

njam:
- I'd still vote for `cm.event` (and not `cm.globalEvents`). Or are there reasons to use `globalEvents` rather? Maybe then on the View call it `bindAppEvent`?

@tomaszdurka please have a look!

vogdb:
@njam Please look. Did I understand you correctly?

njam:
- CM/App.js: Docu should be type `Backbone.Events`?
- CM/App.js bind/unbind: Docu for `callback` lists params which are not present
- CM/App.js trigger: Make `data` optional
- CM/View/Abstract.js: Property should be `appEvents` and not `events` (latter is already used by Backbone!)

@tomaszdurka can you check and let us know what you think of this in general?

tomaszdurka:
I am a bit confused, `events` property on view was mainly used for dom elements' bindings (via Backbone), right?. Now they are binded on global cm app instance. Won't this break whole app?

Regarding usage of `cm.app.events` I am always against extending global singleton usages :)

vogdb:
@njam please review.

tomaszdurka:
@vogdb I don't think @njam meant to rename `app.events` to `app.appEvents` - it was probably just about `view.events` which would conflict with Backbone ones.

I am not sure what you guys want this functionality for, but I would propose to not add `appEvents` property and rather bind in-code when needed.

njam:
- cm.events: should be `cm.event` (as discussed all other similar objects are singular).
- cm.event: bind/unbind: Param-docu for callback should be `Array` instead of `array`?

I'm in favor of `View::appEvents` - it makes usage simpler. @tomaszdurka can you share your concerns?

vogdb:
> cm.event: bind/unbind: Param-docu for callback should be Array instead of array?

No. Because all other like `stream` or `action` use `array`.

@njam please review.

tomaszdurka:
I agree with @njam that it should be `Array` as this is the real type name. 
If you try typing `{array}` instead of `{Array}` in param doc-block intellij will highlight it as an error. 
Of course our `fn()` syntax is something which does not support this syntax check, but still we should use valid class names. If the param suppose to be type of `CM_Class_Abstract` you wouldn't use `cm_class_abstract` would you?

vogdb:
So, why you have used `array` previously?

tomaszdurka:
I think longer time ago we used lowercase array all over the project, but then intellij started to mark them as error - so we search replaced all curly bracket wrapped usages to uppercase. These were missed.

njam:
It looks good to me now!

(btw here's the JSDoc suggestions  on the `callback` topic: http://usejsdoc.org/tags-callback.html)

vogdb:
Yep. Such format exists but should I break the docu consistency with other code? Do you give your permission?

njam:
I would keep it as is. The signature of the callback is pretty straight forward and well known, so I think there's no need to use this advanced syntax, wdyt?

vogdb:
I'd use it as it is now. Do we need to wait for Tomasz's review?

tomaszdurka:
Ok for me.

PR #1611, opened 2015-01-28
IP block delete older should be configurable

mariansollmann:


mariansollmann:
@alexispeter please review.

alexispeter:
lgtm
What's the reason this needs to be configurable?

PR #1612, opened 2015-01-28
Add docblock comment about redis' config elements

njam:
@alexispeter @tomaszdurka what do you think?


alexispeter:
Should be pretty clear what goes in that array :+1:

tomaszdurka:
Looks good

PR #1613, opened 2015-01-28
Migrate to GA Universal Analytics

njam:
@moby86 @fauvel 

Docu: https://developers.google.com/analytics/devguides/collection/upgrade/reference/gajs-analyticsjs

Things to consider:
- [**Cross Domain Tracking**](https://developers.google.com/analytics/devguides/collection/upgrade/reference/gajs-analyticsjs#cross-domain): I will drop it, because we don't need to track sessions across domains. It would require us to list all linked domains on on every website.
- [**Custom Dimensions/Metrics**](https://developers.google.com/analytics/devguides/collection/analyticsjs/custom-dims-mets): These replace the previous *custom variables*. Dimensions are strings, Metrics are numbers.
 - We will need to configure them in GA Admin: "Property > Custom Definitions"
- [**Ajax Navigation Tracking**](https://developers.google.com/analytics/devguides/collection/analyticsjs/single-page-applications): We need to set the `page` whenever we track a new *pageview*, otherwise the tracker object would reuse the *page* from its initialization.

PR changes:
- `CM_Http_Response_*`: Are now service-manager aware, mandatory second argument
- `CMTest_TestCase`: Is now service-manager aware
- `CMService_GoogleAnalytics_Client::addPageView()`:  Path now mandatory
- `CM_Service_Tracking_ClientInterface::trackPageView()`: Path now mandatory

### DIFF
--- a/tests/helpers/CMTest/library/CMTest/TestCase.php
+++ b/tests/helpers/CMTest/library/CMTest/TestCase.php
@@ -17,7 +17,7 @@ abstract class CMTest_TestCase extends PHPUnit_Framework_TestCase implements CM_
             CM_Config::get()->CM_Site_Abstract->class = get_class($siteDefault);
         }
 
-        $this->setServiceManager(CM_Service_Manager::getInstance());
+        $this->setServiceManager(CMTest_TH::getServiceManager());
 
         parent::runBare();
     }
--- a/tests/helpers/CMTest/library/CMTest/TestCase.php
+++ b/tests/helpers/CMTest/library/CMTest/TestCase.php
@@ -17,7 +17,7 @@ abstract class CMTest_TestCase extends PHPUnit_Framework_TestCase implements CM_
             CM_Config::get()->CM_Site_Abstract->class = get_class($siteDefault);
         }
 
-        $this->setServiceManager(new CM_Service_Manager());
+        $this->setServiceManager(CM_Service_Manager::getInstance());
 
         parent::runBare();
     }
@@ -186,7 +186,7 @@ abstract class CMTest_TestCase extends PHPUnit_Framework_TestCase implements CM_
      */
     public function getResponse(CM_Http_Request_Abstract $request) {
         $className = CM_Http_Response_Abstract::getResponseClassName($request);
-        $serviceManager = CM_Service_Manager::getInstance();
+        $serviceManager = self::getServiceManager();
         return $this->mockClass($className)->newInstance([$request, $serviceManager]);
     }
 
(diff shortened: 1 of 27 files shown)


njam:
@fauvel please review

Tracks like this in the browser:
screen shot 2015-01-28 at 15 50 43

njam:
@fauvel there were a few more *adjustments* necessary so that the new GA tracking works well with ajax navigation. Let's have a brief look together? Everything's summarized in the description as well!

njam:
@mariansollmann @vogdb please review

vogdb:
Ok. lgtm but my opinion here is very doubtful.

tests/helpers/CMTest/library/CMTest/TestCase.php
```diff
@@ -14,6 +16,9 @@ public function runBare() {
             ));
             CM_Config::get()->CM_Site_Abstract->class = get_class($siteDefault);
         }
+
+        $this->setServiceManager(new CM_Service_Manager());
```
mariansollmann: not sure, but shouldn't you use `CMTest_TH::getServiceManager`?

tests/helpers/CMTest/library/CMTest/TestCase.php
```diff
@@ -181,7 +186,8 @@ public function createRequestAjax(CM_View_Abstract $view, $methodName, array $pa
      */
     public function getResponse(CM_Http_Request_Abstract $request) {
         $className = CM_Http_Response_Abstract::getResponseClassName($request);
-        return $this->mockClass($className)->newInstance([$request]);
+        $serviceManager = CM_Service_Manager::getInstance();
```
mariansollmann: Shouldn't this use now `$this->getServiceManager`?

tests/library/CM/Http/Response/PageTest.php
```diff
@@ -129,10 +125,9 @@ public function testProcessTrackingViewer() {
      */
     protected function _getServiceManager($codeGoogleAnalytics, $codeKissMetrics) {
         $serviceManager = new CM_Service_Manager();
```
mariansollmann: `$this->getServiceManager`?

njam: Here I rather want to create a new (empty) service-manager - with only the services I register (GA, Kiss), but no database etc.

mariansollmann: agree

mariansollmann:
looks good

mariansollmann:
makes sense. looks good

PR #1614, opened 2015-01-28
make only servicemanager-dependent storage-systems ServiceManagerAware

alexispeter:
Only storage systems that actually use the service manager should be ServiceManagerAware.

Previously, even when using storage system that don't use the serviceManager, setting storage with `$clockworkManager->setStorage($storage)` would throw an exception if `$clockworkManager` had no serviceManager set.


alexispeter:
please review @tomaszdurka

tomaszdurka:
Looks good to me.

PR #1609, opened 2015-01-29
Example JS Error

vogdb:
We should fix this eventually. Happens in iOS in Example.
screen shot 2015-01-27 at 10 03 48

### DIFF
--- a/library/CM/Form/ExampleIcon.js
+++ b/library/CM/Form/ExampleIcon.js
@@ -17,10 +17,7 @@ var CM_Form_ExampleIcon = CM_Form_Abstract.extend({
     this.on('change', function() {
       this._updateCss();
     }, this);
-    var self = this;
-    setTimeout(function() {
-      self._updateCss();
-    }, 0);
+    this._updateCss();
   },
 
   /**
--- a/library/CM/Form/ExampleIcon.js
+++ b/library/CM/Form/ExampleIcon.js
@@ -17,7 +17,10 @@ var CM_Form_ExampleIcon = CM_Form_Abstract.extend({
     this.on('change', function() {
       this._updateCss();
     }, this);
-    this._updateCss();
+    var self = this;
+    setTimeout(function() {
+      self._updateCss();
+    }, 0);
   },
 
   /**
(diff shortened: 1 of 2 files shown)


vogdb:

christopheschwyzer:
Using `setTimeout` seems _hacky_? Why is it necessary? Alternatives?

vogdb:
It is done because `ready` executes before the corresponding template is displayed. As the result `this.getField('sizeSlider | color | ....').getValue()` return incorrect values. Mostly NaN or #000000. To fix this we need to execute `ready` only after the template is displayed. Let me think about the alternatives.

njam:
`_.delay()` :P

vogdb:
Isn't it the same?

vogdb:
Another alternative would be to listen to some other init event like `tabs.changed` for example. Inside `CM_Form_ExampleIcon` we can attach to it `this.getParent().$el.on('tabs.changed', function() {...});`

vogdb:
@njam @christopheschwyzer ?

christopheschwyzer:
I don't know what the best approach is...

njam:
About the loading of `View`s - in what state are child views when the parent's `View.ready()` is called:
- A: The child templates are all loaded and in the DOM
- B: The child `View`s have been instantiated, their objects are registered on the parent
- C: The childrens' `ready()` function has not been called yet.

I can confirm that the slider's value cannot be read with `getValue()` unless the form-field has been `ready()`'d. The reason seems to be that `CM_FormField_Integer.getValue()` contains this code:
```js
  getValue: function() {
    return parseInt(this.$('.noUiSlider').val());
  },
```
So it dependson this "noUiSlider" plugin to be initialized (which happens in `ready()`).
With formfields we usually try to make sure they also have a plain `<input>` field somewhere, and would also work without whatever javascript plugin we have for a nice UI. In this case you can see when we set up the "noUiSlider" plugin we always install hooks to make sure the value of the `<input>` and the value of the "noUiSlider" stay in sync!
Therefore I think the easiest resolution is to remove `CM_FormField_Integer.getValue()` completely. It will then fall back to `CM_FormField_Abstract.getValue()` which will read the value from the `<input>`, which should work even when the fancy UI hasn't been initialized yet!

I can't reproduce similar problems with color fields?

vogdb:
> I can't reproduce similar problems with color fields?

Probably because they don't introduce their own `getValue()` as `CM_FormField_Integer `.
I removed `CM_FormField_Integer.getValue()` and it works now. I checked its usage in search/members. It works without this method too. I guess, the fix is good to apply.
@njam please look.

njam:
lgtm!

PR #1618, opened 2015-01-29
Backslashes in emoticon codes are escaped in markdown

dlondero:


dlondero:
@alexispeter please review. AFAIK `:-\\` is the only emoticon we support with a double backslash which gets modified by markdown filter. Proposing a solution. Feel free to suggest other approaches.

alexispeter:
I don't really know the involved mechanisms, but isn't there a way to address this by changing/correcting the ordering in which filters are applied in this particular mode?

alexispeter:
Imo emoticons should be the first thing to be transformed and then somehow escaped so the markdown doesn't get escaped by other markdown escapers.

njam:
The root cause is that `CM_Usertext_Filter_Markdown` converts `\\` to `\`, correct?
Do you know why this happens? Is this some special syntax in markdown?

If it is then I agree the right approach is to use `EscapeMarkdown/UnescapeMarkdown` to work around this. Maybe a more generic solution is possible though (for `\\`)? But for that we would need to know why markdown does this replacement. Do you know?

dlondero:
That's because backslash is the [escape character](http://daringfireball.net/projects/markdown/syntax#backslash) used in Markdown.

dlondero:
That's because backslash is the [escape character](http://daringfireball.net/projects/markdown/syntax#backslash) used in Markdown.

alexispeter:
Maybe escapemarkdown_emoticon should also escape backslashes?
Basically `:-\\` would turn into `:-\\\\` which escapemarkdown would again turn into `:-\\`.
Another possibility is to wrap emoticons with special characters that escapemarkdown would then ignore, but we would need to use a lexer-approach instead of simple regular expressions to do this.

njam:
Tricky stuff.

Escaping emoticon characters correctly in `CM_Usertext_Filter_Emoticon_EscapeMarkdown` sounds like a good idea.
Maybe also the existing logic could do that (transform `_` into `\_`)?

Two ideas:
- Replace escapable characters of emoticon codes in `CM_Usertext_Filter_Emoticon_EscapeMarkdown`
- Replace emoticon codes with their `:emoticon:` counterpart in `CM_Usertext_Filter_Emoticon_EscapeMarkdown`. This way we'll have more control over what characters emoticons contain.

@alexispeter @dlondero wdyt?

alexispeter:
I would tend towards solution 1. 
However, wouldn't it be possible (better?) to apply `CM_Usertext_Filter_Emoticon` first and just escape the html code this generates?

dlondero:
> However, wouldn't it be possible (better?) to apply CM_Usertext_Filter_Emoticon first and just escape the html code this generates? That way you don't have to worry about escaping characters within emoticon-codes.

I guess this works in `markdown` mode. In `markdownPlain` instead we also apply `CM_Usertext_Filter_Striptags()` at some point and it would strip the image tags...

dlondero:
> Replace emoticon codes with their :emoticon: counterpart in CM_Usertext_Filter_Emoticon_EscapeMarkdown. This way we'll have more control over what characters emoticons contain.

Probably this one is the most clean solution? In the end we have just a few emoticons with alternatives.

alexispeter:
>> Replace emoticon codes with their :emoticon: counterpart in CM_Usertext_Filter_Emoticon_EscapeMarkdown. This way we'll have more control over what characters emoticons contain.

> I guess this works in markdown mode. In markdownPlain instead we also apply CM_Usertext_Filter_Striptags() at some point and it would strip the image tags...

That's why they should be escaped so they don't get stripped :>

I guess in the end it doesn't make a big difference, we will have to scan for emoticons multiple times anyway.

njam:
Let's go with this then (because it's simpler)?
> - Replace emoticon codes with their `:emoticon:` counterpart in `CM_Usertext_Filter_Emoticon_EscapeMarkdown`. This way we'll have more control over what characters emoticons contain.

alexispeter:
agree

dlondero:
@alexispeter please review

alexispeter:
* library/CM/Usertext/Filter/Emoticon/EscapeMarkdown.php
 + line 10: rename `$additionalMapping` to `$additionalCodesMapping`?
 + line 22: add typehinting to the function-argument, saves a line
 + line 29: rename `$filteredEmoticon` to `$emoticon`?

There are also a few codes that are affected by `CM_Usertext_Filter_Escape`.
##### raw #####
```
:angry: :-@ (:-&
:broken_heart: &lt;/3 </3
:cold_sweat: :-\\
:confused: :-[
:cry: :''( ;-( :''-(
:disappointed: :-( :( :[ =(
:dizzy_face: %-) %) O.o o.O
:expressionless: :| :-| :-/
:grimacing: :-X :-x
:grin: :-D ;D =D
:heart: &lt;3 <3
:imp: 3) 3-)
:innocent: O) o-)
:kissing_heart: :-<> :-* :*
:mask: (:-$
:open_mouth: :-O :-o
:smiley: :-) :) =)
:stuck_out_tongue: :P :-P
:stuck_out_tongue_winking_eye: ;P ;-P
:sunglasses: B-) B) 8-) 8)
:wink: ;-) ;)
:yum: :-p
```
##### rendered #####
image

I think it would be best to always apply the emoticon filters first.

alexispeter:
Might be worth checking the codes we use. Do we even need to include html-entities like `&lt;/3` for `</3`?

dlondero:
> I think it would be best to always apply the emoticon filters first.

@alexispeter so you would something like `CM_Usertext_Filter_Emoticon_ReplaceAdditional` before `CM_Usertext_Filter_Escape` in `CM_Usertext_Usertext::setMode()` so we always replace additional codes to :emoticon: ? Sounds good to me, also probably it's better because now replacement is done in `*_EscapeMarkdown` and it's not totally related to Markdown. Wdyt?

alexispeter:
Something like that.
As for the html-encoded entities versions of additional codes: @njam thinks they are historical and we should be able to remove them.

dlondero:
@alexispeter please review

alexispeter:
* mode `escape` should not apply the ReplaceAdditional-filter

@njam just to be sure; when saying we have those html-special-char escaped smileys for historical reasons, do you mean they are no longer needed or do you mean we still need them because they could have been used like this in a text somewhere that should still be able to render them?

njam:
I'm not sure about this! It was merely a guess.
I would suggest to check
 - Does it work without them for new ones
 - Do we have lots of them in existing texts (send me a query)

alexispeter:
After the adjustment to the copperegg-metric, slow queries are significantly down.
image
I'll still look over our application and see what optimisations we can perform.

dlondero:
@alexispeter please review

alexispeter:
Only one thing: The test should also test mode `escape` with a smiley that contains special chars.

alexispeter:
lgtm

PR #1619, opened 2015-01-30
Breaking: Change cm.window.title functionality.

vogdb:
Remove Chat notification. Keep it for saving the original title of the current page.

> For changing the window's title we have cm.window.title - at the moment it's used to scroll a message "New chat message from " through the title.
I would suggest to remove this feature in favor of showing the counter.

### DIFF
--- a/library/CM/App.js
+++ b/library/CM/App.js
@@ -677,47 +677,7 @@ var CM_App = CM_Class_Abstract.extend({
       }
     },
 
-    title: {
-      _messageStop: function() {
-      },
-      _messageTimeout: null,
-
-      ready: function() {
-        var handler = this;
-        $(window).focus(function() {
-          handler._messageStop();
-        });
-      },
-
-      message: function(msg) {
-        if (cm.window.hasFocus()) {
-          return;
-        }
-        var handler = this;
-        var sleeper = function(offset) {
-          offset += 4;
-          if (offset >= msg.length) {
-            handler._messageStop();
-          } else {
-            document.title = msg.substring(offset, msg.length);
-            handler._messageTimeout = setTimeout(function() {
-              sleeper(offset);
-            }, 400);
-          }
-        };
-
-        this._messageStop();
-        var originalTitle = document.title;
-        document.title = msg;
-        this._messageTimeout = setTimeout(function() {
-          sleeper(0);
-        }, 1500);
-        this._messageStop = function() {
-          document.title = originalTitle;
-          clearTimeout(handler._messageTimeout);
-        };
-      }
-    }
+    title: ''
   },
 
   storage: {
--- a/library/CM/App.js
+++ b/library/CM/App.js
@@ -614,7 +614,6 @@ var CM_App = CM_Class_Abstract.extend({
       }).blur(function() {
         handler._hasFocus = false;
       });
-      this.title.ready();
     },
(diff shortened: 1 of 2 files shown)


vogdb:
@njam please review.

njam:
How about providing the functionality we need still on `cm.window.title`?

Can you explain how you wanna implement the counter/prefix? What happens on navigation? I guess we want to keep the counter/prefix?

How about:
- `cm.window.title.setPrefix('(3) ')`
- `cm.window.title.setPrefix(null)`

and the rest happens transparently?

vogdb:
> Can you explain how you wanna implement the counter/prefix? What happens on navigation? I guess we want to keep the counter/prefix?

In Navigation I wanted to have something simple as
```js
document.title = cm.language.get('({$count}) Inbox', {count: count}) + '. ' + cm.window.title;
```

vogdb:
`setPrefix` is ok too. What do you like more?

vogdb:
Wait here. I'm preparing another solution.

vogdb:
Now it is better. @njam please look.

njam:
Yup, I like it!

- _update: Two ternary operators are a bit of an overkill imo :D Can you make it a bit more readable?
- Is `setText()` called on initial page load, now that `this.title.ready()` is removed?

vogdb:
> Is setText() called on initial page load, now that this.title.ready() is removed?

There is no need to replace `this.title.ready()` with `setText` as the functionality of `this.title.ready` is not needed anymore. It just restored the original `document.title` after showing a message. Now we don't have any messages, so no need to restore original title.

vogdb:
@njam please review.

njam:
I don't get it completely yet, imagine this order of events:
1. Initial page loaded (title is: `foo`)
2. Prefix set to "bar" (title is: `bar `, but should be `bar foo`)
3. Prefix is removed (title is ``, but should be `foo`)

Am I missing something?

vogdb:
It's hard to say. I've just recorded the video of how it works now. Please look. Currently is has only one problem. Prefix is flushed on page refresh. I sent the link to you by email.

njam:
I meant when you set the profix *immediately* after the first page load:
1. Load any page
2. Run `cm.window.title.setPrefix("foo")
3. Title is now "foo" only (instead of containing the rest of the title as well)

We should set the "text" part of the title on initial page load as well!

vogdb:
@njam please review.

njam:
Yes that should solve it. Small notes:
- I would suggest to make `_text` not nullable. This is the main document's title and it should always be set (even if it's an empty string)?
- How about keeping the `ready()`. This way the related logic stays in that "module".

vogdb:
@njam please review.

njam:
...one more thing
- Not sure about the validation in `setText()`. It's a bit inconsistent as the other function doesn't have validation either. If we add validation I would also suggest to rather throw (`cm.error.triggerThrow`) in case of invalid values. But imo it can also be removed. wdyt?

vogdb:
I think it's better to remove it then.

vogdb:
@njam please review.

njam:
Agree. Looks good!

njam:
Be aware that this is a breaking change (add to commit message when merging)!

PR #1620, opened 2015-01-30
mongodb increase batchSize to an arbitrary, high number

alexispeter:
Should help us identify slow queries, as only the initial fetch-operation contains the necessary information.

### DIFF
--- a/library/CM/MongoDb/Client.php
+++ b/library/CM/MongoDb/Client.php
@@ -130,6 +130,10 @@ class CM_MongoDb_Client extends CM_Class_Abstract {
      * @return Iterator
      */
     public function find($collection, array $criteria = null, array $projection = null, array $aggregation = null) {
+        $batchSize = null;
+        if (isset(self::_getConfig()->batchSize)) {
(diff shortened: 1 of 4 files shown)


alexispeter:
please review @njam
cc @kris-lab

njam:
Can you explain the reasoning?

Does the batchSize affect the query execution? Or is the query first executed, and the data then just buffered?

njam:
> The problem we face, is that getmore operations don't show enough information to determine what query they belong to.

> ok I see, you want to get rid of "getmore" to be able to debug further?

Make this configurable? This way we won't need to do a change in CM to change this.

alexispeter:
please review @njam

njam:
Add a test? (using `MongoCommandCursor::info()`)

alexispeter:
please review @njam

njam:
lgtm!

alexispeter:
tagged https://github.com/cargomedia/CM/releases/tag/1.97.2

PR #1616, opened 2015-01-29
Track banner impressions and clicks with GA events

njam:
Depends on https://github.com/cargomedia/CM/pull/1613

This will allow us to continuously monitor ad impressions and CTR, and segment by device, geo, gender etc in GA.
(Click tracking will only work for image banners.)

Event parameters:
- **Category, Action, Label**
- 'Banner', 'Click', '[zone-id]'
- 'Banner', 'Click', '[zone-id]'

See:
- http://www.statstory.com/tracking-clicks-and-impressions-in-google-analytics/
- http://www.opensidedigital.com/how-track-outbound-links-google-analytics

cc @moby86 @j-dalrymple @fauvel

PR #1621, opened 2015-01-30
Change declaration of global var `ga` in tracking of banners.

vogdb:
The followup of #1616. Now we have an error in the dev version.


vogdb:
@njam please review.

njam:
Looks very good!

PR #1622, opened 2015-02-02
include aggregation in cachekey-base for CM_PagingSource_MongoDb

alexispeter:
Was overlooked when I incorporated aggregations into our mongodb code. 
With the current implementation, pagings using the same collection, criteria and projection but different aggregation would experience cacheKey conflicts.


alexispeter:
please review @njam

njam:
lgtm

PR #1623, opened 2015-02-02
unittests: only generate internals if necessary

alexispeter:
Only generate internals if they aren't already defined.
Derived codebases should have their internals predefined anyway, not needing to autogenerate them for tests, which can take a long time.

### DIFF
--- a/tests/helpers/CMTest/library/CMTest/TestSuite.php
+++ b/tests/helpers/CMTest/library/CMTest/TestSuite.php
@@ -15,6 +15,9 @@ class CMTest_TestSuite {
     }
 
     public function generateInternalConfig() {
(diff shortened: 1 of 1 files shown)


alexispeter:
please review @njam

njam:
@tomaszdurka please review

tomaszdurka:
I agree, but I don't like the condition (checking if max value set for one of them classes)... what else we could do?

alexispeter:
The maxValue isn't class-specific, it's application-wide.

tomaszdurka:
Ah indeed, I thought we use typesMaxValue for each of *Typed* classes. Well let it be then.

alexispeter:
If it's any consolation, I don't like it either, but don't see any other way.

PR #1626, opened 2015-02-02
openx.js: Add event "Impression-Clickable"

vogdb:
Could you add to `jquery.openx.js` another event `Impression-Clickable` which should be triggered whenever a new banner is loaded and displayed, for which we can track a `Click`. So if it contains an `a[href]`?

### DIFF
--- a/client-vendor/after-body/jquery.openx/jquery.openx.js
+++ b/client-vendor/after-body/jquery.openx/jquery.openx.js
@@ -78,6 +78,9 @@
 
         if (hasContent) {
           trackEvent('Banner', 'Impression', 'zone-' + zoneId);
+          if ($element.is(':visible') && $element.find('a[href]').length > 0) {
+            trackEvent('Banner', 'Impression-Clickable', 'zone-' + zoneId);
+          }
           $element.find('a[href]').on('click', function() {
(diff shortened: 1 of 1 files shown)


vogdb:
@njam please review.
Should I extract a variable from `$element.find('a[href]')`? Or we should not mix these two different events?

njam:
Let's extract! `$link`?

vogdb:
`$countableLink` ?

njam:
As discussed I think `$link` should be enough because it contains all "link" tags that we can select. Others (within iframes are not visible to us anyway).

vogdb:
I think it is ok to put a `click` listener attach into `if($link.length > 0 ...)`. Hdyt?
```js
          if ($element.is(':visible') && $link.length > 0) {
            trackEvent('Banner', 'Impression-Clickable', 'zone-' + zoneId);
            $link.on('click', function() {
              trackEvent('Banner', 'Click', 'zone-' + zoneId);
            });
          }
```
@njam ?

njam:
Makes sense, only in that case we expect to be able to track a click..

vogdb:
@njam please review.

njam:
lgtm!

PR #1627, opened 2015-02-05
Backslashes in emoticon codes are escaped in markdown

dlondero:
Replaces #1618


dlondero:
Closing as CI on #1618 seems to work again.

PR #1628, opened 2015-02-05
mongoClient find() docu

alexispeter:
To avoid possible confusion? The same docu is present in the constructor of CM_PagingSource_MongoDb, but it should probably (especially) also be found on the piece of code that actually exhibits this particular behaviour.


alexispeter:
please review @tomaszdurka

tomaszdurka:
Mhmm, ok then :panda_face:

alexispeter:
You sound sceptical...

tomaszdurka:
Now thanks to Alexis' explanation everything is crystal clear.

PR #1633, opened 2015-02-10
do not test query-batchSize when using default batchSize

alexispeter:
Fixes `PHP Fatal error:  Cannot use object of type stdClass as array in /home/travis/build/alexispeter/CM/tests/library/CM/MongoDb/ClientTest.php on line 158` occurring on boxes running  mongodb driver v1.6.x


alexispeter:
please review @njam

njam:
lgtm

PR #1636, opened 2015-02-12
Clone Request before modifying in Page-Response

njam:


njam:
@dlondero @tomaszdurka please review

We already clone the Request in `CM_Http_Response_Abstract`, but not in the Page Response. The idea is that the Response should not modify the passed Request.

dlondero:
@njam lgtm!

tomaszdurka:
So this modification is needed for all kind for internal rewrites and redirects we do.
Maybe it would be indeed good to rebuild request rather than alter it?

Apart from "deep" thoughts looks good.

njam:
..maybe there should rather be an immutable *HTTP Request*, and then an additional *Application Request* (for routing)...
Anyway I think for now it's sufficient, safer and more consistent to clone the request before modifying it.

PR #1638, opened 2015-02-12
Mark all exceptions in http-request constructor as WARN

njam:
@tomaszdurka @dlondero please review

There's no simple way to mock the response (because of static methods), so I'm using an actual resource-response.


dlondero:
@njam lgtm!

tomaszdurka:
Looks good!

PR #1641, opened 2015-02-14
Adjust <select> width in fancySelect

njam:
screen shot 2015-02-14 at 14 33 13


njam:
@christopheschwyzer @vogdb is this the right thing to do?

vogdb:
If it wasn't some special trick, which I'm not aware of, then it is a good change.

christopheschwyzer:
I can't remember if and why this was necessary. Makes more sense now. Lgtm.

PR #1644, opened 2015-02-17
Pass ViewResponse to Form::prepare()

njam:


njam:
Paired with @tomaszdurka

PR #1645, opened 2015-02-18
Enhance CM_Jobdistribution_Job_Abstract::_runMultipleWithoutGearman()

dlondero:
It should behave in the same way in tests and when running real code.

```php
$params = CM_Params::encode($params, true);
$params = CM_Params::factory(CM_Params::jsonDecode($params), true);
```

Maybe also introduce `CM_Jobdistribution_Job_Abstract::encodeParams()` and `CM_Jobdistribution_Job_Abstract::decodeParams()` to avoid duplication.

### DIFF
--- a/library/CM/Jobdistribution/Job/Abstract.php
+++ b/library/CM/Jobdistribution/Job/Abstract.php
@@ -123,8 +123,7 @@ abstract class CM_Jobdistribution_Job_Abstract extends CM_Class_Abstract {
     protected function _runMultipleWithoutGearman(array $paramsList) {
         $resultList = array();
         foreach ($paramsList as $params) {
-            $params = $this->_encodeParams($params);
-            $params = $this->_decodeParams($params);
+            $params = $this->_decodeParams($this->_encodeParams($params));
             $resultList[] = $this->_executeJob($params);
         }
         return $resultList;
--- a/library/CM/Jobdistribution/Job/Abstract.php
+++ b/library/CM/Jobdistribution/Job/Abstract.php
@@ -61,7 +61,7 @@ abstract class CM_Jobdistribution_Job_Abstract extends CM_Class_Abstract {
         });
 
         foreach ($paramsList as $params) {
-            $workload = CM_Params::encode($params, true);
+            $workload = $this->_encodeParams($params);
             $task = $gearmanClient->addTask($this->_getJobName(), $workload);
             if (false === $task) {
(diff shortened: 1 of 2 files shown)


dlondero:
@njam please review. Was this your idea?

alexispeter:
I'm curious as to what the motivation behind this is. Did something break or is this just a random improvement? :-)

dlondero:
During tests Gearman jobs are run immediately and params are passed (by `->queue()`) to the job without any encoding/decoding process as it happens when a job is really run in background. This could lead to errors in production even with a green build (i.e. passing an object we're not able to recreate in the job). We then want to have the same behaviour tests/prod.

@njam please add more details if needed

njam:
Looks very good.

IMO in `_runMultipleWithoutGearman()` we could inline the encoding-decoding..

@alexispeter you wanna review too?

njam:
Maybe add a test to `CM_Jobdistribution_Job_AbstractTest` to make sure un-serializable objects are omitted (or lead to error?)?

dlondero:
Any suggestion of a CM object to use?

alexispeter:
You could use a stdClass with a closure in it, as closures aren't serialisable. 
```php
$foo = new stdClass();
$foo->closure = function() {
    echo 'bar';
};
```

dlondero:
Ah ok, makes sense. I was thinking to a resource (as they're not serialisable as well) but I thought to look for a CM object for the test.

dlondero:
Something like this? /cc @alexispeter @njam

alexispeter:
Sorry there was a misunderstanding. I thought you were talking about serialising objects but turns out you want to jsonEncode them. According to http://php.net/manual/en/function.json-encode.php the only values not jsonEncodable are resources.

njam:
We don't need to have an unserializable object for this, we could just check that the json serialization *is* happening? E.g. send an instance of any class, then check that on the receiver side it's not an instance of this class anymore?

dlondero:
@njam please review

njam:
Hmm why is `_execute()` returning boolean?
Shouldn't we assert the parameter passed in this method?

dlondero:
Well AFAIK the callback should return `TRUE` or `FALSE` depending if the argument passes verification or not. Or would you only check it's a `CM_Params` object?

njam:
as discussed let's use Mocka for readability

dlondero:
@njam please review

njam:
looks good!

PR #1637, opened 2015-02-18
Splittest::_getVariationFixture has wrong variations cached

alexispeter:
As seen in a recent splittest.
Similar problem was solved here: https://github.com/cargomedia/CM/pull/1450

This is APC-cached, so difficult to debug in production. Restarting the webservers flushed the caches, and the variation was assigned correctly.

The splittest was flushed before the wrong assignment. Maybe due to the flushing the old variation was still assigned?

cc @NicolasSchmutz

### DIFF
--- a/library/CM/Model/Splittest.php
+++ b/library/CM/Model/Splittest.php
@@ -182,6 +182,15 @@ class CM_Model_Splittest extends CM_Model_Abstract implements CM_Service_Manager
 
     /**
      * @param CM_Splittest_Fixture $fixture
+     * @return string
+     */
+    protected function _getCacheKeyFixture(CM_Splittest_Fixture $fixture) {
+        return CM_CacheConst::Splittest_VariationFixtures . '_id:' . $fixture->getId() . '_type:' . $fixture->getFixtureType()
+        . '_splittestCreated:' . $this->getCreated();
+    }
+
+    /**
+     * @param CM_Splittest_Fixture $fixture
      * @param float|null           $weight
      * @throws CM_Exception_Invalid
      */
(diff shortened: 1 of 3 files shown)


alexispeter:
please review @fauvel

library/CM/Model/Splittest.php
```diff
@@ -182,6 +182,15 @@ protected function _getContainingCacheables() {
 
     /**
      * @param CM_Splittest_Fixture $fixture
+     * @return string
+     */
+    protected function _getCacheKeyFixture(CM_Splittest_Fixture $fixture) {
+        return CM_CacheConst::Splittest_VariationFixtures . '_id:' . $fixture->getId() . '_type:' . $fixture->getFixtureType()
+        . '_splittestCreated:' . $this->getCreated();
```
fauvel: Unfortunately, this will not work. We are caching, for each fixture (user/browser), the entire list of the variations assigned for all split tests. So you should not use data from any single split test in the cache key. Please remove `_splittestCreated`, and make this method static if you want to keep it.

library/CM/Model/Splittest.php
```diff
@@ -229,7 +238,7 @@ protected function _getVariationFixture(CM_Splittest_Fixture $fixture) {
                 CM_Db_Db::insert('cm_splittestVariation_fixture',
                     array('splittestId' => $this->getId(), $columnId => $fixtureId, 'variationId' => $variation->getId(), 'createStamp' => time()));
                 $variationListFixture[$this->getId()] = $variation->getName();
```
fauvel: Use the creation date of the split test in the array key instead.

alexispeter:
please review @fauvel

fauvel:
I would have imagined something more simple, like:
```php
$variationListFixture[$this->getId() . '-' . $this->getCreated()] = $variation->getName();
```
but ok, lgtm!

alexispeter:
I know, but there is no way to clean up the cache-entry with this approach, (Besides scanning array-keys).

PR #1646, opened 2015-02-18
Adjust FormField::prepare() signature

njam:


njam:
Paired with @mariansollmann

PR #1648, opened 2015-02-18
Fix floatbox lets page expand beyond viewport

christopheschwyzer:
@vogdb This is the fix for the issue @njam mentioned today!


christopheschwyzer:
@vogdb please review
@njam fyi

njam:
Ah I remember .. this will not work for pages with a width > 100% (like myspace...).
Since we don't use it I would say lgtm!

vogdb:
It's my lgtm! =)
Verified on iOS8/iPhone5s.

PR #1649, opened 2015-02-19
Fix vertical middle of "remove" icon position of Select2.

vogdb:


vogdb:
IMO it would be much cleaner to have this rule in `select2.css` but it's the original css file of the library, so it's kind of a bad style to change it, although Select2 lib is going to receive a major change soon. The version we use is 3.5.2 and it is the last stable. The new version is going to be 4.0.0 and it brings a lot of inner refactoring and breaking changes. So, we can put our rule into `select2.css` as an exception of this occasion.

layout/default/css/jquery.select2.less
```diff
@@ -15,7 +15,8 @@
       font-size: 12px;
 
       .select2-search-choice-close {
-        top: 6px;
+        top: 50%;
+        margin-top: -6.5px;
```
christopheschwyzer: You should be able to use `transform: translateY(-50%);` here. Unfortunately the image background positioning is off in the first place. I would suggest to not fix this at all, and wait for the next release.

vogdb:
Lets not wait for the next release cause it can take a long time and we can have a lots of problems with upgrading to it. I replaced `margin-top` with `translate`. If to make the solution of `background-position` which element's position should I change?

vogdb:
@christopheschwyzer ?

christopheschwyzer:
Trust you on this. Code lgtm.

PR #1651, opened 2015-02-19
Make Tracking::getTrackings() public

njam:
@fauvel please review

I added `CMService_GoogleAnalytics_MeasurementProtocol_Client::getRandomClientId()` so we can wrap this logic here?


fauvel:
Looks good so far!

PR #1650, opened 2015-02-19
ElementQuery media polyfill.

vogdb:
Experimental issue to try http://www.smashingmagazine.com/2013/06/25/media-queries-are-not-the-answer-element-query-polyfill/.

### DIFF
--- a/layout/default/css/tableField.less
+++ b/layout/default/css/tableField.less
@@ -1,9 +1,5 @@
 .tableField {
-  padding: 5px 0 5px 5px;
-
-  &:first-child {
-    border-width: 0;
-  }
+  padding: 6px;
 
   > .label {
     box-sizing: border-box;
--- a/layout/default/css/tableField.less
+++ b/layout/default/css/tableField.less
@@ -15,7 +15,7 @@
     }
   }
 
-  &[min-width="250px"] {
+  &[min-width="@{breakpointSmall}"] {
     > .label {
       width: 100px;
       float: left;
--- a/layout/default/css/tableField.less
+++ b/layout/default/css/tableField.less
@@ -1,29 +1,28 @@
 .tableField {
   padding: 5px 0 5px 5px;
-}
 
-.tableField:first-child {
-  border-width: 0;
-}
-
-.tableField > .label {
-  box-sizing: border-box;
-  font-weight: bold;
-  color: @fontColorStrong;
-
-  @media (min-width: @breakpointSmall) {
-    width: 100px;
-    float: left;
-    text-align: right;
+  :first-child {
+    border-width: 0;
   }
-}
 
-.tableField > .label:after {
-  content: ":";
-}
+  > .label {
+    box-sizing: border-box;
+    font-weight: bold;
+    color: @fontColorStrong;
 
-.tableField > .value {
-  @media (min-width: @breakpointSmall) {
-    margin-left: 110px;
+    :after {
+      content: ":";
+    }
+  }
+
+  &[min-width="250px"] {
+    > .label {
+      width: 100px;
+      float: left;
+      text-align: right;
+    }
+    > .value {
+      margin-left: 110px;
+    }
   }
 }
--- a/layout/default/css/tableField.less
+++ b/layout/default/css/tableField.less
@@ -1,7 +1,7 @@
 .tableField {
   padding: 5px 0 5px 5px;
 
-  :first-child {
+  &:first-child {
     border-width: 0;
   }
 
@@ -10,7 +10,7 @@
     font-weight: bold;
     color: @fontColorStrong;
 
-    :after {
+    &:after {
       content: ":";
     }
   }
--- a/client-vendor/after-body/css-element-queries/ElementQueries.js
+++ b/client-vendor/after-body/css-element-queries/ElementQueries.js
@@ -1,0 +1,334 @@
+/**
+ * Copyright Marc J. Schmidt. See the LICENSE file at the top-level
+ * directory of this distribution and at
+ * https://github.com/marcj/css-element-queries/blob/master/LICENSE.
+ */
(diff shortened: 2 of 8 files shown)


vogdb:
@christopheschwyzer please look and test how it looks. Is it acceptable for us?

christopheschwyzer:
Looks good. I like it. 
- I would use the same media-query breakpoint (@breakpointSmall). Also I noticed that this variable should be defined in CM. 
- We should use the same behavior in _form.less_ for the form layout

vogdb:
> I would use the same media-query breakpoint (@breakpointSmall). Also I noticed that this variable should be defined in CM.

So, should I define it if it is not defined?

christopheschwyzer:
> Should I define all the breakpoint variables? 

Yes, we should define the variables in CM, if they are used in CM..

christopheschwyzer:
IMO you can move all the breakPoint variables...

vogdb:
Hm. According to http://leafo.net/lessphp/changelog.html
> - Selector interpolation works inside of brackets in selector

works since 0.4. We have 0.3.9. So, my question is. Can we update `lessphp` to 0.4.0? @njam please answer.

njam:
I think we should update!
There's even 0.5.0 on github: https://github.com/leafo/lessphp/releases

You wanna try?

vogdb:
Yes, please!

vogdb:
Should I just update the dependency on composer.json of CM to?
```json
"leafo/lessphp": "0.5.*",
```

njam:
Yes, and also test that the CSS in SK still compiles properly!
Also let's switch to `~0.5.0'` - https://getcomposer.org/doc/01-basic-usage.md#next-significant-release-tilde-and-caret-operators-

vogdb:
@njam @christopheschwyzer please review.

vogdb:
@christopheschwyzer Should I remove an issue in SK for removal of `breakpoint` variables there because now we have them in CM.

layout/default/css/tableField.less
```diff
@@ -1,29 +1,28 @@
...
+  > .label {
+    box-sizing: border-box;
+    font-weight: bold;
+    color: @fontColorStrong;
 
-  @media (min-width: @breakpointSmall) {
-    width: 100px;
-    float: left;
-    text-align: right;
+    :after {
```
christopheschwyzer: You need "&" for this to work

layout/default/css/tableField.less
```diff
@@ -1,29 +1,28 @@
 .tableField {
   padding: 5px 0 5px 5px;
-}
 
-.tableField:first-child {
-  border-width: 0;
-}
+  :first-child {
```
christopheschwyzer: You need "&" for this to work

christopheschwyzer:
> Should I create an issue in SK for removal of breakpoint variables

Yes - or you can do it in https://github.com/cargomedia/sk/issues/1218

vogdb:

christopheschwyzer:
Can you also remove 
```
  &:first-child {
    border-width: 0;
  }
```
Doesn't do anything. And can you use
```
.tableField {
 padding: 6px;
}
```
so it's streamlined with form fields...

vogdb:
Yes, Master!

vogdb:
@christopheschwyzer please look

christopheschwyzer:
Lgtm!
Restarted tests.

vogdb:
@njam please review my last commit @3a4bd50

njam:
- function.lessVariable.php: Why did you remove the check of the return value of `preg_match`? I would keep it for clarity?

vogdb:
Because now the lessphp lib throws an exception in case if a variable is not defined. Previously it didn't throw and the check did this throw instead of lessphp. Should I return the check?

njam:
I think it should still be there, as it's a different error condition: Basically if the regexp doesn't match. This could happen for different reasons - e.g. if the returned CSS is not what we expect.

vogdb:
@njam please look.

njam:
Looks good.

Did you do some benchmarking to understand the performance impact of adding this library?

vogdb:
Nope. Will do. Do we have a company standard for benchmarking?

njam:
This case is quite tricky. I would maybe measure the time required to rotate/resize the window on a slow phone?

vogdb:
lessphp 5.0

|Type of device | Load time | Rotate time |
|--------------------|---------------|-----------------|
| Emulator 4.1.2 | 0:50, 0:56, 2:25, 2:02. After browser restart 0:30, 0:34 | ~0:05-0:06 |
| LG p500 4.0.4  | 0:30, 0:33, 0:35 | ~0:05-0:06 |


lessphp 3.9

|Type of device | Load time | Rotate time |
|--------------------|---------------|-----------------|
| Emulator 4.1.2 | 1:34, 1:40 After browser restart 0:33, 0:30 | ~0:05-0:06 |
| LG p500 4.0.4  | 0:35, 0:30 | ~0:05-0:06 |

Conclusion: Android devices are so unpredictable, slow and unstable that it is hard to measure anything serious there. Althought with lessphp 5.0 it seems to me that I saw a more continuous presence of this state:

njam:
Just to be sure: I meant what the impact of the *ElementQueries* plugin is.
I guess the 3.9 version is without and the 5.0 version is with ElementQueries?

If so then it looks like it doesn't make much difference, and I'd say it's good to go?

vogdb:
> I guess the 3.9 version is without and the 5.0 version is with ElementQueries?

Yep.

vogdb:
Ok. @christopheschwyzer @njam if you don't have any more comments then I'm merging this.

christopheschwyzer:
GoGoGo

njam:
⁠

PR #1653, opened 2015-02-21
Convert CM_Elasticsearch_Client into a service

tomaszdurka:


tomaszdurka:
@njam please review

njam:
- CM_Elasticsearch_Cluster: Should be service-manager-aware. This way it will get the manager injected automatically and you can use it to access the debug service.
- getClients: Rename to getClientList?
- query: You're three times `getRandomClient()`, so every time a different client?
- Test files: Use `CMTest_TH::getServiceManager()` instead of singleton one - the idea is to have a single place for accessing service manager in tests, so it can be refactored easily in the future.

tomaszdurka:
- done (as discussed: I will replace all those global service manager usages after all the singleton-services refactoring)
- done (although I feel better when using human names like getClients :) )
- done
- done (I would see it a bit different, with servicemanager being created before test, but this can be adjusted later)

tomaszdurka:
@njam Please review
- done (as discussed: I will replace all those global service manager usages after all the singleton-services refactoring)
- done (although I feel better when using human names like getClients :) )
- done
- done (I would see it a bit different, with servicemanager being created before test, but this can be adjusted later)

njam:
lgtm

Issue #1572, opened 2014-12-31
Use "runtime" instead of "apc(u)" for local-cache when in CLI mode?


njam:
Problem: We experience APC(u) being unstable/problematic in high memory usage situations. Segfaults are usually the result.

Idea: Instead of using the APC-storage for `LocalCache` we could use the Runtime-storage (php array-based).

Possible concerns:
- PHP array storage might use more memory than APC storage. To conclude we should test what's the difference. Also we should determine how much local-cache memory our CLI processes use.

@ppp0 @kris-lab @tomaszdurka wdyt?

tomaszdurka:
You want to use runtime cache instead of apc only for cli, right?
Additionally we do not limit apc max memory within configuration? Or it's segfaulting when hitting it's own limit...?

njam:
> You want to use runtime cache instead of apc only for cli, right?

Yes

> Additionally we do not limit apc max memory within configuration? Or it's segfaulting when hitting it's own limit...?

Do you mean apcu's `apc.shm_size` setting? We set it to [1.5GB here](https://github.com/cargomedia/puppet-cargomedia/blob/e4c3d889e3db3a941a61843d2eb957b432d48c73/modules/cargomedia/files/hiera-data/datacenter/softlayer.json#L4).

njam:
@alexispeter maybe for you? Let me know so we can discuss..

alexispeter:
Would it also be an option to not cache at all in cli-mode? If we're only going to use runtime-cache, a cli process will only hit the cache when accessing entries created by the process itself, which I don't think a lot of tasks do.

njam:
We have quite a few entries in APC that need to be cached even for CLI: Translations, Splittests, Usertext, Classchildren. Some of these are heavy to calculate so we should cache them!

njam:
As discussed please check memory usage for APC and "Array" for ~100MB in ~1000 keys. Based on that we can decide how to proceed.

alexispeter:
##### Memory Usage for 1000 keys à ~0.1MB
**apc:** 145MB
**runtime:**125MB

There is a threshold at ~309MB, starting from which APC will not use more memory for increasing data. Runtime-cache memory usage on the other hand will continue to grow linearly.

njam:
So it looks like "runtime" can store more efficiently even, but of course it doesn't do mmapping. This might be a problem if *lots* of data is put into local cache.

@cargomedia/devops before we proceed I'd like to get everyone's opinion here.
[...]

PR #1655, opened 2015-02-21
FB-130 Convert caches into services

tomaszdurka:


tomaszdurka:
@njam plaese take a look

Cache usages are still based on `CM_Cache_Local/Shared` to not oversize this pull-request.

I am not sure about the `CM_Cache_Service` (which could be renamed ofc). It contains all the cache storages plus two "meta-caches" (local and shared). Maybe having those meta-caches as separate services (outside of cache service) would make it be cleaner and cross a clear line between those.

njam:
- ServiceManager::getCache: Call is `getCaches()`, similar to `getDatabases()`
- Cache service: Do we really need to make memcache, apc etc available here? Wouldn't it be better to only provide getters for "local", "shared" and "file" cache, and remove `CM_Cache_Local` and `CM_Cache_Shared`? -> see also https://github.com/cargomedia/CM/issues/1572
- When finished let's do some performance testing. These code paths are used a lot, any change here can impact performance!

PR #1658, opened 2015-02-22
FB-294 Convert CM_Adprovider into a service

tomaszdurka:


tomaszdurka:
@dlondero please review

tomaszdurka:
@dlondero please review

tomaszdurka:
Sorry for amound of commits, this pull-request is about moving CM_Adprovider into a service. Apart from basic creation of the service I've also moved all adprovider component into folder/module CM_Adprovider, resulting in a lot of renaming.

I've also added Factory method for creating clients and adapters - mostly for config/service registration. 

*Note: CM_Site_Abstract still uses CM_Adprovider as config key (not new CM_Adprovider_Client) as it's concerning whole Adprovider topic not only client class. Generally would be glad to move away from combining configuration keys with class names.*

dlondero:
@tomaszdurka lgtm. I was wondering about the move of a lot of stuff but then I read your comment here.

dlondero:
One thing: how about using just one way to define arrays in the config? We now have mixed `[]` and `array()`.

tomaszdurka:
Yes we should probably have some guidelines for this or simply allow everyone to decide on their own. I use a lot of array() in config - as it was used in other places. Place I use short ones are mostly if there is an array of arrays: `array([], [])`. It's mostly common sense...

dlondero:
I guess we already allow everyone to decide, just open `default.php` and you can see all kinds of different styles and combinations :P

PR #1659, opened 2015-02-22
Convert Gearman into JobManager service

tomaszdurka:


tomaszdurka:
Current implementation is very minimalistic. It seems like creation of this service is forced. Generally approach to have instance requesting some global services interanally is quite problematic. Here job is the one.

I propose to flip around service and job and put service on top of jobs. Would mean to move queue, execute and all other method using a service to JobManager instance.

I propose it in this pr.

tomaszdurka:
@njam have you tried to install gearman on travis machine? Are things like starting your own new service is even possible on travis? Thanks.

njam:
http://lmgtfy.com/?q=gearman+travis :P

It's definitely possible (root can install anything).
The more we install the slower our build will become. Maybe we should check how long a jenkins-based build would run (based on our vagrant box)?

tomaszdurka:
@njam Wdyt about skipping these tests on travis? As temporary solution...

njam:
I don't think that's a good idea: It will be almost like having no tests, regressions are very likely, and we will notice them on production.

Since travis is running ubuntu I can imagine they actually have recent gearmand packages?

tomaszdurka:
I would like to install gearman-job-server on our travis machine, but I am not sure if the code I've added does the job (`sudo apt-get install gearman-job-server`). 

All I get are exceptions like:
```
GearmanException: Failed to set exception option in /home/travis/build/cargomedia/CM/library/CM/Jobdistribution/JobWorker.php on line 74
```
Which I've identified to be caused by connection to gearman server problem.

@njam I would like to verify if the gearman-job-server is installed correctly on our travis instance. I know you've been doing some travis debugging, so here are my questions:

- Is there a way to connect to travis instance? :o
- Are there any vagrant (virtualbox) boxes I could spin up and reproduce test env locally?
- Any other ways?

Thanks.

njam:
Travis doesn't provide VM access publicly anymore, but I've read on the Internet that you can reach out to their support and they might provide you with an image!

I think they're using stable Ubuntu - so maybe just spin up a default Ubuntu vagrant image instead, and debug there?
I could imagine you would need to do something like:
```
sudo su -c 'echo "ENABLED=\"true\"" >> /etc/default/gearman-server'
```
Maybe also look for other builds that have gearmand successfully installed?

njam:
@tomaszdurka you did some progress here?

tomaszdurka:
Closing in favor of #2513 .

PR #1661, opened 2015-02-23
Convert CM_Option into a service

tomaszdurka:


tomaszdurka:
@dlondero please review

dlondero:
If we name tables after the class name should `cm_option` become `cm_options`? Rest looks good.

tomaszdurka:
We usually name as you said when describing models, but models refer to a single object. Here `CM_Options` refer to collection of options so I believe naming is fine. `CM_Options` could be maybe even a paging - `CM_Paging_Option_All`, but that would change logic behiing it a lot and be overkill atm imo.

tomaszdurka:
I've readdd CM_Option class for backwards compatibility I think it would be nice to remove with new major realease only.

tomaszdurka:
@dlondero wdyt?

dlondero:
@tomaszdurka please merge upstream

tomaszdurka:
Done.

dlondero:
@tomaszdurka looks good

PR #1660, opened 2015-02-23
Add support for navigation to page anchors

fauvel:


fauvel:
@njam please review

tomaszdurka:
Looks good, only a few things:

**Is it neccesary to have this anchor navigating feature in Layout/Abstract.js?**

**Fragment in request should be nullable.** You could compare fragment to null (not empty string) when generating url which is cleaner.
Only problem I see is with extracting fragment from request using `parse_url('...', PHP_URL_FRAGMENT)`:
  - `http://foo/bar#bar` fragment should be `'bar'`, is `'bar'`
  - `http://foo/bar` fragment should be `null`, is `null`
  - `http://foo/bar#` fragment should be `''`, is `null`

Introducing some additional trailing hash check could help. Wdyt?

tomaszdurka:
When I think about it, maybe there should be nothing like empty string hash? I would just allow null for `setFragment()` then.

njam:
Just talking with @vogdb who needs the same feature.
I don't think the hash should reach the server (it doesn't in HTTP). We could just keep it on the client side, and concatenate it with the URL coming back from the server before calling `replaceState()`?

fauvel:
> I don't think the hash should reach the server (it doesn't in HTTP).

I agree that the hash doesn't **ought to** reach the web server. As a matter of fact (and as a best practice), most web browsers won't pass it along with the request when they handle the navigation by themselves (URL bar / HTML links). But – as a side note – there's nothing at the protocol level that doesn't allow you to do so; in fact, using the implementation I've suggested for our ajax navigation, you actually get the fragment in `$_SERVER['REQUEST_URI']` (at least in my dev box, with nginx and our php-fpm config).

So I also agree we should remove support for fragments in request URLs on the server side.

To handle it on the client side, we could (a) remove the fragment from the request URL before loading the page in the ajax navigation, (b) check whether the URL of the page response matches the request URL (without fragment), and (c) in this case – and in this case only! – re-add the fragment to the URL of the page response before storing it in the navigation history, and jump to the anchor.

Is this what you have in mind?

vogdb:
He definitely meant that. Lets not wait any minute longer and finish this issue! =)

fauvel:
@njam @tomaszdurka @vogdb please review!

library/CM/App.js
```diff
@@ -42,8 +42,8 @@ var CM_App = CM_Class_Abstract.extend({
       return view;
     }
     return _.find(this.views, function(view) {
-      return view.hasClass(className);
-    }) || null;
+        return view.hasClass(className);
```
vogdb: indent typo?

fauvel: indent fix :)

vogdb:
- Should we use `responsePath` instead of `pathResponse` if we use the names as `modulePath`, `templatePath` and etc. in other places?
- Lets be more js-style!
```js
if ('' !== window.location.hash) {
```
with
```js
if (window.location.hash) {
```

fauvel:
@njam @tomaszdurka good to merge?

vogdb:
lgtm!

njam:
- CM/Layout/Abstract.js: Now that we check if the server changed the fragment, we can call `replaceState()` only if necessary?
 - Maybe introduce a boolean variable to clarify on this - `responsePathChange`?
- CM/Layout/Abstract.js: Move the anchor-scrolling into `_onPageSetup()`?
- CM/Layout/Abstract.js: You renamed "fragment" to "path". While I agree it's probably the better choice I think we should keep it consistent. At the moment anything after the domain and port we call "fragment". I'd suggest to stick with it?

fauvel:
> Now that we check if the server changed the fragment, we can call `replaceState()` only if necessary?

Are you sure that the path passed to `loadPage` is in the navigation history already?

vogdb:
Ok. If we decided to be perfect then lets do a replacement:
```js
      var anchor = document.getElementById(hash);
      if (!anchor) {
        var anchorList = document.getElementsByName(hash);
        if (anchorList.length) {
          anchor = anchorList[0];
        }
      }
```
with
```js
var anchor = $('#' + hash)[0] || $('[name=' + hash + ']')[0];
```
or
```js
var anchor = $('#' + hash).get(0) || $('[name=' + hash + ']').get(0);
```
whatever you prefer =)

njam:
> Are you sure that the path passed to loadPage is in the navigation history already?

I think so, but maybe it's safer and more robust to keep it!

fauvel:
@vogdb I always think about code injection when I see things like that: `'[name=' + hash + ']'` – I would rather keep the dom implementation. @njam good to merge?

njam:
agree - http://youmightnotneedjquery.com/ :D

Should we test this on some *edge case devices* first? - https://github.com/cargomedia/sk/wiki/Supported-Browsers

@vogdb maybe you can help with that?

vogdb:
>  I always think about code injection when I see things like that:
> agree - http://youmightnotneedjquery.com/ :D

So, why to include jQuery in CM codebase? if we have a library then we should use it on 100% otherwise don't use it at all.
In this case I don't see how the injection can be done properly. The processing of hash happens only on the client side.

vogdb:
> @vogdb maybe you can help with that?

Will test on Desktop browsers + my readl Android + iPhone simulators.

njam:
Agree there's no real injection possible. It's just a sizzle selector, and we only use it to get the DOM element!

fauvel:
I don't see a big security threat here either, and I have absolutely nothing against jQuery! In the worst case, someone could forge a link to our website with a hash that would break the jQuery selector, that wouldn't be the end of the world. But it just hurts the eye to see that we're not escaping user input. And escaping it would make the whole construct so complicated, that I would prefer to stay with the straight-forward, fuzzy-free DOM approach. You really don't need the whole jQuery overhead to handle such a simple thing as the selection of an anchor. There's also quite a few jQuery implementation of the scrolling to the anchor itself, but `scrollIntoView()` is really cross-browser compatible afaik. @vogdb if it works in all browsers we're supporting, I would advocate to keep it as is. But I agree it's only a matter of taste.

vogdb:
Ok. Then lets not prolongate the issue! Lgtm!

vogdb:
Oh. Wait. Need to test it carefully.

fauvel:
Yes, I'd rather be more productive and discuss less on such details, too, but maybe we'll learn something from your testing?

vogdb:
You should test on the edge Android devices that we have in the office.

vogdb:
Ok. Works fine. Unfortunately I don't have a device with Android Browser =( My Android device is too new and already has replaced it with Mobile Chrome.

fauvel:
So good to merge? @njam agree?

njam:
So we should still test it on Android?
@NicolasSchmutz could you do that?

NicolasSchmutz:
It took a while but it looks good on all Android devices we have.

fauvel:
Tagged 1.104.1

PR #1662, opened 2015-02-23
Introduce CM_Util_NamedArgs for easier services configuration

tomaszdurka:


tomaszdurka:
@alexispeter please review
@njam fyi

Any objections? Disadvantages?

alexispeter:
I suppose this is to enhance readability of the config?

tomaszdurka:
Yes, that's the goal. We can use named argumetns in config without this - as I do already. This adds some kind of validation to it.

alexispeter:
will review it first thing tomorrow

njam:
Good idea!
Did you research/test the impact of using reflection in production code?

alexispeter:
* library/CM/Service/Manager.php
 + line 203: Including `$method` in the if-clause is unnecessary since getMethod() always returns an instance of ReflectionMethod or throws if it fails.
* resources/config/default.php
 + lines 146 + 159: param `adapter` should be `adapterClassName`

tomaszdurka:
- ok
- ok

Wdyt about removing `isNamedArgs` condition and forcing to always passed named arguments?

tomaszdurka:
Regarding the performance it will be run just couple of times and:
- single isNamedArgs run for 3 arg array takes 0.0214ms
- single matchNamedArgs for 3 arg array with 3 arg method takes 0.01806ms

I would make it standard and remove isNamedArgs check. Agree?

alexispeter:
agree to removing isNamedArgs

tomaszdurka:
@alexispeter please review

alexispeter:
* library/CM/Bootloader.php
 + line 180 Using registerInstance() instead of register() might lead to problems regarding forking and service-manager reset so I'd suggest to switch back to register() until we have a better solution for this fork-problem.
* tests/library/CM/Model/SplittestTest.php
 + line 144 use registerInstance() for conformity?

tomaszdurka:
@alexispeter please review

alexispeter:
lgtm

PR #1664, opened 2015-02-23
Convert CM_Debug into a service

tomaszdurka:


tomaszdurka:
@dlondero please review

dlondero:
@tomaszdurka lgtm

njam:
For such generic services it might make sense to introduce an interface - so that it can be replaced easily?
Maybe `CM_Service_Debug_DebugInterface`?

PR #1665, opened 2015-02-23
Convert CM_Stream_Video into a service

tomaszdurka:

PR #1666, opened 2015-02-24
Convert CM_Stream_Message into a service

tomaszdurka:
### Breaking Notice
Breaks in-code public api (class names). 
Breaks public cm commands:
- `stream start-message-synchronization` -> `message-stream start-syncrhonization`.


tomaszdurka:
@njam please review

tomaszdurka:
Actually postpone this review. I meant to ask for review of #1665 first.

tomaszdurka:
@njam This code is ready to review. I would be really glad to close all those service-related issues which have been started.

tomaszdurka:
### Breaking Notice
This code is breaking as changes public classes and method in-code. 
Additionally cm command has changed:
`stream start-message-syncrhonization` -> `message-stream start-syncrhonization`.

njam:
Thanks for those detailed change notes!
There's a small typo in it - "syncrhon...".
Also I would suggest to move them to the ticket description - this is where I usually look if I upgrade CM in denkmal for example, wdyt?

tomaszdurka:
Done.

tomaszdurka:
Strange test failure. Haven't see it before.
```
1) CM_Paging_AbstractTest::testGetItemRandMean
Failed asserting that 1.3799999999999999 matches expected 0.90000000000000002.
```

njam:
`testGetItemRandMean` is expected to fail very rarely I think cc @fauvel

njam:
- CM_MessageStream_Service::getAdapterClass: Null check not needed
- CM_MessageStream_Service::getInstance: Remove?
- CM_MessageStream_Service::__construct: Consider moving adapter creation into a dedicated factory? Like `CM_File_Filesystem_Factory`?
- CM/RenderAdapter/Layout.php: Introduce a variable for serviceManager?
- 

Further:
- What happens to the stored clockwork entry for `CM_Stream_Message::synchronize` - we discussed this problem with Alexis, but I don't remember what the outcome was?
- Requires adjustments on other systems (infrastructure, Procfile)..
- Full change should be acceptance tested

tomaszdurka:
@njam 

- Null check is again needed as adapter might not be set - then it's disabled...
- Removed.
- Did so as in VideoStream.
- This has been simplified with use of newly introduced `getClientOptions()`.


Further:
- Well clockwork will be restarted with release. Old synchronize will be stored in Clockwork storage, but not used anymore. New synchronize will be started immediately and rerun every minute.
- Indeed, as in Breaking Notice.
- I will do testing while adjusting SK (example component and chat).

fauvel:
> Maybe mock randomness? :)

Let's introduce a service for that! ;)

njam:
- CM_MessageStream_Service: Add a test (especially for case without adapter?)

Rest looks very good

tomaszdurka:
@njam please do final check, thanks!

njam:
lgtmbnt (but not travis)

PR #1511, opened 2015-02-26
Invalidate translation caches after a few hours

fauvel:
Currently translation caches are kept for either 24h or until the next deployment:
- Static translations (without any variables - `smarty_compiler_translateStatic`): After a release the translation is requested once and written out into the "compiled" html file. It only changes with the next release, when smarty's "compiled cache" is flushed.
- Variable translations (with variables - `smarty_function_translateVariable`): The translation is _not_ written into the compiled file directly, but requested from the translation system every time. There it is cached in APC cache for 24h, after which it will be requested from the database again.

This is problematic because when changing any translation the result will be visible only after a long period of time.

Idea: Make sure that all involved caches expire after a defined period of time. We could make that timeframe configurable, and default to something like 1h.
- APC is no problem (just set TTL).
- Smarty compiled templates we need to investigate how to expire them.
- We should make sure not all caches drop at the same time to avoid big performance  impacts. Randomize TTLs?


fauvel:
Benchmark follows asap.

fauvel:
Benchmark of the index page, after the first rendering, with (dozens of) hard-coded translations in the compiled templates (=before):
```
ab -n 100 -c 10 http://www.swing.dev/
This is ApacheBench, Version 2.3 <$Revision: 1554214 $>
Copyright 1996 Adam Twiss, Zeus Technology Ltd, http://www.zeustech.net/
Licensed to The Apache Software Foundation, http://www.apache.org/

Benchmarking www.swing.dev (be patient).....done


Server Software:        nginx
Server Hostname:        www.swing.dev
Server Port:            80

Document Path:          /
Document Length:        107167 bytes

Concurrency Level:      10
Time taken for tests:   41.331 seconds
Complete requests:      100
Failed requests:        1
   (Connect: 0, Receive: 0, Length: 1, Exceptions: 0)
Total transferred:      10731569 bytes
HTML transferred:       10718569 bytes
Requests per second:    2.42 [#/sec] (mean)
Time per request:       4133.093 [ms] (mean)
Time per request:       413.309 [ms] (mean, across all concurrent requests)
Transfer rate:          253.56 [Kbytes/sec] received

Connection Times (ms)
              min  mean[+/-sd] median   max
Connect:        0    4  19.9      0     101
Processing:  3640 4120 248.1   4100    4624
Waiting:     3586 4063 252.8   4063    4543
Total:       3641 4124 251.6   4100    4625

Percentage of the requests served within a certain time (ms)
  50%   4100
  66%   4290
  75%   4349
  80%   4382
  90%   4453
  95%   4539
  98%   4575
  99%   4625
 100%   4625 (longest request)
```

The same, with cached translations (=after):
```
ab -n 100 -c 10 http://www.swing.dev/
This is ApacheBench, Version 2.3 <$Revision: 1554214 $>
Copyright 1996 Adam Twiss, Zeus Technology Ltd, http://www.zeustech.net/
Licensed to The Apache Software Foundation, http://www.apache.org/

Benchmarking www.swing.dev (be patient).....done


Server Software:        nginx
Server Hostname:        www.swing.dev
Server Port:            80

Document Path:          /
Document Length:        112533 bytes

Concurrency Level:      10
Time taken for tests:   41.385 seconds
Complete requests:      100
Failed requests:        96
   (Connect: 0, Receive: 0, Length: 96, Exceptions: 0)
Total transferred:      11164379 bytes
HTML transferred:       11151379 bytes
Requests per second:    2.42 [#/sec] (mean)
Time per request:       4138.500 [ms] (mean)
Time per request:       413.850 [ms] (mean, across all concurrent requests)
Transfer rate:          263.45 [Kbytes/sec] received

Connection Times (ms)
              min  mean[+/-sd] median   max
Connect:        0    0   0.1      0       1
Processing:  3662 4125 373.1   4012    5036
Waiting:     3619 4065 376.4   3929    5030
Total:       3662 4125 373.2   4012    5036

Percentage of the requests served within a certain time (ms)
  50%   4012
  66%   4237
  75%   4326
  80%   4350
  90%   5017
  95%   5034
  98%   5035
  99%   5036
 100%   5036 (longest request)
```

So that's an increase from 413.309 ms to 413.850 ms per request (+0.13%). I think we can neglect that?

njam:
Agree, we can neglect that!

fauvel:
@njam please review

njam:
I think the prefilter is not need anymore now, it was only to differentiate "static" and "dynamic" translations during template compilation?

fauvel:
It looks like it's still needed, or we'd have to make some laborious refactoring... Custom functions don't support shorthand attributes as far as I know, this triggers a `SmartyCompilerException` with the message `Syntax error in template "example.tpl" on line 1 "{translate 'Example'}" too many shorthand attributes`. See `Smarty_Internal_CompileBase->getAttributes()`. We'd have to add the `key=` fragment to all usages of `{translate ` throughout the code... But sure, it's possible. Should I do that?

njam:
Ah... let's keep the precompiler then! Good to know (shorthand attributes are useful sometimes..).

So if we release this it will introduce an expiration time for the whole language tree after 1h. I'm a bit concerned about this. Building the whole translation tree is a heavy operation (we had mysql deadlocks because of it in the recent past).
Could you benchmark how long it takes to build the tree?
Any ideas how to avoid that?

fauvel:
Update proactively in cron?

fauvel:
Er... not that easy :/ The problem is that we cannot target explicitly each web server behind the load balancer to let it update its APC cache...

fauvel:
We can at least mitigate the issue by increasing the ttl in high-load applications (we have a config option for that).

fauvel:
Another approach would be to fire an APC update with an appropriate HTTP request on a random web server at regular intervals, and hoping that we'll have hit each of them before the cache invalidates...

njam:
Hmhmhm...

In bipbip we're using the `cgi-fcgi` command line tool to issue requests to php-fpm on each app server ([code](https://github.com/cargomedia/bipbip/blob/master/lib/bipbip/plugin/fastcgi_php_fpm.rb)).
Maybe we could use a similar approach?

fauvel:
Yes, it would really be great to be able to invalidate or update the APC caches by dispatching requests to every web server behind the load balancer! The app itself would have to be "network aware" somehow (I guess you have stored IP addresses and ports in the BipBip config?), but I think it's worth the effort! I've been missing this feature quite a few times already; it would make it much more convenient to work with the local cache, and allow for some performance improvements by moving some models from Memcache to APC.

fauvel:
Should we try it out with a 6 hours TTL in prod first, and only implement a proactive cache refresh if needed?

njam:
I'm quite worried to invalidate translation caches. We've had mysql deadlocks on the translation tables in the past in situations of high concurrency when reading translations. And this currently only happens on release. During normal operations this is even riskier..

For the `cgi-fcgi` we don't need to know the network topology. We could just install something locally on every machine where the app is installed. A clockwork job could run `cgi-fcgi` in a shell to connect to the local php-fpm webserver. We would just need to configure the php-fpm port in the app config, which is a bit strange..

What do yo uthink?

fauvel:
No need to try to invalidate caches if you already know by experience that it's going to trigger a deadlock...

Firing up a cache update on every web server before it invalidates (or even in real-time, whenever a translation has been updated) sounds like a reasonable solution. Definitely a bit overkill for the nice-to-have admin usability feature we have in mind here, but I don't see any reliable alternative either...

How would you secure the system so that we don't end up with a DoS attack on the translation update endpoint? With an HTTPS connexion and some secret key stored in the private part of the app config?

njam:
We could have a separate endpoint `public/cron.php`, which we can request when using php-fpm directly.

On the webserver only `public/index.php` will be accessible ([config](https://github.com/cargomedia/puppet-packages/blob/master/modules/cm/manifests/vhost.pp#L47)).

In this `cron.php` endpoint we could possibly use a `CM_Clockwork_Manager` instance and just run `runEvents()` (which is usually ran in a loop in the clockwork daemon)?

fauvel:
Ok, the security setup looks good; so you're seeing it as a kind of private tunnel to the web servers for background jobs? I've never heard of this sort of setup before, but  it sounds great!

njam:
How about adding to `CM_Maintenance_Cli`?:
- `runOnceWebserver()` (calls `runEvents()`)
- `_registerCallbacksWebserver()`

fauvel:
@njam let's discuss what's still to do here:
- Agree on the status quo = putting static translations into the APCu cache and invalidating all of them after one hour by default?
- Should we refresh the cache proactively to avoid deadlocks, e.g. with a clockwork job running on every web server and sending HTTP requests to itself every 30 minutes by default?

fauvel:
Last open question is how to trigger the local php-fpm, resp. where to configure this best (port/socket).

njam:
Too old

PR #1670, opened 2015-02-27
CM_Jobdistribution_Job_Abstract should throw if any object passed is not instance of CM_ArrayConvertible

dlondero:
Round 2 of #1645 that we though to be easier...


dlondero:
@njam please review

njam:
looks good, restarted 1 build

PR #1671, opened 2015-02-28
Wrap "pecl install" in "travis_retry" because it fails often

njam:


njam:
@tomaszdurka @dlondero @vogdb please review

tomaszdurka:
This magic looks good. How many times it tries?

vogdb:
It's hard for me to judge whether it is ok or not =) Lgtm!
Btw why did we need the `--dev` param?

njam:
- --dev: composer changed this in recent versions, it's now default and actually shows a deprecation warning if passed...
- How many times it tries?: 3

dlondero:
@njam lgtm!

PR #1673, opened 2015-03-03
Breaking: Revert "ElementQuery media polyfill. #1650"

vogdb:
The used library has a bug that it applies itself only to the visible elements on the page after dom-ready event. The elements that became visible after some user action are not tracked by this library.


vogdb:
@christopheschwyzer @njam please review.

christopheschwyzer:
lgtm

christopheschwyzer:
Although, lessphp and breakpoint-variable changes could don't need to be reverted.

vogdb:
We will include them in the next issue with a fix of ElementQuery library.

christopheschwyzer:
Alright

njam:
ok

PR #1674, opened 2015-03-05
Make $.scrollTo($el) work with scrollables other than document

njam:
@vogdb please review


vogdb:
ok. Should it work for a case when `target` is outside of `scrollable` ?

njam:
I think that case should not be supported. Maybe we should throw?

vogdb:
sgtm

njam:
@vogdb please review

vogdb:
lgtm. What about js docu?

njam:
@vogdb you meant this? please review

vogdb:
lgtm! :D

PR #1676, opened 2015-03-05
Persist clockwork manager start-time for first-time interval-based events

alexispeter:
Adresses https://github.com/cargomedia/sk/issues/1152

### DIFF
--- a/tests/library/CM/Clockwork/ManagerTest.php
+++ b/tests/library/CM/Clockwork/ManagerTest.php
@@ -50,6 +50,7 @@ class CM_Clockwork_ManagerTest extends CMTest_TestCase {
     }
 
     public function testIntervalEventsNotPushedBackByClockworkRestart() {
+        // infinite pushback on restart problem for long intervals
         $currently = new DateTime('midnight', new DateTimeZone('UTC'));
         $lastRuntime = null;
         $process = $this->mockClass('CM_Process')->newInstanceWithoutConstructor();
@@ -57,7 +58,7 @@ class CM_Clockwork_ManagerTest extends CMTest_TestCase {
         $forkMock->set(function () use ($forkMock) {
             return $forkMock->getCallCount();
         });
-        $storage = $this->mockObject('CM_Clockwork_Storage_Memory');
+        $storage = new CM_Clockwork_Storage_Memory();
         /** @var CM_Clockwork_Storage_Abstract $storage */
         $managerMock = $this->mockClass('CM_Clockwork_Manager');
         $managerMock->mockMethod('_getCurrentDateTimeUTC')->set(function () use (&$currently) {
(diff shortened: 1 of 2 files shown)


alexispeter:
please review @fauvel 

Some explanations regarding the unit-tests (mocking clockwork is rather complicated):
The return-value of CM_Process::fork() (normally the forked process' pid) is used to identify the event executed in that process and is the key of the corresponding workload-result in the array returned by CM_Process::listenForChildren()

tests/library/CM/Clockwork/ManagerTest.php
```diff
@@ -49,11 +61,62 @@ public function testRunNonBlocking() {
...
+        $lastRuntime = null;
+        $process = $this->mockClass('CM_Process')->newInstanceWithoutConstructor();
+        $forkMock = $process->mockMethod('fork');
+        $forkMock->set(function () use ($forkMock) {
+            return $forkMock->getCallCount();
+        });
+        $storage = new CM_Clockwork_Storage_Memory();
+        /** @var CM_Clockwork_Storage_Abstract $storage */
+        $managerMock = $this->mockClass('CM_Clockwork_Manager');
+        $managerMock->mockMethod('_getCurrentDateTimeUTC')->set(function () use (&$currently) {
```
fauvel: You could remove the `&` I guess? Objects are passed by reference anyway.

tests/library/CM/Clockwork/ManagerTest.php
```diff
@@ -49,11 +61,62 @@ public function testRunNonBlocking() {
...
+        $currently->modify('4 seconds');
+        $this->assertFalse($_shouldRun->invoke($manager, $event1));
+        $this->assertFalse($_shouldRun->invoke($manager, $event2));
+        $process->mockMethod('listenForChildren')->set([]);
+        $manager->runEvents();
+
+        $this->assertEquals($startTime, $storage->getLastRuntime($event1));
+        $this->assertNull($storage->getLastRuntime($event2));
+
+        $managerMock->mockMethod('_getCurrentDateTimeUTC')->set(function () use (&$currently) {
```
fauvel: idem

tests/library/CM/Clockwork/ManagerTest.php
```diff
@@ -2,6 +2,18 @@
...
 
+    public function testRegisterEvent() {
+        $manager = new CM_Clockwork_Manager();
+        $manager->registerEvent(new CM_Clockwork_Event('foo', '1 second'));
+        $manager->registerEvent(new CM_Clockwork_Event('bar', '1 second'));
+        try {
+            $manager->registerEvent(new CM_Clockwork_Event('foo', '1 second'));
+            $this->fail('Registered duplicate event');
+        } catch (CM_Exception $ex) {
+            $this->assertSame('Duplicate event-name', $ex->getMessage());
```
fauvel: You could move this to a separate test and use a phpUnit annotation

fauvel:
The app code looks good, but I don't understand the unit tests to 100%  – maybe split it up into different test cases, so that others can better understand which behaviors exactly are being tested?

tests/library/CM/Clockwork/ManagerTest.php
alexispeter: There is still a difference between a reference to an object and a reference to a variable holding an object (see other tests), but in this case you're right, it doesn't need it.

tests/library/CM/Clockwork/ManagerTest.php
alexispeter: This test is already as atomic as it will get, the only thing being tested is that you can't register multiple events with the same name.

alexispeter:
please review @fauvel 
The tests I added all cover exactly one behaviour (but multiple aspects of this behaviour) :)
I could split it up a bit further, but the amount of setup-code required is just too much. I would be copying 30 lines and only changing 2 of them.

fauvel:
Lgtm!

fauvel:
Tagged 1.104.3

alexispeter:
over 100 :D

PR #1677, opened 2015-03-05
Adjust navigation to page anchors.

vogdb:
The problem comes when page has a position fixed header. The action `anchor.scrollIntoView();` does not count in the header and the anchor line gets hidden under the header.

The fix would be just to scroll up on the volume of the header's height. Will be this artificial solution inside CM acceptable to us? I'm asking because it is dictated by the integration of CM rather than the CM itself.

@njam @fauvel please share your thoughts.

### DIFF
--- a/library/CM/Layout/Abstract.js
+++ b/library/CM/Layout/Abstract.js
@@ -118,6 +118,7 @@ var CM_Layout_Abstract = CM_View_Abstract.extend({
       }
       if (anchor) {
         anchor.scrollIntoView();
+        $(document).scrollTop($(document).scrollTop() - page.$el.offset().top);
       }
     }
   },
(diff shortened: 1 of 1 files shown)


vogdb:
@njam @fauvel please look.

njam:
Hmm now we're scrolling twice - doesn't the second scoll override the first one?

vogdb:
No, it complements the first one.

vogdb:
@njam please look the second variant.

njam:
Now it might really make sense to use jQuery for finding that element? :)

vogdb:
Indeed! @njam please look.

fauvel:
Lgtm :P

njam:
lgtm

PR #1678, opened 2015-03-09
Add toggle method to jQuery.toggleNext plugin.

vogdb:


vogdb:
@njam please review.

njam:
I would prefer to slit this into two parts:
1. Parsing arguments, deciding which function to call
2. Actual code to execute for "default" and for "toggle".

I think we have a good structure in `jquery.offCanvas.js`, wdyt?

vogdb:
Absolutely agree. @njam please check.

njam:
- Define all prototype properties in one assignment (like in `jquery.offCanvas.js`)
 - Also define the "instance variable"-type properties for completeness ($toggler, $content). Add one-line type docu for them.
- constructor: Add docu
- constructor: Raise if $content has no length?
- toggle: Store `currentState` on the instance instead of the DOM (class)
- initializer: Don't set $toggler's data to `true`. Has no effect as its currently overwritten later in the main entry function.

vogdb:
> - Define all prototype properties in one assignment (like in `jquery.offCanvas.js`)
When you do something like this:
```js
Phone.prototype = {};
```

you erase prototype chain. It's not critical but it's not good either.

@njam Are you sure that we should do like this ?

njam:
> you erase prototype chain.

Not sure whether I understand the problem. We define the object in this file, so there's no possibility that anything else is on the prototype?

vogdb:
Yep. It's just considered to be a not good practice but lets stick to it.

vogdb:
@njam please review.

njam:
- I like the docu in the constructor! Agree, if done like this we don't need to assign the whole prototype in one go. You decide! I'm okay with reverting be566ed19242be8f2072bf1a956d74c2937d9a26
- constructor: change if-else to just `if` (because we throw anyway)?
- initialize: Pre-active state: Use `toggle()` function to apply, to avoid duplication?
 - Introduce `skipAnimation` for `toggle()`?

vogdb:
@njam please review.

njam:
lgtm!

PR #1679, opened 2015-03-09
Expect openx banner to contain tracking beacon even when no banner is delivered (as of Revive 3.1.0)

njam:
paired with @kris-lab
@vogdb please review


vogdb:
lgtm

PR #1647, opened 2015-03-10
Implement a jQuery-like approach for findAll functionality of CM_View_Abstract.

vogdb:
@fauvel @christopheschwyzer @njam FYI

As the result we will replace such cases
```js
    var notifications = this.findChild('SK_Component_Notifications');
    if (notifications) {
      notifications.toggleChatUnread(state);
    }
```
with
```js
   this.findChild('SK_Component_Notifications').toggleChatUnread(state);
```
The latter will not produce an error in case of `null` or not-found view.

Lets start the discussion!


vogdb:
I pushed here the code that I used to do the performance measuring. Please don't merge it =)

vogdb:
@njam please look at the results.

njam:
Could you try these things:
- Run the `_mockChild` 1000 times, so we know how much time it needs exactly
- Optimize `_mockChild` a bit - maybe not use `_.each` but native forEach?

njam:
@vogdb could you run those benchmarks?

vogdb:
@njam I remember about this issue but it does not have a high priority, so I postponed it. Do you want to implement it now ?

njam:
@vogdb agree it doesn't have high prio. We're still investigating whether it even makes sense. I think we shouldn't let it lie around too long though.
I was rather asking about my comment above ("Could you try these things"), so we can decide how to continue.

vogdb:
Ok. Here are the results measured in ms:
##### 1000 times mockChild
Chrome
```
69.59400000050664, 44.31599999952596, 42.852999999013264, 42.48699999880046, 41.74000000057276, 54.2249999998603, 48.98399999910907, 40.774000000965316, 42.192000000795815, 48.94899999999325, 42.97199999928125
```
iPhone5.1/iOS8.1
```
20.852ms, 14.753ms, 13.940ms, 13.651ms, 13.459ms, 17.209ms, 15.478ms, 16.107ms, 13.347ms, 15.242ms
```
Android 4.4
```
3232.6709999999803, 2675.422999999966, 3155.850999999988, 2773.224000000031, 3287.351999999999, 2921.343000000037, 3175.680999999997, 2683.601999999999, 2876.4610000000102, 2765.6390000000247, 3621.3759999999893
```

##### This gives us the value for single mockChild:

Chrome
```
0.06959400000050664, 0.04431599999952596, 0.042852999999013264, 0.04248699999880046, 0.04174000000057276, 0.0542249999998603, 0.04898399999910907, 0.040774000000965316, 0.042192000000795815, 0.04894899999999325, 0.04297199999928125
```
iPhone5.1/iOS8.1
```
0.020852ms, 0.014753ms, 0.013940ms, 0.013651ms, 0.013459ms, 0.017209ms, 0.015478ms, 0.016107ms, 0.013347ms, 0.015242ms
```
Android 4.4
```
3.2326709999999803, 2.675422999999966, 3.155850999999988, 2.773224000000031, 3.287351999999999, 2.921343000000037, 3.175680999999997, 2.683601999999999, 2.8764610000000102, 2.7656390000000247, 3.6213759999999893
```

##### This gives us the average value for single mockChild:
Chrome
```
0.0471896363634931
```

iPhone5.1/iOS8.1
```
0.0154038
```

Android 4.4
```
3.015329363636366
```

njam:
So what's your feeling? Is 3ms for a `findChild()` call too much?

vogdb:
It's not much but the thing that bothers me are such use cases:
```js

    var selector = this.findChild('SK_Component_Selector_TextFormatterImage');

    if (selector) {
      if (state) {
        selector.$el.slideDown('fast');
      } else {
        selector.$el.slideUp('fast');
      }
    //...
    }

```

How should we proceed with `selector.$el.slideDown('fast');` ? I didn't analyse all the cases but this is not a single usage of that type.

njam:
Hmmm, true, properties. Usually we shouldn't access views' properties from the outside, I think `$el` should be an exception.
How about setting the `$el` property to the jQuery-union of all `$el`s?

vogdb:
What all `$el`s do you mean? The `$el` of the element that we fake and all of its parent's `$el`s  ? Will we get a predicted results in this case for an operation like `selector.$el.slideDown('fast');`?

njam:
I meant something like:
```js
viewMock.$el = $().add(view1.$el).add(view2.$el)
```

When accessing `viewMock.$el` you will access a new jquery element containing references to all views' nodes.

vogdb:
But would it be correct to call `.slideDown('fast')` on all views' nodes ?

njam:
In that case yes, as that's the purpose of applying a function to multiple views.
Then the return of `findAll()` would be something like a `CM_ViewCollection`, which proxies function calls to all views in the collection.

vogdb:
Another edge case are methods that return values:
```js
    var currentPage = cm.getLayout().findPage();
    return !currentPage || !(
      currentPage.hasClass('SK_Page_Account_Premium') ||
      currentPage.hasClass('SK_Page_Payment_Abstract') ||
      currentPage.hasClass('SK_Page_Payment_Cascade')
      );
```
`findPage` uses `findChild`. So, when a mock page will be returned, its method `hasClass` will be empty function that does not return anything. In this particular case it's ok cause javascript will  typecast `udnefined` to `false` but if we had relied on `false` result here? Obviously it would have produced a silent error which is very bad. 

@njam wdyt ?

njam:
you mean if we would use `findAllAsCollection()` (tbd) instead of `findPage()` and then call `hasClass()` on it?
That's a general problem that if a function returns something and you suddenly have multiple return values. jQuery has the same problem, and sometimes they solve it by just returning the first return value.

We should use the `ViewCollection` mechanism to return things I think, but only to call functions on multiple views.

I agree with all these limitations it doesn't look so sexy anymore. Maybe we should invert the logic and rather emit an event for which the multiple views can bind?

vogdb:
> you mean if we would use `findAllAsCollection()` (tbd) instead of `findPage()` and then call `hasClass()` on it?

yep.

> I agree with all these limitations it doesn't look so sexy anymore. Maybe we should invert the logic and rather emit an event for which the multiple views can bind?

Something like this?
```js
cm.findViewAsync('CM_Component_Batman').done(function(batmanCmp){
  batmanCmp.gotham('Argh');
});
```

Now I think we should keep as it is. All these bad cases come from bad design solutions when one component wants to know about another component, which is not its parent or child, for some hack. We better minimise the number of those hacks.

njam:
Agree! So let's try to avoid these, maybe using events.
I'm closing then. Thanks for the research!

PR #1683, opened 2015-03-11
move persistence deletion into _onDelete()

alexispeter:
Provides more control over the deletion-process of simple models.
This will allow us to keep persistent data by simply overwriting _onDelete()

Needed for https://github.com/cargomedia/sk/issues/1367

This essentially forces us to adhere to the principle of only performing persistence deletion in _onDelete() and putting other deletion-related code in the _onDeleteBefore()/After()-hooks, what we already *should* be doing anyway.

This is only one of several possible solutions to the problem, but I think it's the least complex one.

### DIFF
--- a/library/CM/Model/Abstract.php
+++ b/library/CM/Model/Abstract.php
@@ -86,9 +86,6 @@ abstract class CM_Model_Abstract extends CM_Class_Abstract implements CM_Compara
             $asset->_onModelDelete();
         }
         $this->_onDelete();
-        if ($persistence = $this->_getPersistence()) {
-            $persistence->delete($this->getType(), $this->getIdRaw());
-        }
         if ($cache = $this->_getCache()) {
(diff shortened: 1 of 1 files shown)


alexispeter:
please review @njam

njam:
I like it.
@fauvel agree?

alexispeter:
tagged https://github.com/cargomedia/CM/releases/tag/1.107.0

fauvel:
@alexispeter can you check whether all the overriding methods do call their parent?

alexispeter:
What method do you mean exactly?
* _onDelete(): the idea is that you either override or don't override, but never call parent::_onDelete()
* _onDeleteBefore(): None of the models where I had to switch _onDeleteBefore() with _onDelete() need to call parent::_onDeleteBefore()

fauvel:
Ok, makes sense, sorry!

PR #1686, opened 2015-03-13
FB-293 Remove request singleton

fauvel:
We want the tracking services to store the clientId on the response's request object and not on the request singleton, otherwise the cookie isn't being sent to the client. To avoid such situations in the future (which are difficult to debug), I suggest to remove the request singleton altogether, and pass the request object all the way via the `CM_Frontend_Render`.

### DIFF
--- a/layout/default/Layout/Abstract/default.tpl
+++ b/layout/default/Layout/Abstract/default.tpl
@@ -56,7 +56,7 @@
   </head>
   <body id="{$viewResponse->getAutoId()}" class="{$viewResponse->getCssClasses()|implode:' '}">
 
-    {if CM_Http_Request_Abstract::hasInstance() && !CM_Http_Request_Abstract::getInstance()->isSupported()}
+    {if $request && !$request->isSupported()}
       <div id="browserNotSupported">
         <h2><span class="icon-warning"></span> {translate 'Your browser is no longer supported.'}</h2>
         <p>{translate 'We recommend upgrading to the latest Internet Explorer, Google Chrome, Firefox, or Opera. Click here for <a href="{$url}">more information</a>.' url='http://whatbrowser.org'}
--- a/library/CM/Form/Example.php
+++ b/library/CM/Form/Example.php
@@ -23,11 +23,8 @@ class CM_Form_Example extends CM_Form_Abstract {
     }
 
     public function prepare(CM_Frontend_Environment $environment, CM_Frontend_ViewResponse $viewResponse) {
-        if (CM_Http_Request_Abstract::hasInstance()) {
-            $ip = CM_Http_Request_Abstract::getInstance()->getIp();
-            if ($locationGuess = CM_Model_Location::findByIp($ip)) {
-                $this->getField('location')->setValue($locationGuess);
-            }
+        if ($location = $environment->getLocation()) {
+            $this->getField('location')->setValue($location);
         }
     }
 }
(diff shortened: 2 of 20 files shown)


fauvel:
@tomaszdurka please review

fauvel:
@njam please have a look, too!

tomaszdurka:
I did brief look on this pull-request (actually had a comment ready, but for some undesired reason Chrome was closed...).

Generally reasoning about creating `Frontend_Environment` was to store all the required "components" for `Frontend_Render` - including viewer, language, site, etc.
I think we map a few things from Request and explicitely put them to Environment (allowing some small modifications). After the change we will have following "problems":
- Render storing Request and Environment which are storing similiar things...
- Render ideally should not be aware of http concept
  - Unsupported browser warning could work as redirect instead of extra div? We could handle that somewhere higher - in Http module. Otherwise maybe we should pass along `isSupported` in Environment?
  - I am not sure why you need to pass Request to `getHtml()` - to get some extra variables? (additionally there is small bug - you not passing the request in abstract layout)

fauvel:
> I am not sure why you need to pass Request to getHtml()

This is the main reason why we are doing this PR actually. The request object does not only store input, but it also configures the response output! Specifically, you have these few lines of code in `CM_Http_Response_Abstract::process()`:
```php
if ($this->getRequest()->hasClientId()) {
    $requestClientId = $this->getRequest()->getClientId();
    if ($this->getRequest()->getCookie('clientId') != $requestClientId) {
        $this->setCookie('clientId', (string) $requestClientId, time() + (20 * 365 * 24 * 60 * 60));
    }
}
```
The clientId stored on the request object does **not** come from the client initially, but has been set during the response processing in `_process()`, and ought to be passed to the client as a cookie.

A requirement (not sure how important it is) is that not every request should trigger the generation of a clientId; this should only happen when the app needs it for some reason (anonymous tracking, split test on guest visitors).

Do you see any better way of telling the response that it should send a clientId cookie? Should we rather store the generated clientIds on the`CM_Frontend_Render `?

njam:
I also think render should *not* depend on any HTTP thingys.
To accomplish what we want now we would probably need to de-couple the request-client-id concept a bit from the HTTP-request? I.e. introduce something in-between. These are only very rough thoughts obviously.

Maybe we shouldn't use the client-id in GA afterall, I don't think there's a big advantage?

fauvel:
What about managing output cookies in `CM_Frontend_GlobalResponse`?

njam:
My point of view:
- A cookie is something directly connected to HTTP
- The whole rendering system should be usable without HTTP
- The GA tracking's HTML requires (optionally) an HTTP-request's client-id.
- The tracking's HTML is required when rendering a Layout.
- We *have* an optional way to let a Page know its corresponding HTTP request/response: `prepareResponse()` (called from `CM_Http_Response_Page`).

As I said I'm a bit hesitant to store the HTTP request on the Render, because it feels like the wrong place. Alternatively I could imagine:
- Passing it to the Layout's render adapter. We don't do something like this currently but it would have minimal impact.
- Calling something like `prepareResponse()` on the layout, and then storing *something* on the instance. But how to make it lazy created?
- Store `clientId` on `Frontend_Environment`. But could we still make it lazy created? -> tricky without a request/response

Overall this would be much easier without lazy-creating the client id. But I think it's worth keeping that mechanism for performance reasons?

tomaszdurka:
What's the status of this one? :)

njam:
We're not using the clientId of the request in the GoogleAnalytics client anymore!
@fauvel could you have a look if it's easily doable now, or what the roadblocks are?

njam:
Ticket was closed, so closing the review

PR #1690, opened 2015-03-13
Wrap child process code in fork() in try/catch

njam:
Otherwise an exception in the child process might lead to catch-blocks from the parent process to be executed (that were copied during forking).

Paired with @alexispeter
@tomaszdurka fyi


tomaszdurka:
Interesing - lgtm.

PR #1691, opened 2015-03-16
Reset serviceManager instances on forking

alexispeter:
To avoid sharing mysql connection between processes.
Addressing https://github.com/cargomedia/puppet-cargomedia/issues/724

After forking, in the child call something like:
```php
CM_Service_Manager::getInstance()->resetInstances();
```

Can add an acceptance test with something like this:
```php
$assert = function ($expected, $actual) {
    if ($expected !== $actual) {
        throw new Exception('Assertion failed: `' . $expected . '` !== `' . $actual . '`');
    }
};

$mysql = CM_Service_Manager::getInstance()->getDatabases()->getMaster();
$mysql->setBuffered(false);

$process = CM_Process::getInstance();
$process->fork(function () use ($assert, $mysql) {
    for ($i = 0; $i < 1000; $i++) {
        $result2 = $mysql->createStatement('SELECT "bar"')->execute()->fetchColumn();
        $assert('bar', $result2);
    }
});

for ($i = 0; $i < 1000; $i++) {
    $result1 = $mysql->createStatement('SELECT "foo"')->execute()->fetchColumn();
    $assert('foo', $result1);
}

$process->waitForChildren();
```

@alexispeter for you?

### DIFF
--- a/library/CM/Bootloader.php
+++ b/library/CM/Bootloader.php
@@ -174,7 +174,7 @@ class CM_Bootloader {
     protected function _registerServices() {
         $serviceManager = CM_Service_Manager::getInstance();
 
-        $serviceManager->registerInstance('debug', new CM_Debug($this->isDebug()));
+        $serviceManager->register('debug', 'CM_Debug', ['enabled' => $this->isDebug()]);
         $serviceManager->register('filesystems', 'CM_Service_Filesystems');
         $serviceManager->register('filesystem-tmp', 'CM_File_Filesystem', array(
             new CM_File_Filesystem_Adapter_Local($this->getDirTmp()),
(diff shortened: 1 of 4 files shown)


njam:
Why is this
```php
$serviceManager->registerInstance('debug', new CM_Debug(CM_Bootloader::getInstance()->isDebug()));
```
needed? What's the story here @tomaszdurka ?

alexispeter:
It's the only service that's instantiated programatically instead of via config, because the debug-flag is set programatically on the bootloader. Alternatively we could set theservice-config from within the bootloader to work around this.

njam:
As discussed we'll keep the approach, but make "debug" a service with configuraton.

alexispeter:
please review @njam

njam:
lvgtm

PR #1698, opened 2015-03-18
Make CM_Usertext_Usertext extensible regarding custom filters in inheriting classes (clearFilters() call)

alexispeter:


alexispeter:
please review @njam

enables addition of filters in chlid-classes without resetting them.

njam:
lgtm

PR #1699, opened 2015-03-18
Rework create, update, delete indexes for Elasticsearch

tomaszdurka:


tomaszdurka:
@alexispeter please review

tomaszdurka:
@njam Please review...

tomaszdurka:
Tbh, I am not too proud of how this ended up. I pass Type to cluster->createIndex() method which anyway uses client of the Type. Maybe once the client is present on the Type there should be rather Type->createIndex()?

njam:
- It might be good to have `CM_Elasticsearch_Cluster::updateIndex()` and `CM_Elasticsearch_Type_Abstract::updateItem()` in the same file (as they are two sides of the same queue). How about moving both to the type?
- CM_Elasticsearch_Cluster::createIndex: I don't think this should refresh the index. This might be a heavy operation and I wouldn't expect it to happen when calling this API? Can we move it to the usages?
 - CM_Elasticsearch_Cluster::createIndex,deleteIndex: Remove both of there?
- CM_Elasticsearch_Index_Cli: Very good to get rid of all this logic from here!
- CM_Elasticsearch_Index_Cli: Introduce a method to return either all types or one based on name?

tomaszdurka:
- agree then
- ok with the refresh, but still I think those methods should be proxied on Type so we don't need to dive into Elastica
  - move to Type?
- ok
- ok

njam:
- CM/Elasticsearch/Type/Abstract.php: I would prefer `createVersioned`, it makes clear what the function does plus it goes well together with `_create`. Otherwise there's a public create**Index** and a private one just _create (so rather relating to **type**).
- CM_Elasticsearch_Index_Cli: Introduce a method to return either all types or one based on name?

njam:
lgtm, what about tests?

njam:
lgtm

PR #1697, opened 2015-03-18
Fix of absolute header for page anchors do not work.

vogdb:
"btw when you navigate to this anchor link about cupids you won't see full paragraph due to "absolute header" :\" (c) the greatest test engineer.


vogdb:
@fauvel please review.

njam:
- So this will replace `SK_Layout_Default.scrollTo()`
- Isn't the `if ($anchor.length)` still needed? Otherwise we'll always scroll to `0` if there's no anchor?

vogdb:
> - So this will replace SK_Layout_Default.scrollTo()

Yes.
@njam please review.

njam:
lgtm

fauvel:
So what was the problem? `page` was not the right one in this context?

vogdb:
It's hard to explain only in the context of CM =( Lets watch how it will behave.

PR #1700, opened 2015-03-18
Revert "Make CM_Usertext_Usertext extensible regarding custom filters in...

alexispeter:
Reverts cargomedia/CM#1698

merge was premature

fyi @njam


njam:
good

PR #1695, opened 2015-03-18
Destroy jquery.autosize

vogdb:
Via https://github.com/jackmoore/autosize/issues/210
We currently don't destroy the autosize plugin when it's not used anymore.
Should be done on view destruction, and maybe set up in `cm.dom.setup`?

```
$('textarea').trigger('autosize.destroy');
```

@vogdb thoughts?


vogdb:
Ok. I've added to teardown all of the disposals that are possible.
@njam please review.

njam:
lgtm

PR #1706, opened 2015-03-23
Move mail delivery agent configuration to protected function

ppp0:
Right now part of  `_send`.. awkward to overwrite in subclasses

Better: separate function

### DIFF
--- a/library/CM/Mail.php
+++ b/library/CM/Mail.php
@@ -364,7 +364,6 @@ class CM_Mail extends CM_View_Abstract implements CM_Typed {
 
     /**
      * @throws CM_Exception_Invalid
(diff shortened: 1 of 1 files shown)


ppp0:
@njam please review

njam:
- _send: Docu: Only include excepections that will be thrown directly from here (`CM_Exception_Invalid`)

ppp0:
@njam please review

njam:
- _send: I don't think it can throw `phpmailerException` - we catch those and rethrow?

ppp0:
@njam sorry, my bad

njam:
lgtm

PR #1707, opened 2015-03-25
Add jquery.lazyload.

vogdb:
For image lazy loading.


njam:
- CM_Usertext_MarkdownImgLazy: How about rather adding this functionality directly to `CM_Usertext_Filter_Markdown` (with an argument). This way we'll reduce duplication and can combine the functionality better with other markdown functionality?
- CM_Usertext_MarkdownImgLazy: Add visibility keywords for functions
- CM_Usertext_MarkdownImgLazy::addLazyAttrs: Add docu

vogdb:
@njam please review.
Should I fix visibility keywords for the rest of existing code in CM_Usertext_Markdown ?

njam:
> Should I fix visibility keywords

Yes!
Could you also add a test that covers the new functionality?

vogdb:
@njam please review.

vogdb:
@njam please hold here with review.

vogdb:
@njam please review.

njam:
- CM/Usertext/Markdown.php: _doImages_*: Instead of returning early if *not* in lazy mode I would put the additional transformation into an `if ($this->_imgLazy)`. This way  the code gets shorter and more portable (can add more features later)
- CM_Usertext_Filter_MarkdownTest: Tests fail
- CM/Usertext/Markdown.php::_addLazyAttrs: I would make this more strict (and robust), with a regexp replacement, maybe à la?:
```php
        $tagText = preg_replace('#^<img src="(.+?)" alt="(?<alt>.+?)" />$#', '<img data-src="\1" alt="\2" class="lazy" />', $tagText, -1, $count);
        if (0 === $count) {
            throw new CM_Exception_Invalid('Cannot replace img-tag `' . $tagText . '`.');
        }
```

vogdb:
@njam please look.

njam:
- _addLazyAttrs: Throws-docu should mention exception class name. PhpStorm will autocomplete this for you!

njam:
@NicolasSchmutz wanna acceptance test this? Lazy loading of images in the feed as you scroll down. Me or Aleksei can help you set it up locally.

njam:
- How about adding the call of the plugin to `cm.dom.setup`? This way it will work out of the box.

vogdb:
I think should avoid it currently because we want to apply this plugin partially. All of lazy images, which are inside clipSlide or other structures that hide images, should not be loaded until clipSlide reveals them.

vogdb:
@NicolasSchmutz what about acceptance test ?

vogdb:
@njam please review btw.

njam:
It looks good, @NicolasSchmutz please test thoroughly.

NicolasSchmutz:
see: https://github.com/cargomedia/sk/pull/1421

njam:
@vogdb did you see this? http://callmecavs.github.io/layzr.js
Should we consider using it?

vogdb:
No, actually I missed it. Seems to be a very useful finding. Will try it.

vogdb:
Actually it does not allow you to restrict the images that you want to lazyload. It applies itself to all of the images. We can't do that because we want to skip most of the images inside clipSlide. So, no luck with it.

vogdb:
Depends on #1738.

njam:
- I think on the markdown level the code shouldn't mention "lazy", as it's rather about image dimensions here
- Good finding to use the *extra attributes* for the dimensions. How about overriding `doExtraAttributes($tag_name, $attr)`, add a special case for tag name `img` where we parse a value of `123x456` and return width/height? (and otherwise just call the original implementation). This way we don't need to remove the `id` arg later on again.

vogdb:
@njam please review.
I removed `imgLazy` flag entirely. Now it's more about markdown images format. Images can have dimensions or class attrs. If they have dimensions then we transform them.

vogdb:
@njam please postpone the review.

njam:
- CM/Usertext/Markdown.php: _transformDimensions: Some attribute matchers use `(.+?)` and some `([^"]+)` - I'd streamline to the latter?
- CM/Usertext/Markdown.php: _transformDimensions: $ratio: Skip the `% 100`? I.e. if the dimensions are "300x100" then the ratio should be `3`?

*postponing further review*

vogdb:
CM builds do not work because of Travis. My local builds do not work because of `CM_EventHandler_EventHandlerTraitTest`
```
PHP Fatal error:  Call to undefined method
CM_EventHandler_EventHandlerTraitTest::mockTrait() in
/home/vagrant/cm/tests/library/CM/EventHandler/EventHandlerTraitTest.php on line 7
```

Waiting here.

njam:
@tomaszdurka can you look into the mocka problem Aleksei is experiencing?

tomaszdurka:
Composer update should do the job.

christopheschwyzer:
Approach:
```css
.embeddedWrapper {
  position: relative;
  display: inline-block;
  overflow: hidden;

  .embeddedWrapper-ratio {
    display: block;
    width: 100%;
    padding-bottom: 100%;
  }

  .embeddedWrapper-object {
    position: absolute;
    top: 0;
    left: 0;
    width: 100%;

    &.centerVertically {
      top: 50%;
      transform: translateY(-50%);
    }
  }

}
```

Demo:
http://jsfiddle.net/c2bk3trb/9/

See also: https://github.com/suitcss/components-flex-embed/blob/master/lib/flex-embed.css

njam:
@vogdb about caching check out this ticket: https://github.com/cargomedia/CM/pull/1737

In this case *local* caching (APC) would be best, so you would use `CM_Cache_Local` instead of `CM_Cache_Shared`.

vogdb:
@njam please review. I decided to move `lazy` classname to Markdown cause it belongs to it. `_transofrmDimensions` also changed `src` to `data-src`, so in the end we couldn't split the logic on dimensions and laziness of images. That is why I decided to be honest and return its lazy meaning back.

library/CM/Frontend/TemplateHelper/RatioKeeper.php
```diff
@@ -0,0 +1,36 @@
...
+        } elseif (isset($params['ratio'])) {
+            $ratio = $params['ratio'];
+            $width = 100;
+            $height = $width * $ratio;
+        } else {
+            $width = $height = 1;
+        }
+
+        $cache = CM_Cache_Local::getInstance();
+        $imageData = $cache->get($cache->key('_ratioKeeper:', $width, $height), function () use ($width, $height) {
```
tomaszdurka: We have a general placeholder for all cache key prefixes - `CM_CacheConst`. I am not super happy about this solution but at least it makes sure we do not have any conflicts.

What do you think about using class and method name here?
```php
$imageData = $cache->get($cache->key(__CLASS__, __METHOD__, $width, $height), function () use ($width, $height) {
```

library/CM/Frontend/TemplateHelper/RatioKeeper.php
```diff
@@ -0,0 +1,36 @@
+<?php
+
+class CM_Frontend_TemplateHelper_RatioKeeper {
```
tomaszdurka: This class main purpose is to create an image element with blank image as source. Wrapped in div element.

The name of the helper does not really describe what this class is about.
What about calling this class `CM_Frontend_TemplateHelper_ContentPlaceholder` or `CM_Frontend_TemplateHelper_Placeholder`?


Additionally I see you did simplest thing - copied smarty helper and created static method.
Let's try to use this opportunity to model this helper bit better.

We can introduce common private method for creating blank image with ceratin size.
```php
private function _create($width, $height, $stretch = null) {
       $cache = CM_Cache_Local::getInstance();
       $imageData = $cache->get($cache->key('_ratioKeeper:', $width, $height), function () use ($width, $height) {
           $image = imagecreate($width, $height);
           ob_start();
           imagecolorallocate($image, 255, 255, 255);
           imagepng($image);
           $imageData = ob_get_contents();
           ob_end_clean();
           return $imageData;
       });
       $imageSrc = 'data:image/png;base64,' . base64_encode($imageData);
       $output = '<div class="ratioKeeper' . ((isset($params['stretch']) && $params['stretch']) ? ' stretch' : '') . '">';
       $output .= '<img class="ratioKeeper-size" src="' . $imageSrc . '">';
       $output .= $this->_content;
       $output .= '</div>';
       return $output;
}
```

And then additionally two public methods:

```php
public function create($width, $height);
public function createUsingRatio($ratio);
```

@vogdb wdyt?

tomaszdurka:
Review is done. I only had some notes regarding TemplateHelper class itself. Rest (usages) looked fine for me.

library/CM/Frontend/TemplateHelper/RatioKeeper.php
vogdb: And what about `stretch` and `content` params ?

njam:
@tomaszdurka thanks for stepping in!

library/CM/CacheConst.php
```diff
@@ -89,4 +89,6 @@ class CM_CacheConst {
     const AdproviderZones = 'AdproviderZones';
 
     const Emoticons = 'Emoticons';
+
+    const Frontend_TemplateHelper_RatioKeeper = 'Frontend_TemplateHelper_RatioKeeper';
```
tomaszdurka: Probably you want to rename this constant into ContentPlaceholder too. Thanks.

library/CM/Frontend/TemplateHelper/ContentPlaceholder.php
```diff
@@ -0,0 +1,41 @@
...
+class CM_Frontend_TemplateHelper_ContentPlaceholder {
+
+    /**
+     * @param string $content
+     * @param int    $width
+     * @param int    $height
+     * @param bool   $stretch
+     * @return string
+     */
+    public static function create($content, $width = 1, $height = 1, $stretch = false) {
```
tomaszdurka: Another convention we follow for all our php code. We do only use null as default values. Some of us says it simplifies passing default values - as you simply pass null. Some other think different, but for now we have convention to use only nulls.

It usually leads to something like that:
```php
public static function create($content, $width = null, $height = null, $stretch = null) {
  if (null === $width) {
    $width = 1;
  }
  if (null === $height) {
    $height = 1;
  }
  $stretch = (bool) $stretch;
 // ...
```

Yes, we could argue if this is more understandable. but let's follow it this time.

tomaszdurka:
What about renaming library/CM/SmartyPlugins/block.ratioKeeper.php to ContentPlaceholder. I think it describes its role way better. @vogdb agree?

library/CM/Frontend/TemplateHelper/ContentPlaceholder.php
vogdb: As you wish, master :rage4:

tests/library/CM/Usertext/Filter/MarkdownTest.php
```diff
@@ -19,4 +19,14 @@ public function testProcessWithoutLinks() {
...
+
+    public function testProcessWithImgLazy() {
+        $text = "#Headline#\n * element1\n * element1\n\nparagraph\n\nfoo_bar_foo\nfoo _bar_ foo\nLink: [google.com](http://www.google.com)\ntest2\n![Img text](/path/to/img.jpg){\$200x100}";
+        $filter = new CM_Usertext_Filter_Markdown(null);
+        $actual = $filter->transform($text, new CM_Frontend_Render());
+        $this->assertStringStartsWith("<h1>Headline</h1>\n<ul>\n<li>element1</li>\n<li>element1</li>\n</ul>\n<p>paragraph</p>\n<p>foo_bar_foo</p>\n<p>foo <em>bar</em> foo</p>\n<p>Link: <a href=\"http://www.google.com\">google.com</a></p>\n<p>test2</p>", $actual);
+
+        $expected = '/<div class="contentPlaceholder"><img class="contentPlaceholder-size" src="[^"]+"><img data-src="\/path\/to\/img\.jpg" alt="Img text" class="lazy"\/><\/div>/';
+        $this->assertRegExp($expected, $actual);
+    }
```
tomaszdurka: Is it easier to use regexp than our dom-test-helpers? (curious)

tomaszdurka:
Looks good! :)

tests/library/CM/Usertext/Filter/MarkdownTest.php
vogdb: what dom-test-helpers you mean? possibly I don't know about them.

tomaszdurka: Example usage:

```php
public function testFind() {
        $list = new CM_Dom_NodeList('
          <div class="foo">
            lorem ipsum dolor
            <div class="bar" id="bar1">
              lorem ipsum
              <div class="bar" id="bar2">
                hello world
               </div>
            </div>
          </div>
          <div class="foo">lorem</div>
        ');
        $this->assertSame(0, $list->find('.foo')->count());
        $this->assertSame(2, $list->find('.bar')->count());
        $this->assertSame(1, $list->find('#bar1 .bar')->count());
        $this->assertSame(1, $list->find('#bar2')->count());
        $this->assertSame(1, $list->find('> .bar')->count());
        $this->assertSame(1, $list->find('.bar > .bar')->count());
    }
```

vogdb:

tomaszdurka:
Just for my own information: the space before slash is not required anymore or it was never required?

christopheschwyzer:
@vogdb you can cherry-pick this commit for the usertext adjustment https://github.com/cargomedia/CM/commit/95218078abc36a3cd86b6fba0fa943a3fac2d273

@tomaszdurka I think since Netscape™ it's not required anymore. Nowadays you don't need self-closing tag `/` either. We just do it for readability. 

tl;dr: 
`<br>`, <br/>` and `<br />` are all valid according to _HTML5_ specs http://www.w3.org/html/wg/drafts/html/master/syntax.html#void-elements

vogdb:
It's hard for me to tell about was the space ever required. All I can tell is that it's ok without it.
http://www.w3.org/TR/REC-xml/#sec-starttags
http://en.wikipedia.org/wiki/XHTML#Common_errors

vogdb:
@christopheschwyzer it's done. Please verify.

christopheschwyzer:
Lgtm @tomaszdurka maybe wanna have a look as well

christopheschwyzer:
One last detail: Can you add `vertical-align: top;`  to `.contentPlaceholder-size`

```
  .contentPlaceholder-size {
    max-width: 100%;
    max-height: 100%;
    vertical-align: top;
  }
```

vogdb:
done. @tomaszdurka what is your opinion?

christopheschwyzer:
lgtm

vogdb:
Ok. What about you, @tomaszdurka ?

tomaszdurka:
Looks magnificent!

library/CM/App.js
```diff
@@ -340,7 +340,26 @@ var CM_App = CM_Class_Abstract.extend({
       $dom.find('.openx-ad:visible').openx();
       $dom.find('.epom-ad').epom();
       $dom.find('.fancySelect').fancySelect();
+      this._setupLazyImages($dom);
```
tomaszdurka: Oh wait, could this not be simpler? :)

library/CM/Frontend/TemplateHelper/ContentPlaceholder.php
```diff
@@ -0,0 +1,48 @@
...
+    }
+
+    /**
+     * @param int $width
+     * @param int $height
+     * @return string
+     */
+    private static function _createBlankImage($width, $height) {
+        $cache = CM_Cache_Local::getInstance();
+        return $cache->get($cache->key(CM_CacheConst::Frontend_TemplateHelper_ContentPlaceholder, $width, $height), function () use ($width, $height) {
```
njam: I would not even use a cache-const here!
See this note: https://github.com/cargomedia/CM/pull/1737#issue-69310775

tomaszdurka: Oh you've probably didn't see our discussion before? I've proposed cache const as it kinda makes sure we don't reuse cache keys unintentionally. I've actually proposed __CLASS__ __METHOD__ solution as alternative, but thought it would be better to discuss this pattern before-hand - haven't noticed comment you've pointed..

library/CM/App.js
vogdb: This code was proved by @njam! So, it can't.

vogdb:
@tomaszdurka please review.

library/CM/App.js
```diff
@@ -340,7 +340,26 @@ var CM_App = CM_Class_Abstract.extend({
...
+
+      var $hiddenImageList = $();
+      $dom.find('.clipSlide').each(function() {
+        var $imageList = $(this).find('img.lazy:gt(0)');
+        $hiddenImageList = $hiddenImageList.add($imageList);
+      });
+
+      $dom.find('img.lazy').not($hiddenImageList).unveil(600);
+    },
+
```
tomaszdurka: What about restructuring it a bit.

Outside of this method:
1. Remove `.lazy-autoload` class from all latter `clipSlide` images. 
2. One-time bind on `toggle.clipSlide` to trigger `load` on children images.

In setupLazyImages:
1. One-time bind on `load` event to all the `.lazy` images to trigger `unveil`.
2. Trigger `load` event for all `.lazy-autoload` images.

tomaszdurka: @vogdb @njam thoughts?

vogdb: I think it's more complicated the the current solution.

vogdb: @njam please share your wisdom.

tomaszdurka: I did not consider that unveil actually works differently when accepting multiple images. Seems it throttles images to not load all together. The approach I've proposed probably is not possible cause of that.

vogdb:
@tomaszdurka please review.

vogdb:
@tomaszdurka so wdyt ?

library/CM/App.js
njam: Talking with Tomasz, ideas:
- Moving `.on('toggle.clipSlide'` into the `each` to reduce duplication
- Rename `$hiddenImageList` to `$imageListNotLoad`???

vogdb:
@christopheschwyzer please look at `img.external` in the last commit. Is it ok ?

vogdb:

library/CM/App.js
tomaszdurka: This code looks ok for me (for future review requests).

christopheschwyzer:
css is fine!

tomaszdurka:
@vogdb Build is failing:
```
There was 1 failure:
1) CM_Usertext_Filter_MarkdownTest::testProcess
Failed asserting that two strings are identical.
--- Expected
+++ Actual
@@ @@
 <p>test2</p>
-<p><img src="/path/to/img.jpg" alt="Img text" /></p>
+<p><div class="usertext-image"><img src="/path/to/img.jpg" alt="Img text"/></div></p>
/home/travis/build/cargomedia/CM/tests/library/CM/Usertext/Filter/MarkdownTest.php:11
/home/travis/build/cargomedia/CM/tests/helpers/CMTest/library/CMTest/TestCase.php:23
```

PR #1708, opened 2015-03-27
Epom Adprovider

vogdb:


vogdb:
@njam please review.

njam:
@vogdb let's pair-finish this? lmk when you have time!

kris-lab:
look good to me!

njam:
Few notes:
- What is `key`, shouldn't this be configurable?
- Add a test
- Maybe use `CM_Params::jsonEncode()` (which has error checking)?

vogdb:
> - What is `key`, shouldn't this be configurable?
 - Add a test
 - Maybe use `CM_Params::jsonEncode()` (which has error checking)?

We are just testing a new provider. So, I decided not to be so perfect here. There is a big chance that we throw this code away, isn't it ?

kris-lab:
@njam I agree on all point but what is the chance that we could stick to this provider? Do we have another options?

njam:
Agree that this is code which might not live long, and we shouldn't invest too much time in making it shiny.

But making the `key` configurable is really just *deleting a line of code*, so not really more effort.
Regarding the test it would be a *bit* more effort to implement, but then again less to review (as you know the code runs and works)!

PR #1709, opened 2015-03-31
Followup of Epom Adprovider.

vogdb:
Connects to #1708.

### DIFF
--- a/client-vendor/after-body/jquery.epom/jquery.epom.js
+++ b/client-vendor/after-body/jquery.epom/jquery.epom.js
@@ -46,6 +46,8 @@
         }
       };
 
+      variables['window_width'] = $(window).width();
+      variables['window_height'] = $(window).height();
       variables['format'] = 'jsonp';
       $.getJSON(src + '?callback=?', variables, loadCallback);
     });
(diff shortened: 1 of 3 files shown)


vogdb:
@kris-lab please review.
@njam FYI.

njam:
How about just passing the "accessKey" for each zone through the zone config?

vogdb:
@njam please review.

kris-lab:
- Maybe instead of passing array with accessKey we could do something:
`deafult.php`
```
$config->CM_AdproviderAdapter_Epom->accessKey = 'aaaaaaaaaaa'
```
`CM_AdproviderAdapter_Epom`
```
private function _getAccessKey() {
        return self::_getConfig()->accessKey;
}
```

@njam wdyt?

vogdb:
> - Maybe instead of passing array with accessKey we could do something:
`deafult.php`
```
$config->CM_AdproviderAdapter_Epom->accessKey = 'aaaaaaaaaaa'
```
`CM_AdproviderAdapter_Epom`
```
private function _getAccessKey() {
        return self::_getConfig()->accessKey;
}
```

It was like it before =)

njam:
@kris-lab I told Aleksei to remove that, mainly because I'm not too fond of the whole class-config-approach. Also in the ad provider case we have this zone-configuration built-in.
Potentially we could load zones from multiple epom accounts/servers, so I think configuring it on the zone level is not bad.

Restarted the travis build.

vogdb:
@kris-lab or @njam please review

njam:
lgtm

PR #1711, opened 2015-04-01
Optimize split tests with UCB1 Tuned allocation

fauvel:
Algorithm based on the seminal work of Auer, Cesa-Bianchi & Fischer: http://homes.di.unimi.it/~cesabian/Pubblicazioni/ml-02.pdf

See in particular Part 4: Experiments. I've adjusted the tuned upper confidence bound to take into account the fact that the rewards distribution are not defined on `[0, 1]`, but on some unknown interval like `[0, max order value]`. Since this interval is unknown, you cannot easily compute an upper confidence bound on the variance itself. So I'm simply using the observed variance to tune the algorithm, instead of using its upper confidence bound. Accordingly, I initialize the algorithm by choosing the least used variation until all of them have a positive variance (=a first conversion, in practice).


fauvel:
@njam please review!

njam:
CM_Model_Splittest:
- While we're at it how about introducing a `public static function create()` with a non-array signature. Would help writing that call, especially now that new parameters are added.

CM_Model_SplittestVariation: 
- In general there's too many cache invalidations now. I understand this is only for tests - maybe we should invalidate caches manually there?
- _getVariationRandom: How about splitting this into two (or three) functions (instead of adding comments), for random, weighted and optimized allocation?

Mysql Indices `cm_splittestVariation_fixture`:
- id
- requestClientId, splittestId
- userId, splittestId
- splittestId, variationId, conversionStamp
- splittestId, variationId, createStamp
- ... Splittest::_setConversion: Don't use conversionStamp

fauvel:
@njam please review

njam:
Looks very good!

- _getVariationWithUpperConfidenceBoundPolicy: Add a link to the paper and section?
- I've sent you a dump of current splittest tables, could you verify:
 - Time needed for migration
 - Performance of `_getAggregationData()` [1]

---
[1]
```php
$splittest = new CM_Model_Splittest('Index-Versions');
/** @var CM_Model_SplittestVariation $variation */
$variation = $splittest->getVariations()->getItem(0);

CM_Util::benchmark();
for ($i = 0; $i < 30; $i++) {
    $variation->getConversionCount();
    CM_Cache_Local::getInstance()->flush();
}
echo CM_Util::benchmark() . PHP_EOL;
```

fauvel:
Benchmark, before: 22242.46 ms
Migration script: 32.0 s
Benchmark, after: 1010.77 ms

fauvel:
@njam please review and merge!

njam:
lgtm!

PS: This might make a good blogpost!

fauvel:
Tagged 1.109.0

njam:
@fauvel you tagged this as a breaking change - because of this PR? Please remember to prefix the breaking change(s) in the tag with "Breaking: ".

fauvel:
My PR is not breaking, but I wasn't sure about the other changes... That's why.

Issue #1712, opened 2015-04-01
manifest.json (Chrome) not loading from cross-domain location


njam:
E.g. www.denkmal.org has the following *link* tag:
```html
<link rel="manifest" href="http://cdn.denkmal.org/layout/de/35/1427044499/manifest.json">
```

That `manifest.json` is never loaded by Chrome. If the domain in the `href` is changed to `www.denkmal.org` it *is* loaded.

[W3C spec](http://www.w3.org/TR/appmanifest/) seems to allow CORS (6.1.1 Content security policy), and we set the appropriate headers in `CM_Http_Response_Resource_Abstract`.

@christopheschwyzer @vogdb  @cargomedia/devops  ideas, thoughts?

ppp0:
Are you sure you are setting the header right? If I inspect the request/response of http://cdn.denkmal.org/layout/de/35/1427044499/manifest.json I get:

```
HTTP/1.1 200 OK
Content-Type: text/plain; charset=utf-8
Transfer-Encoding: chunked
Connection: keep-alive
Server: nginx
Date: Thu, 02 Apr 2015 09:45:08 GMT
Access-Control-Allow-Origin: http://www.denkmal.org
Expires: Fri, 01 Apr 2016 09:45:08 GMT
Cache-Control: max-age=31536000
Content-Encoding: gzip
Age: 2493
X-Cache: Hit from cloudfront
[...]
```

njam:
What is it that looks wrong - there is `Access-Control-Allow-Origin`?

ppp0:
yes, but allowing `http://www.denkmal.org` not `cdn.denkmal.org`? Did I get wrong what the problem is?

njam:
I think that's correct, as the file is on `cdn.denkmal.org` but included in the website at `www.denkmal.org`. See http://stackoverflow.com/questions/10636611/how-does-access-control-allow-origin-header-work

One thought (which I didn't verify yet), is that the browser issues an `OPTIONS` request which we don't support?

ppp0:
See https://w3c.github.io/manifest/ Section 2.1 / Note - Suggests that the `manifest.json` is not suitable to be cached by a CDN?

njam:
Hmm, you're right, but what about 6.1.1?

2.1:
> Please also note that manifest's URL has to be same origin, so you can't use them directly with a CDN.

6.1.1:
[...]

PR #1713, opened 2015-04-03
Serve manifest.json from same origin

njam:
Addressing #1712

@dlondero @mariansollmann please review

### DIFF
--- a/layout/default/Layout/Abstract/default.tpl
+++ b/layout/default/Layout/Abstract/default.tpl
@@ -40,7 +40,7 @@
     <link rel="icon" type="image/png" href="{resourceUrl path='img/meta/favicon-96x96.png' type='layout'}" sizes="96x96">
     <link rel="icon" type="image/png" href="{resourceUrl path='img/meta/favicon-16x16.png' type='layout'}" sizes="16x16">
 
-    <link rel="manifest" href="{resourceUrl path='manifest.json' type='layout'}">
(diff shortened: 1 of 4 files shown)


njam:
This file needs to be served from the same site as the web app:
> Please also note that manifest's URL has to be same origin, so you can't use them directly with a CDN.
http://www.w3.org/TR/appmanifest/

njam:
@vogdb maybe you can review too?

tomaszdurka:
Ok! Bit hacky, but there is no other way I guess.

njam:
Unfortunately seems to be the only way. The spec is [not consistent though](https://github.com/cargomedia/CM/issues/1712#issuecomment-88882940), and this is very new technology.
Let's do it like this for now and maybe it can be changed in the future.

vogdb:
And why did you put me on the review ? O_o I was researching about this stuff, analysing...

njam:
@vogdb sorry, wdyt, is it okay? Feedback still appreciated! :)

vogdb:
It's hard to say. I know a little about this topic. I was ready to spend some time on researching but now it's merged. No point =) I'm happy to return to my current task.

njam:
I experimented with this and found that for Chrome the file *has* to be served from the same hostname, no way around it. The spec also mentions it.
So I think it's the only way.

vogdb:
lgtm then!

PR #1714, opened 2015-04-03
Don't specify density for PNG in app manifest

njam:
I don't think we should specify a *density* for PNGs, as they all are suited only for a density of 1.
Afaik the browser itself picks the right version based on the screen size and density.

See https://w3c.github.io/manifest/#icon-object-and-its-members

@vogdb please review
@christopheschwyzer fyi


vogdb:
lgtm

njam:
@christopheschwyzer I think we could also just serve 1 SVG if it makes things easier..

christopheschwyzer:
Agree on using _SVG_. Needed to test if this worked properly.

PR #1716, opened 2015-04-04
Use file extension to detect mime-type of javascript files

njam:
@fauvel @mariansollmann please review


mariansollmann:
looks good

PR #1715, opened 2015-04-04
Allow to retrieve resource files from sameOrigin on client

njam:
Via https://github.com/cargomedia/CM/pull/1713

@vogdb @tomaszdurka please review

PR #1717, opened 2015-04-04
Introduce RootProxy response to serve layout resource files from root dir

njam:
Some resource files need to be served from the root directory of the web site.
E.g. a service worker should be served at `/service-worker.js`.

I'm introducing a "root proxy" response which can serve any resource files off the
root directory, by escaping `/` with `--` in the URL.

This replaces #1715

For example we can then use this URL:
```
https://www.denkmal.dev/resource-layout--de--35--1428164686--manifest.json
```


njam:

vogdb:
lgtm

vogdb:
btw front-end is lgtm. Didn't dig deep into back-end.

tomaszdurka:
Woah, was not suspecting that some files really need to be served from root.
This manifest.json seems really problematic.

njam:
The manifest.json doesn't need to be on root, but a JS service worker has to!
@tomaszdurka agree on these changes?

tomaszdurka:
Well, I guess we have no choice! 
Looks ok.

njam:
:)

PR #1720, opened 2015-04-10
Move GA tracking from end of <body> to beginning

njam:
@tomaszdurka please review

This is according to GA universal tracking guidelines.
Problems could arise if we'd modify the tracking instance during page rendering. But as far as I can see we only do so in `prepareResponse()`, so it should be fine.

### DIFF
--- a/layout/default/Layout/Abstract/default.tpl
+++ b/layout/default/Layout/Abstract/default.tpl
@@ -54,6 +54,8 @@
     {block name='head'}{/block}
   </head>
   <body id="{$viewResponse->getAutoId()}" class="{$viewResponse->getCssClasses()|implode:' '}">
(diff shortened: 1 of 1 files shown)


fauvel:
Lgtm!

tomaszdurka:
As discussed looks ok for now.

fauvel:
`CM_Service_Tracking_ClientInterface::trackSplittest()` would not work any more if split tests are being used during page rendering (called directly from `CM_Model_Splittest::_getVariationFixture()`). But it's not implemented anyway in `SKService_GoogleAnalytics_Client` at the moment, only in `CMService_KissMetrics_Client`. So it's not an issue for now.

njam:
1 build failed due to flaky test, succeeds locally and in second travis worker, merging..

PR #1725, opened 2015-04-16
Upgrade autosize plugin to version 3

vogdb:
See https://github.com/jackmoore/autosize/releases

We're still using the version 1 (which is wrapped in a jQuery plugin). The new version 3 seems to close many bugfixes and doesn't come as a jQuery plugin anymore.

@vogdb When you have a moment could you check how we can best include the new version?
cc @christopheschwyzer


vogdb:
The easy-kick upgrade didn't work. Need to investigate.

vogdb:
Btw. The problem is that initial height is incorrect.

vogdb:
Depends on jackmoore/autosize#217

tomaszdurka:
Looks ok, never sure why these options are prefered over normal params.
Let's not look into this now.

Issue #1719, opened 2015-04-08
Verify efficiency of hidden img (email opening tracking)


ppp0:
- A/B test with 
  - no attributes (Currently we `display:none` and size 0 the image)
  - unescaped IMG SRC url (currently we escape `&` to `&amp;` in the mail)

ppp0:
SendGrid does it like this
```
html
<img src=3D"https://u1703816.ct.sendgrid.net/wf/open?upn=3DEOpstEwxYPZNS7zF=
upvnSC9k98Qn5BFwyl28xsdTQETIHdd9IB-2BvPpFpv-2B-2Bu0rQP3hxZb30-2BGiPy5Blt3CK=
S4-2FALbcRI2VsmOMJTgDifM-2BDehz-2BvPIZKhFRTvFYtZ0IHdzuXQdXPKLreSYdLpvpLl1on=
hF-2B0UTXcgdtILtZDHGpji-2Fkl-2F6QOTyDemrxw7Bgq-2B5D5-2B34KRWtR31lyRaqMjzoWs=
vZ5uSRK8IYfm5RkI9g-3D" alt=3D"" width=3D"1" height=3D"1" border=3D"0" style=
=3D"height:1px !important;width:1px !important;border-width:0 !important;ma=
rgin-top:0 !important;margin-bottom:0 !important;margin-right:0 !important;=
margin-left:0 !important;padding-top:0 !important;padding-bottom:0 !importa=
nt;padding-right:0 !important;padding-left:0 !important;"/>
```

njam:
I'd go with Google here and use a *naked* img-tag :)

ppp0:
DynECT does it like this:
```html
<img src="http://trk.email.dynect.net/trk.php?a=/o2/575/i5UKlPjcTsp07wz%2BZ9OiDsJGvWitMIFuIlqoD%2F%2B%2Fpq8cuVqTHYov26SCvklj4jhlGegJkFn32bNcjhMH4QW1x9g%2F4LVQeQsW4YYwy2bbV6l4ZF%2BoxrmQeQ%3D%3D/i.gif&i=20150408142332.0D6F0308FAB4%40mail6-07-ewr.dyndns.com&x=" alt="" width="2" height="2" />
```

ppp0:
Tests speak in favour of adding dimension attributes (e.g. 1x1)

njam:
@ppp0 closeable?

ppp0:
yes

[...]

PR #1726, opened 2015-04-16
Move mail img tag generation to protected function

ppp0:
Part of #1719 

For the split-test, generation of a different image in a derived class (in sk)


ppp0:
@njam please review

njam:
lgtm

PR #1729, opened 2015-04-17
Only enable split tests for new users

fauvel:


fauvel:
@njam please review

njam:
- CM/Model/Splittest.php::_getVariationFixture: Shouldn't the skipping logic rather to `_isVariationFixture`, then we don't have to return empty string. 
 - Maybe also move this logic to `CM_Model_Splittest_RequestClient` and `CM_Model_Splittest_User` so we don't need the `switch`?

fauvel:
> CM/Model/Splittest.php::_getVariationFixture: Shouldn't the skipping logic rather to `_isVariationFixture`, then we don't have to return empty string. 

I think it's better for unit testing / mocking if we have it in `_getVariationFixture`. But I could returning `null` for instance. Wdyt?

njam:
But the public API only exposes the `is*` functions (as we don't wanna allow to retrieve the actual variation).
Why is it better to put it into `get*` for testing? We could still mock either `_isVariationFixture()` or even `_getVariationFixture()` when testing the public interface `isVariationFixture`?

fauvel:
I agree it looks cleaner like this... It is a bit strange that we have model fields in `CM_Model_Splittest` that are only used in its implementations, but ok.

fauvel:
@njam please review!

njam:
lgtm

fauvel:
Tagged 1.114.2

PR #1736, opened 2015-04-17
Add deprecation warning about Request::getInstance()

njam:
Related to #1686


fauvel:
Ok... quoting @njam for the records:

> There are a few related problems and solutions though:
- For Splittests: Use the request in `prepareResponse()` - see #1376
- For GA: Don't use the request-clientId - see https://github.com/cargomedia/CM/pull/1733
- In general: Get rid of singleton request - see https://github.com/cargomedia/CM/pull/1686

PR #1737, opened 2015-04-18
Add optional getter closure for Cache::get()

njam:
To simplify cache usages for when the cache-key is used in only one function or class.
Together with the existing `Cache::key()` function some code could be changed from:
```php
$cache = CM_Cache_Shared::getInstance();
$cacheKey = CM_CacheConst::Some_Constant . '_bar:' . $bar;
if (false === ($result = $cache->get($cacheKey))) {
    $result = 12;
    $cache->set($cacheKey, $result);
}
```
to:
```php
$cache = CM_Cache_Shared::getInstance();
$result = $cache->get($cache->key(__CLASS__, 'foo', $bar), function () {
    return 12;
});
```

@tomaszdurka @dlondero @mariansollmann please review

### DIFF
--- a/library/CM/Cache/Abstract.php
+++ b/library/CM/Cache/Abstract.php
@@ -5,12 +5,28 @@ abstract class CM_Cache_Abstract extends CM_Class_Abstract {
     /** @var CM_Cache_Storage_Abstract */
     protected $_storage;
 
-    public function __construct() {
-        $storageClassName = static::_getConfig()->storage;
-        if (!is_subclass_of($storageClassName, 'CM_Cache_Storage_Abstract')) {
-            throw new CM_Exception('Invalid cache storage: `' . $storageClassName . '`');
+    /** @var int */
+    protected $_lifetime;
+
+    /**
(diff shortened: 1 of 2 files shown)


tomaszdurka:
I like that. I recall that idea from long time ago, what happened that you decided to implement it?

What about using array as key? Could be bit more *magical*:

```php
$cache = CM_Cache_Shared::getInstance();
$result = $cache->get([__CLASS__, 'foo', $bar], function () {
    return 12;
});
```

njam:
I think we should still use `$cache->key()`, as it transforms non-scalar values to scalar (with md5)?

Marian and Daniel need some one-point caching in a few models, I think with this approach it can be simplified.
@mariansollmann @dlondero wdyt?

dlondero:
@njam looks very good as there's no need to set a `CM_CacheConst` for the usages you're referring to and this solutions helps a lot. Afaiu the closure should return the value we want to set into the `key` if not already set, correct?

njam:
exactly!

I'm gonna merge this then.

dlondero:
Thanks!

PR #1732, opened 2015-04-20
Allow to set terminationCallback without using waitForChildren()/listenForChildren()

tomaszdurka:


tomaszdurka:
@njam Please review. There is a slight change to use EventHandler for handling process events. Not sure about that, as pcnt_singal has this special callback SIG_DFL which I cannot reproduce using EventHandler. Is this good idea or rather revertable?

njam:
I think that's a good idea!

What `pcntl_signal(<some signal>, SIG_DFL);` does is to reset any signal handling (dfl = default). This is needed because child processes "inherit" installed signal handlers from the parent process, so we want to *reset* that after forking.

- ProcessTest: Let's also add a test with sending a TERM signal to the process (instead of triggering an event) - as this is the real-world use case!

tomaszdurka:
Well, my idea was to replace signal handlers with event-handler and then simply unbind when we fork (create new process), but unfortunately tests are failing with that. Seems we need that SIG_DFL - do you know what does it do?

tomaszdurka:
Well, idea is considered "not the best" as it needs to use both models pcntl_handler and event-handler, wdyt @njam?

njam:
If you call `pcntl_signal(SIGTERM, SIG_DFL);` it resets a previously set signal handler, e.g. `pcntl_signal(SIGINT, $handler, false);`. So `_reset()` reverts what was done in `_installSIgnalHandlers()` in the children.

You probably would have the same issue when using an *event bus* - the child process should unbind from any signal-events that it inherits from the parent?

I'm okay with reverting back to just plain signals though.

tomaszdurka:
Hmm, I understand, but that was the problem - unbinding all events still resulted with different behaviour.
I will investigate and paste results in free time.

tomaszdurka:
I looked into the problems I've encountered and seems there is some strange (imo undesired) implementation of the singal handlers. 

Once we instantiate process, we installSignalHandlers which are replacing default ones. This means in case of process being killed, it won't execute any code, but silently continue...

I think additionally to binding to events we should throw exception that process has exited.
@njam thoughts?

njam:
As discussed:
- Let's *use* eventHandler
- Only install signal handlers when needed (event binding?)
- Only bind *killChildren* on `fork()`

tomaszdurka:
@njam Please take a final look. I am pretty much happy with current implementation.
Wdyt?

njam:
- CM_Cli_CommandManager::unlockCommand: Can we now make it protected?
- CM_ProcessTest: Add a test with binding on `exit`?
- CM_Process: Remove `$_terminationCallback`
- CM_Process::trigger: Don't the `$arguments` include `$event` too now?
 - Cover in tests
- CM_Process::fork: What if you fork once, then the child terminates and then you fork again - won't we bind to `exit` twice then?

tomaszdurka:
- I'll try :)
- Ok
- Ok
- Oh, right...
- We will. I remeber I thought about unbinding there. But it's not so easy. Should we have some event namespaces? :)

njam:
> I remeber I thought about unbinding there. But it's not so easy.

How about just a boolean flag to keep track of binding...?

tomaszdurka:
Depends on 1732

tomaszdurka:
@njam Please review

njam:
Hmm, but unbinding in `_wait()` won't catch all cases?
I mean one could `_fork()`, and then *not* wait for children (but still have a child exit), in which case the even won't be unbinded and we will end up with two bindings?

tomaszdurka:
Well, that's true we don't really have information when child process exits, unless we use wait. 

On the other hand it wil work well, as we did not as unbind and won't bind again as there are forks on the CM_Process.... which is bad, but this in general is bit strange.

I think concept used in CM_Process is two-level, structural way and is getting hard to overview and maintain when processes are nested multiple times. Fork could potentially return some kind of CM_Process instance and use the same bindings.  Parent could listen to those child events. Neverthless we probably don't want to implement it this way as it triggers quite refactoring.

njam:
That is true of course!
I'm okay with this approach then!

tomaszdurka:
@njam Ive added unbind also when manually killing children...

njam:
makes sense

PR #1741, opened 2015-04-22
assertFormResponseSuccess should also check assertViewResponseSuccess

dlondero:
also consider assertFormResponseError


dlondero:
@njam please review

njam:
Can you elaborate on why `assertFormResponseError` calls `assertViewResponseSuccess`? When we use it do we expect the response to generally succeed, and to contain form errors?

If so, then all good

dlondero:
That's correct: we expect the response to be successful but with form errors.

njam:
ok, all good then!

PR #1745, opened 2015-04-24
Move GA tracking into <head> tag

njam:


njam:
There's contradictory information about where GA tracking should be added. In GA admin they say to the beginning of `<body>`, in [GA help](https://support.google.com/analytics/answer/1008080?hl=en) they say to the end of `</head>`.
AdWords support recommended to put it into `<head>`.

@fauvel please review

fauvel:
Lgtm, and both versions probably work equally well "in the wild".

PR #1752, opened 2015-04-28
Implement adding custom headers to mail

ppp0:


ppp0:
@njam I am not sure if this is considered breaking as it adds a field to table `cm_mail`. I don't know how tolerant the code is towards missing db fields in this case? If not, we should prepend a `Breaking: ` to the issue's title

njam:
- Adjust `structure.sql` (used for new installations, and also in tests)
- addCustomHeader: Add docu
- addCustomHeader: Prefer `null === $foo` instead of `is_null()`
 - Usually for cases like these what I use is: `if (null !== $foo) { $foo = (int) $foo; }`
- _send: This call to `_addCustomHeaders()` looks strange
- Add a test!

ppp0:
- Ok
- Added to private method _addCustomHeaders (Is this what was meant?)
- Ok
- Agree - I moved adding the custom header `X-MDA` to the constructor
- Sure thing!

@njam please review

ppp0:
Travis fails, successful on VM
(CM_MailTest)
```
Time: 998 ms, Memory: 20.25Mb

OK (10 tests, 26 assertions)
```

njam:
- CM_Mail::_send: Has a bug - see IDE warning (yellow) - 
- CM_Mail::addCustomHeader: If condition should be not-equals
- Move MDA-header to send (don't serialize)
- CM_Mail: Use type hinting when possible
- CM_Mail: Check PhpMailer/CM_Mail confusion

ppp0:
- true :)
- oops
- ok
- not needed anymore
- nna

@njam please re-review

njam:
- `addCustomHeader` accepts `$value = null` - is this on purpose? In `_send` it will be converted to empty string.

Rest looks good

ppp0:
I can remotely imagine a usage where you just add a header's label (field name in https://tools.ietf.org/html/rfc2822#section-2.2) and no value.. If you read the specs, there should be also a limit to charset 7-bit ASCII or even less. I'd say we can skip this as we do not sanitize strings in other places either. @njam wdyt?

njam:
> A field name MUST be composed of printable US-ASCII characters

So it should be a string (even an empty one)? So how about just expecting string?

ppp0:
>field name

not field value!

ppp0:
as discussed @njam please final check

njam:
sorry, copied wrong sentence :)
> A field body may be composed of any US-ASCII characters

lgtm, please merge upstream

ppp0:
`CM_MailTest` succeeding on VM:
```
OK (10 tests, 27 assertions)
```

PR #1758, opened 2015-05-04
CM_Mail: Debug appending key/value to array

ppp0:


ppp0:
@njam please review

tomaszdurka:
Looks ok, but...

You're dropping ability to pass multiple headers with the same. According to http://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html#sec4.2 it is possible to send headers with the same name as single with values separated by commas. Should we support this? I vote for no.

The only misleading thing now name of the method. We should probably rename it to "setCustomHeader" as it will actually replace previoous one (not add).

njam:
Agree, I would keep the ability to set multiple headers with the same key.

ppp0:
I gather from this:
- @tomaszdurka you are ok with the current solution (unique header name) but want a renaming of the setter
- @njam you agree but would like to keep setting multiple headers with the same name

this is an interest conflict..

I vote for multiple headers with the same name, which is standard.. in order to do that, I'd refactor to encaspulate every single k/v pair into a single array (and therefore have an array of arrays). I need to refactor the "explosion" into single headers - easy :v:

ppp0:
@tomaszdurka @njam who wants to review?

njam:
Philipp and me are looking into this now

tomaszdurka:
I would prefer to not keep array-like types with 2 values. Rather:
```php
if (!isset($this->_customHeaders[$label])) {
  $this->_customHeaders[$label] = array();
}
$this->_customHeaders[$label][] = $value;
```

tomaszdurka:
@ppp0 I just meant that if we agree on single-header then the name sould be set* as it's replaces previous value.

ppp0:
@tomaszdurka @njam I value your comment aka please review

tests/library/CM/MailTest.php
```diff
@@ -74,28 +77,64 @@ public function testSetText() {
         $this->assertSame($text, $mail->getText());
     }
 
+    public function getMockPhpMailer() {
+        return $this->_mockPHPMailer;
```
mariansollmann: If you have a private variable why you need this getter?

tests/library/CM/MailTest.php
```diff
@@ -74,28 +77,64 @@ public function testSetText() {
...
-            $this->fail('Sent mail without recipient');
-        } catch (Exception $e) {
-            $this->assertContains('No recipient specified', $e->getMessage());
-        }
-        $mail->addTo('tomasz@durka.pl');
-        $mail->setSubject('foo');
-        $mail->setText('bar');
-        $this->assertSame(0, CM_Mail::getQueueSize());
+        CM_Config::get()->CM_Mail->send = true;
+        $this->_mockPHPMailer = $this->mockObject('PHPMailer');
```
mariansollmann: I think this should go to the `setUp`, wdyt?

tests/library/CM/MailTest.php
ppp0: I call this method from another class `Test_CM_Mail` (line 190++) - for this I need the getter?

tests/library/CM/MailTest.php
```diff
@@ -74,28 +77,64 @@ public function testSetText() {
...
+                $this->assertSame('bar,foo', $value);
+            })
+            ->at(1, function ($name, $value) {
+                $this->assertSame('X-Bar', $name);
+                $this->assertSame('foo', $value);
+            });
+
+        $sendMethod = $phpMailer->mockMethod('Send');
+
+        $mail = new Test_CM_Mail('foo@example.com', null, null, $this);
```
mariansollmann: Why you are using a new class? Couldn't you just mock `CM_Mail` and "overwrite" `_getPHPMailer` with your mock `_mockPHPMailer`, wdyt?

ppp0:
@mariansollmann fancy reviewing please?

mariansollmann:
looks good

njam:
Why is it now called `setCustomHeader()`, isn't it *adding* a header (i.e. we can have twice a header with the same name)?

ppp0:
It is overwriting the header's key, though

njam:
...but only if it's the *same*, so not really overwriting :)

PR #1757, opened 2015-05-05
CM_Util::varDump: Show contents of CM_Params

njam:
`CM_Util::varDump()` is used to format arguments in stack traces (`CM_ExceptionHandling_SerializableException::_extractTraceRow()`).

Currently it handles all object the same, except `CM_Model_Abstract` (where it prints the ID). It would be useful if we would show the contents of `CM_Params` to change this line:
```
CM_Jobdistribution_Job_Abstract->_executeJob(SK_Params)
```
into something like:
```
CM_Jobdistribution_Job_Abstract->_executeJob(['foo' => 12])
```

We could add another `if` to `varDump()` or maybe even allow to override that behaviour on the class itself.
@alexispeter @tomaszdurka wdyt?

### DIFF
--- a/layout/default/Component/LogList/default.tpl
+++ b/layout/default/Component/LogList/default.tpl
@@ -41,7 +41,7 @@
           {foreach $log.metaInfo as $key => $value}
             <div class="tableField clearfix">
               <div class="label">{$key|escape}</div>
-              <div class="value">{$value|@varline|escape}</div>
+              <div class="value">{$value|escape}</div>
             </div>
           {/foreach}
         </div>
--- a/library/CM/Debug/DebugInfoInterface.php
+++ b/library/CM/Debug/DebugInfoInterface.php
@@ -1,6 +1,6 @@
 <?php
 
-interface CM_DebugInfo_DebugInfoInterface {
+interface CM_Debug_DebugInfoInterface {
 
     /**
      * @return string
--- a/library/CM/DebugInfo/DebugInfoInterface.php
+++ b/library/CM/DebugInfo/DebugInfoInterface.php
@@ -1,0 +1,9 @@
+<?php
+
+interface CM_DebugInfo_DebugInfoInterface {
+
+    /**
+     * @return string
+     */
+    public function __debugInfo();
+}
--- a/library/CM/DebugInfo/DebugInfoInterface.php
+++ b/library/CM/DebugInfo/DebugInfoInterface.php
@@ -5,5 +5,5 @@ interface CM_DebugInfo_DebugInfoInterface {
     /**
      * @return string
      */
-    public function __debugInfo();
+    public function getDebugInfo();
 }
--- a/library/CM/Exception.php
+++ b/library/CM/Exception.php
@@ -70,16 +70,10 @@ class CM_Exception extends Exception {
     }
 
     /**
-     * @param boolean|null $raw
-     * @return string[]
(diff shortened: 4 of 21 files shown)


njam:
@alexispeter @tomaszdurka please review
Some controversial changes here:
- I changed the `metaInfo` on both `SerializableException` and `Paging_Log` to be one-dimensional. I.e. it's an array of strings. This simplifies a few parts of the code substantially imo.
- Add a "recursive" option for `varDump()`
- Introduce interface `CM_DebugInfo_DebugInfoInterface` for `__debugInfo()`, which is called from `varDump()`

tomaszdurka:
I looked into the implementation quite a bit.

So varDump is diviced into two parts. First part of logic is included in the varDump itself where it's checking the type of variable. Other part is within DebugInfo interface, specifically for custom-class-objects. This seems fine.

Idea was to use `__debugInfo` magic method which is also used for built-in `var_dump` method. Actually mysefl brought this functionality up, but now I am not sure anymore. While using `var_dump` we would probably expect some other (more) data to be printed, e.g $data variable on all the models (?). We use our `varDump` in specific context - mostly backtraces, we want data to be very minimalistic. It probably also depends on the context.

I thought now these might be two different use cases and maybe `DebugInfo` interface should not use `__debugInfo` magic method, but rather different, custom one. 

wdyt?

njam:
I agree - renamed to `getDebugInfo()`.

@tomaszdurka wdyt? Rest of the changes are good?

tomaszdurka:
Yes, I think it's good.

One minor thing, I thought that all of the debug code (maybe also logging) could reside in one module at some point. What about calling it `CM_Debug` and placing newly created interface there?

njam:
Agree, @tomaszdurka please check last commit

tomaszdurka:
Looks good. 

Side note from hip-chat discussion: *maybe it's good moment to move CM_Util::varDump() into something like CM_Debug_VariableInspector->dump()*

alexispeter:
> I changed the metaInfo on both SerializableException and Paging_Log to be one-dimensional. I.e. it's an array of strings. This simplifies a few parts of the code substantially imo.

What there a reason for this other than simplification of the code? I'd like to include an array in the meta-info but it just looks like this now.
```
CM_Exception_InvalidParam: Not enough parameters in /home/vagrant/cm/library/CM/Params.php on line 229
    0. {main} /home/vagrant/cm/scripts/test.php:0
    1. CM_Params->getLocation('location') /home/vagrant/cm/scripts/test.php:10
    2. CM_Params->getObject('location', 'CM_Model_Location') /home/vagrant/cm/library/CM/Params.php:336
    3. CM_Params->{closure}('CM_Model_Location', []) /home/vagrant/cm/library/CM/Params.php:242
parameters: []
class: 'CM_Model_Location'
```

njam:
> I changed the metaInfo on both SerializableException and Paging_Log to be one-dimensional. I.e. it's an array of strings. This simplifies a few parts of the code substantially imo.

> What there a reason for this other than simplification of the code? I'd like to include an array in the meta-info but it just looks like this now.

@alexispeter your problem was resolved?

One thought was that using a one-dimensional data structure will make it easier to add search..

alexispeter:
You could say it's resolved because it is still possible to store 2dimensional metaInfo and the log-cmp still shows it correctly.

PR #1762, opened 2015-05-05
Make flushing split tests more robust under load

fauvel:
Paired with @kris-lab 

@njam the problem arises when you flush a split test while there is such a high load on the master database that the proxy switches to an outdated slave, so that you put the "old" flushStamp into the cached split test model.

After the database recovers, the flushStamp you put into the variation fixtures is the new one, and you get no chance to match the split test's flushStamp again, since it has the old one in the cache.

So invalidating the cached split test data should help!

### DIFF
--- a/library/CM/Model/Splittest.php
+++ b/library/CM/Model/Splittest.php
@@ -254,6 +254,7 @@ class CM_Model_Splittest extends CM_Model_Abstract implements CM_Service_Manager
                 $this->getServiceManager()->getTrackings()->trackSplittest($fixture, $variation);
             } catch (CM_Db_Exception $exception) {
                 $variationListFixture = $this->_getVariationListFixture($fixture, true);
+                $this->_change();
                 if (!array_key_exists($this->getId(), $variationListFixture) ||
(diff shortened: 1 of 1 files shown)


fauvel:
@njam restarted ci. please review!

njam:
The proxy is only used for "slave" queries - i.e. if we use `execRead()`. Otherwise the query will always go to the master.
I could still imagine that under high load the wrong `createStamp` is stored in cache, due to concurrency.

So the change looks good.

I needed quite a bit to understand the code in `_getVariationFixture()`. How about extracting the retrieving and checking of the `$variationListFixture` into a function or closure, to label it and reduce duplication?
E.g.:
```php
        function(CM_Splittest_Fixture $fixture) {
            $variationListFixture = $this->_getVariationListFixture($fixture);
            if (!array_key_exists($this->getId(), $variationListFixture) || $variationListFixture[$this->getId()]['flushStamp'] != $this->getCreated()) {
                return null;
            }
            return $variationListFixture;
        };
```

fauvel:
Agree, please review!

njam:
lgtm

PR #1763, opened 2015-05-05
Allow mongoids for models in Schema Definition

tomaszdurka:


tomaszdurka:
@njam please review

tomaszdurka:
Tbh I am not super happy about how model-type is validated in schema defintiion. We had constraints about integer ids, also we still have constraint about id field in multi-value (array-like) keys. Additionally `getId()` is cast to integer and needs to be changed in the model for mongodb-models. Shouldn't `id` be defined in the schema too?

njam:
I thought you maybe change your git config before committing..

Yes, configuring the ID format in the schema would probably be good! Open a ticket for that?
For now it looks good to me.

PR #1767, opened 2015-05-07
Clear cache between "unload" and "load" to prevent using old cache entries during "load"

njam:
Otherwise setupScripts might use cached entries from previous installation and fail.


fauvel:
Sounds reasonable!

tomaszdurka:
ok

PR #1754, opened 2015-05-07
CM_App.ajax: Return usable Promise

vogdb:
Currently `CM_App.ajax()` returns the original jqXHR "Deferred" object.
It only rejects if the ajax request fails (e.g. network problem or non-200 response). But if we experience an expected application error (e.g. "auth required") it will *resolve*.

It would be great if `ajax()` would return a usable "Promise", that:
- Resolves when the request finished successfully (like `success` callback)
- Rejects when the request fails (like `error` callback)

We should also consider using [ES6's Promise](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Promise) interface instead of jQuery's Promise: https://www.promisejs.org/#jquery

@vogdb @tomaszdurka thoughts?


vogdb:
@tomaszdurka please look at bluebird implementation of promise.

tomaszdurka:
Drafting the solution
```js
console.clear();

var CustomError = function() {
    return Error.apply(this, arguments);
};
CustomError.prototype = Object.create(Error.prototype);
var custom = new CustomError();

var NetworkError = function() {};
var AppError = function() {
    this.name = 'AppError';
};
AppError.prototype = Object.create(Error.prototype);

var ajax = function(url) {
    var ajaxPromise = new Promise(function(fulfil, reject) {
        setTimeout(function() {
            //reject('network');
            fulfil({success: 'content is king'});
            fulfil({error: 'not authorized'});
        }, 1000);
    });
    return ajaxPromise.then(function(response) {
        if (response) {
            if (response.error) {
                throw new AppError(response.error);
            } else {
                return response.success;
            }
        }
    });
};

var loadComponent = function() {
    return ajax('/page')
    .then(function(content) {
        console.log('showing content:', content);
    })
    .catch(AppError, function(error) {
        console.log('error from loadPage', error);
        throw error;
    });
};


var loadPage = function() {
    return loadComponent()
    .then(function() {
        throw 'boom';
        console.log('running');
    }).catch(AppError, function(error) {
        console.log('custom handler', error);
    });
};

loadPage();
```

njam:
@tomaszdurka it's missing the global handler I think?

vogdb:
@tomaszdurka @njam here is the updated version http://jsfiddle.net/13u1movy/2/. Please review.

```js
console.clear();

function AppError() {
    Error.apply(this, arguments);
}
AppError.prototype = Object.create(Error.prototype);
AppError.prototype.constructor = AppError;

//var appError = new AppError()
//console.log('custom instanceof CustomError: ' + (appError instanceof AppError));
//console.log('custom instanceof Error: ' + (appError instanceof Error));    

window.addEventListener("unhandledrejection", function(e) {
    e.preventDefault();
    var reason = e.detail.reason;
    var promise = e.detail.promise;
    console.log(reason instanceof AppError);
    //console.log(promise);
});

var ajax = function(url) {
    var ajaxPromise = new Promise(function(fulfil, reject) {
        setTimeout(function() {
            //reject('network');
            //fulfil({success: 'content is king'});
            fulfil({error: 'not authorized'});
        }, 1000);
    });
    return ajaxPromise.then(function(response) {
        if (response) {
            if (response.error) {
                throw new AppError(response.error);
            } else {
                return response.success;
            }
        }
    });
};

var loadComponent = function() {
    return ajax('/page')
    .then(function(content) {
        console.log('loadPage success', content);
    })
    .catch(AppError, function(error) {
        console.log('loadPage AppError', error);
        throw error;
    });
};


var loadPage = function() {
    return loadComponent()
    .then(function(content) {
        console.log('loadComponent success', content);
    }).catch(AppError, function(error) {
        console.log('loadComponent AppError', error);
        throw error;
    });
};

loadPage();
```

vogdb:
Very interesting solution with [vow](https://github.com/dfilatov/vow) http://jsfiddle.net/hc36ztuo/1/. No need for global handler.
```js
console.clear();

function AppError() {
    Error.apply(this, arguments);
}
AppError.prototype = Object.create(Error.prototype);
AppError.prototype.constructor = AppError;


var ajax = function(url) {
    var ajaxPromise = new vow.Promise(function(fulfil, reject) {
        setTimeout(function() {
            //reject('network');
            //fulfil({success: 'content is king'});
            //fulfil({error: 'not authorized'});
        }, 1000);
    });
    return ajaxPromise.then(function(response) {
        if (response) {
            if (response.error) {
                console.log('ajax AppError');
                throw new AppError(response.error);
            } else {
                console.log('ajax success');
                return response.success;
            }
        }
    }).catch(function(error){
        console.log('ajax network error');
        throw error;
    });
};

var loadComponent = function() {
    return ajax('/page')
    .then(function(content) {
        console.log('loadComponent success', content);
    })
    .catch(function(error) {
        console.log('loadComponent error', error);
        throw error;
    });
};


var loadPage = function() {
    return loadComponent()
    .then(function(content) {
        console.log('loadPage success', content);
    }).catch(function(error) {
        console.log('loadPage error', error);
        throw error;
    });
};

loadPage().then(function(content){console.log('### huge success')});
```

vogdb:
Some research about other libraries. Criteria: work in web-browser, *es6* syntax compatible and `finally` method.

- [RSVP](https://github.com/tildeio/rsvp.js). We can try it although it does not have methods to check state of promise like `isRejected`.

- [when](https://github.com/cujojs/when). Can be tried. The minus is that it introduces its custom syntax of creating Promise. To use it in *es6* compatible way we must use a special [shim](https://github.com/cujojs/when/blob/master/docs/es6-promise-shim.md). The problem is that this shim does not offer `finally` functionality. Honestly I will not insist on testing it.

- [Q](https://github.com/kriskowal/q). Can be tried but has its own syntax of creating Promise. No *es6* compatibility by default https://github.com/kriskowal/q#using-qpromise.

- [davy](https://github.com/lvivski/davy). Can be tried. It introduces *es6* compatible promises just under another namespace `Davy`. Has `throw` method besides `catch`. Very interesting to try.

- [promiz](https://github.com/Zolmeister/promiz). Ignore it. A very simple implementation. Does not have `finally`.

- [promise](https://github.com/then/promise). Ignore it. Does not have `finally`.

- [promiscuous](https://github.com/RubenVerborgh/promiscuous). Ignore it. Does not have `finally`.

- [lie](https://github.com/calvinmetcalf/lie). Ignore it. Does not have `finally`.

- [P](https://github.com/rkatic/p). The short version of Q. Has the same problems as Q.

- [kew](https://github.com/Medium/kew). The short version of Q. Has the same problems as Q.

- [deferred](https://github.com/medikoo/deferred). Inspired by Q. Has the similar syntax and hence not *es6* compatible by default.

- [MicroPromise](https://github.com/kaerus-component/uP). We can't use it as is. Even it states that it can be used in web, its implementation depends on `require` functionality which we don't have. https://github.com/kaerus-component/uP/blob/master/index.js#L13

njam:
@tomaszdurka @vogdb fyi http://pouchdb.com/2015/05/18/we-have-a-problem-with-promises.html

vogdb:
I think we have a problem here. I still can not find any info about rejecting/resolving/cancelling of promise from outside of it. So, no matter what happened, the promise `then`, `catch` always will be executed. That makes our usecase is cumbersome. I mean the usecase when we want to ignore one conversation loading in favor of another conversation loading because user changed his mind and picked another conversation while the first one hasn't been loaded yet.
The only solution out of the box is a bluebird's cancellation.
We also can do something like this:
```js
    var ajaxPromise = new Promise(function(fulfil, reject) {
        setTimeout(function() {
            if (!ajaxPromise.cancelled) {
              fulfil({success: 'content is king'});
            }
        }, 1000);
    });
```
But it's a half-measure cause the promise will not be resolved/rejected/cancelled immediately, only its `then`, `catch` execution will be skipped. Plus in case when we have a lot of promises, we need to mark and store them in order to access them from inside of themselves. Such storing mechanism will introduce a significant complication to the whole promise implementation. Like this:
```js
    var ajaxPromise = (function () {
        var promise = new Promise(function(fulfil, reject) {
          setTimeout(function() {
              if (!promise.cancelled) {
                fulfil({success: 'content is king'});
              }
          }, 1000);
        });
        return promise;
    }());
```

So, we have no choice but bluebird only =(

tomaszdurka:
I think it's ok to use bluebird with its functionality for the moment.
On the other hand from time to time I think we might expect more than we should from Promise pattern and rather implement something on top of it.

library/CM/Form/Abstract.js
```diff
@@ -172,8 +170,7 @@ var CM_Form_Abstract = CM_View_Abstract.extend({
     }
 
     if (_.size(errorList)) {
-      deferred.reject();
-
+      return Promise.resolve();//TODO is this replacement ok. It is not rejected but resolved with empty value.
```
vogdb: @tomaszdurka please check this TODO comment.

library/CM/Form/Abstract.js
```diff
@@ -194,8 +191,7 @@ var CM_Form_Abstract = CM_View_Abstract.extend({
               }
             }
 
-            handler.trigger('error error.' + action.name);
-            deferred.reject();
+            throw cm.error.create({msg: 'error error.' + action.name, type: null, isPublic: true});//TODO is this replacement ok?
```
vogdb: @tomaszdurka please check.

vogdb:
For a formality. @tomaszdurka please review.

client-vendor/after-body/bluebird/config
```diff
@@ -0,0 +1 @@
+node tools/build.js --features "core cancel"
```
tomaszdurka: What is reason for adding this file? I understand these are instructions for build tool to include cancel.

I thought that our client vendor manager is very limited and that for now we only put end-files into the repo.

Wdyt?

library/CM/App.js
```diff
@@ -792,71 +807,52 @@ var CM_App = CM_Class_Abstract.extend({
...
+        if (xhr.status === 0) {
+          return; // Ignore interrupted ajax-request caused by leaving a page
+        }
 
-    return jqXHR;
+        var msg = cm.language.get('An unexpected connection problem occurred.');
+        if (cm.options.debug) {
+          msg = xhr.responseText || textStatus;
+        }
+        reject(cm.error.create({msg: msg, type: null, isPublic: false}));
```
tomaszdurka: I am not sure if I am so happy with this creation of errors.
Is this in-between solution? I would prefer to keep errors simple with message only. Maybe we could replace type with additional error classes?

library/CM/App.js
```diff
@@ -792,71 +807,52 @@ var CM_App = CM_Class_Abstract.extend({
...
-    callbacks = callbacks || {};
-    return this.ajax('rpc', {method: methodName, params: params}, {
-      success: function(response) {
+  rpc: function(methodName, params) {
+    return this.ajax('rpc', {method: methodName, params: params})
+      .then(function(response) {
         if (typeof(response.result) === 'undefined') {
           cm.error.trigger('RPC response has undefined result');
         }
-        if (callbacks.success) {
```
tomaszdurka: Shouldn't we rather throw (reject promise) here and allow global handler to take care of it? I understand it's not possible to have custom handler for `rpc` now.

client-vendor/after-body/bluebird/config
vogdb: I think https://github.com/cargomedia/CM/blob/master/client-vendor/before-body/01-modernizr/config

library/CM/App.js
vogdb: We can work on alternative solution. I made this one to be compatible with the existing error mechanism of `cm.error.trigger`. Besides the type we also need to think about `isPublic` property.

library/CM/View/Abstract.js
```diff
@@ -304,23 +290,48 @@ var CM_View_Abstract = Backbone.View.extend({
...
+   * @param {Object|Null} [params]
+   * @param {Object|Null} [options]
+   * @return Promise
+   */
+  showComponent: function(className, params, options) {
+    return this.prepareComponent(className, params, options)
+      .then(function(component) {
+        component.popOut();
+        return component;
+      });
```
tomaszdurka: I really like this distribution of the functionalities. Only thing we could change is naming of the methods.

`showComponent` sounds like it should take an existing component as argument. What about `displayComponent` or `popOutComponent`? Although I need to check behaviour of popOut.

library/CM/View/Abstract.js
```diff
@@ -653,18 +664,13 @@ var CM_View_Abstract = Backbone.View.extend({
 
   /**
    * @param {Object} response
-   * @param {Function} [successPre]
-   * @private
+   * @return CM_Abstract_View
    */
-  _injectView: function(response, successPre) {
+  _injectView: function(response) {
```
tomaszdurka: God bless you for removing this argument....

vogdb:
@tomaszdurka please review.

tomaszdurka:
Looks good now.

I am still worried shouldn't we rather deal with cancelations in different way. E.g. with setting some flag on a Promise and then checking it's value within resolve.

vogdb:
I'm afraid that the current mechanism does not allow this. Also if you get cancellation, you do not go to resolve clause => no point to wait in resolve cause resolve ain't going to be executed.

tomaszdurka:
Oh I meant to not cancel/reject as we do via bluebird, but instead set this flag to let's say `false` and then resolve would not resolve :)

vogdb:
But then you need to check for this flag in every resolve that is based on this promise?

vogdb:
IMO they are equal
```js
.catch(function(err){if(err instanceOf Cancellation){...}})
```

```js
.then(function(result){if(result.flag){...}})
```

tomaszdurka:
Looks good! Let's release this today.

 One question, are we js-catching (non-promise) all the possible exceptions properly now? It would great to see (via js log) any kind of problems related with Promises once we release. 
On the other hand ajax calls which are used for storing errors are now Promise-based.

PR #1769, opened 2015-05-08
Make Action::isAllowed more powerful

njam:
The current business logic is that if `isAllowed` returns `false` then the action is *disallowed*. In case `true` is returned the normal checks are applied (action limits).

That's quite confusing. I'd suggest to rename it to `isDisallowed()`, and maybe additionally introduce a `isAllowed()` which implements the opposite (always allow action if `true` returned).

@fauvel wdyt?


njam:
@fauvel please review

The change is pretty small. I'd recommend to look through commits, as I've refactored tests a bit.

tomaszdurka:
Hmmm, ok I understand this un-confuses a bit usage with negating `_isAllowed`?
Looked at implementation, doesnt this isAllowedOrDisallowed not overcomplicating implementation?

njam:
There are two things done here (sorry, I talked them over with Sébastien but didn't write anything down:
- **Un-confuse `isAllowed`**. The logic is actually `isDisallowed`, so I'm renaming and it and will change the implementations accordingly.
- **Introduce opposite - now `isAllowed`**: With the current approach (`isDisallowed`) we can override action limits by returning `true` in `isDisallowed`, but we cannot do the opposite. If the new `isAllowed` returns `true` it will override any action limits - the action will just be allowed. --> With those two methods in place we can finally implement all our ACL needs, which we cannot currently do. This will allow to do some followup simplifications in SK.

@tomaszdurka Does this answer your questions?

fauvel:
Honestly – I don't like it at all :) I think it's over-complicated and even more confusing than it was before. If you have a very special use case in which you want to ignore action limits, let's introduce a method `CM_Action_Abstract::ignoreActionLimits()` and call it before you prepare the action!

njam:
So you mean to remove both `_isAllowed` and `_isDisallowed` in favor of an action-limit override? Remember it will be used in both `isAllowed()` and `prepare()`

How about having a `isAllowed(): bool|null`? The return value would be interpreted as:
- `true` Force allow
- `false` Force disallow
- `null` Check action liimits

njam:
...I tink the ACL rules should be defined on the action definition itself - independent of their usage. The same action is usually used in multiple places, e.g. to only show a button if the action is allowed, and then in the ajax function to verify that the action can be executed.

fauvel:
You mean `_isAllowed*(): bool|null`? Sgtm!

tomaszdurka:
I'm also not so fond of having both `_isAllowed` and `_isDisallowed`. This brings more confusion and need for understanding how this works internally. Generally this pattern of having single implementation of public method calling various implementations of internal protected method is confusing and I would avoid that. What about simplifing this as below.

Another solution is two have base implementation of `isAllowed` with action limits and simply override method in child classes.

Force allow:
```php
return (isAllowed() || parent::isAllowed())
```

Force disallow:
```php
return (isDisallowed() && parent::isAllowed())
```

njam:
I think the exclamation mark is missing?
```php
return (!$isDisallowed() && parent::isAllowed())
```
I have the feeling this can be confusing as well, especially if the method body is a bit bigger.

The bigger problem though is that the actionlimit-check can't work solely with a *boolean result*. There are two cases where it's being used:
- `isAllowed`: here boolean is sufficient
- `prepare`: here we want to call `overshoot()` on the first action limit reached. If that doesn't throw then we throw manually.

Let me draft up `isAllowed(): bool|null` and then let's have another look?

njam:
Ok it's done. The change is now fairly trivial.

@tomaszdurka @fauvel wdyt?

library/CM/Action/Abstract.php
```diff
@@ -193,23 +202,27 @@ protected function _getActionLimitList() {
...
     protected function _isAllowed() {
         $arguments = func_get_args();
-        $methodName = '_isAllowed' . CM_Util::camelize($this->getVerbName());
 
-        if (method_exists($this, $methodName)) {
-            return call_user_func_array(array($this, $methodName), $arguments);
+        $methodNameAllowed = '_isAllowed' . CM_Util::camelize($this->getVerbName());
+        if (is_callable([$this, $methodNameAllowed])) {
+            $isAllowed = call_user_func_array(array($this, $methodNameAllowed), $arguments);
+            if (null !== $isAllowed) {
```
fauvel: Not needed I think

fauvel:
lgtm!

tomaszdurka:
Let's say it's good!

PR #1770, opened 2015-05-14
Update email template

njam:

I will then extend from this template in sk.


christopheschwyzer:
ah `function.mailButton` - cool. 
It's a bit weird that styles are defined inside the function. Maybe we should extract all button styles from `form.less` into a `button.less` and than require this file. Than again - quite different requirements for mail... 
There is also quite some duplication of styles in `mailHtml` and `layout.less` - not sure it's easy to extract.

Otherwise lgtm.

njam:
I agree the duplication between web and email styles is not so nice. But oftentimes there's very special behaviour in old email clients, and the styles (and markup) need to be adjusted (for example some clients don't have `max-width`).

If we wanted to combine the definition of web and email styles I think we would need a CSS inliner tool, which would be whole different story. I don't think it's worth doing that.

I've moved the button definition to a "helper" smarty template (instead of a function). Also I've moved the base row layout for the email into the template.
What do you think?

It looks good on all clients now: https://litmus.com/pub/6b1284f
Waiting for Windows Phone to be available on Litmus again..

christopheschwyzer:
Small detail: How about adding `text-decoration: none !important;` in `button.tpl` to prevent hover effect?! - or even do it globally for all `<a>`?

njam:
Works good, please double check!

christopheschwyzer:
lgtm!

PR #1776, opened 2015-05-25
Add user currency to frontend environment

mariansollmann:


mariansollmann:
@njam please review.

njam:
- CM_Model_Currency: How about combining `findByCountry` and `findByLocation` (keeping the latter), and then using it in `getByLocation`? I think a lot of code could be simplified/removed then.
- CM_Mail: Is there a reason to keep `_render()` or can we inline it into `render()`?

mariansollmann:
@njam please review.

mariansollmann:
@njam please review.

njam:

I reviewed the initial migration of the *currency* model from SK to CM already, but had some comments. Also additional changes were introduced since then. Anyway I think it's good if another pair of eyes takes a look at this.

alexispeter:
* library/CM/Model/Currency.php
 + create() use `@return self` in docbloc
 + findByAbbreviation() use `@return self` in docbloc
 + findByLocation() use `@return self` in docbloc
 + getByLocation() use `@return self` in docbloc
 + getByAbbreviation() use `@return self` in docbloc
 + getDefaultCurrency() use `@return self` in docbloc
 + getCurrency() I find the `withDefault` parameter a bit weird an unintuitive in such a function (simple getter). How about using a ternary operator in such cases? e.g. `$currency = $user->getCurrency() ? : CM_Model_Currency::getDefaultCurrency();
* combine update-scripts
* test that $currency->delete() invalidates `CM_Paging_Currency_All`

mariansollmann:
> create() use @return self in docbloc
> findByAbbreviation() use @return self in docbloc
> findByLocation() use @return self in docbloc
> getByLocation() use @return self in docbloc
> getByAbbreviation() use @return self in docbloc
> getDefaultCurrency() use @return self in docbloc

Not sure if this is correct. please have a look on other models (e.g. Location, Language)

@alexispeter please review.

alexispeter:
It's correct. We don't use it often as often though. What we use most is `@return static` due to polymorphism. Writing out the name like here is just as correct, but had the original implementation used `return self`, you would not have had to adjust it ;)

alexispeter:
lgtm

mariansollmann:
created tag 1.122.3

PR #1777, opened 2015-05-28
Daemon loops don't process signals

njam:
Since we don't install signal handlers any more in child processes (https://github.com/cargomedia/CM/pull/1732) our daemon loops fail to process termination signals.

E.g. `CM_Clockwork_Manager::start()`:
```php
        while (true) {
            $this->runEvents();
            sleep(1);
        }
```
A termination signal interrupts the `sleep()`, then the loop continues from the top and the signal is never processed.

Possible solutions:
- A: Install custom signal handlers also in child processes
- B: Call [`pcntl_signal_dispatch()`](http://php.net/manual/en/function.pcntl-signal-dispatch.php) on every iteration of such loops.

@tomaszdurka @kris-lab @alexispeter wdyt?

### DIFF
--- a/library/CM/Clockwork/Manager.php
+++ b/library/CM/Clockwork/Manager.php
@@ -84,9 +84,11 @@ class CM_Clockwork_Manager {
     }
 
     public function start() {
+        $process = CM_Process::getInstance();
         while (true) {
             $this->runEvents();
             sleep(1);
+            $process->handleSignals();
         }
     }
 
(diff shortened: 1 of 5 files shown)


njam:
@alexispeter please review

I recommend looking at commits (did a small refactoring first).

alexispeter:
* /library/CM/Clockwork/Manager.php
 + use _getProcess() to get the process instance
* /library/CM/Process.php
 + _wait(): replace pcntl_signal_dispatch with handleSignals()?

njam:
@alexispeter makes sense, please review!

alexispeter:
lgtm

PR #1782, opened 2015-06-01
remove <script type="text/template"> elements from DOM parsing

fvovan:
Exclude  <script type="text/template"> elements from adding to DOM with PHP DOMDocument because is was added incorrectly and broke tests.

@njam, @vogdb please review


njam:
lgtm

njam:
@vogdb we discussed this before - do you agree on this change?

vogdb:
Yes yes. A very useful feature! I need to remember also where I needed this.

vogdb:
Ah, remembered. Chat!

fvovan:
create tag 1.122.2

PR #1785, opened 2015-06-02
catch child exceptions in runWithCatching code

alexispeter:
Instead of catching only the explicitly specified exceptions, _runWithCatching() will now also catch child exceptions. 

Introduced for a new http-response that should always output json and therefore circumvent the default exception handler as much as possible.


alexispeter:
please review @tomaszdurka

library/CM/Http/Response/Abstract.php
```diff
@@ -244,14 +244,16 @@ protected function _runWithCatching(Closure $regularCode, Closure $errorCode) {
...
+            foreach ($exceptionsToCatch as $exceptionClass => $errorOptions) {
+                if (is_a($ex, $exceptionClass)) {
+                    $catchException = true;
+                    if (isset($errorOptions['log'])) {
+                        $formatter = new CM_ExceptionHandling_Formatter_Plain_Log();
+                        /** @var CM_Paging_Log_Abstract $log */
+                        $log = new $errorOptions['log']();
+                        $log->add($formatter->formatException($ex), $ex->getMetaInfo());
+                    }
+                    break;
```
tomaszdurka: You've put code which you tend to use only once into foreach loop. I guess that's comfy to do :)

What about using iteration method like `Functional\some` to determine if exception should be caught? Just like array_key_exists.

tomaszdurka:
Review done.

alexispeter:
please review @tomaszdurka

alexispeter:
please review @tomaszdurka

tomaszdurka:
Hey, I've thought I've already reviewed that :>

Will I be pain in the ass saying that actually previous version with foreach was more readable? :panda_face: I didn't notice you actually intercept errorOptions... I think it's very hacky with first(). Wdyt? I am ok with both version to not make you hop in and out.

alexispeter:
graaaaah

alexispeter:
better @tomaszdurka ?

tomaszdurka:
Yep, let's say it's good :P

alexispeter:
tagged https://github.com/cargomedia/CM/releases/tag/1.122.4

PR #1701, opened 2015-06-05
Wrong param for Params::getLocation() should raise InvalidParam

alexispeter:
Current error:
```
ErrorException: E_WARNING: Missing argument 2 for CM_Model_Location::__construct(), called in /home/example/releases/20150317115306/vendor/cargomedia/cm/library/CM/Params.php on line 216 and defined in /home/example/releases/20150317115306/vendor/cargomedia/cm/library/CM/Model/Location.php on line 19
[...]
    15. SK_Page_Users_Search->prepare(CM_Frontend_Environment, CM_Frontend_ViewResponse) /home/example/releases/20150317115306/vendor/cargomedia/cm/library/CM/RenderAdapter/Component.php:47
    16. CM_Params->getLocation('location') /home/example/releases/20150317115306/library/SK/library/SK/Page/Users/Search.php:37
    17. CM_Params->getObject('location', 'CM_Model_Location') /home/example/releases/20150317115306/vendor/cargomedia/cm/library/CM/Params.php:318
    18. CM_Params->{closure}('CM_Model_Location', array) /home/example/releases/20150317115306/vendor/cargomedia/cm/library/CM/Params.php:224
    19. CM_Model_Location->__construct(array) /home/example/releases/20150317115306/vendor/cargomedia/cm/library/CM/Params.php:216
    20. CM_ExceptionHandling_Handler_Abstract->handleErrorRaw(2, 'Missing argument 2 f...', '/home/example/relea...', 19, array) /home/example/releases/20150317115306/vendor/cargomedia/cm/library/CM/Model/Location.php:19
```

Should throw exception: `CM_Exception_InvalidParam` (like most failures in `CM_Params`)? So it can be treated a warning if coming from user input.

@fauvel fyi

### DIFF
--- a/library/CM/Params.php
+++ b/library/CM/Params.php
@@ -213,7 +213,15 @@ class CM_Params extends CM_Class_Abstract implements CM_Debug_DebugInfoInterface
     public function getObject($key, $className, $default = null, Closure $getter = null) {
         if (!$getter) {
             $getter = function ($className, $param) {
-                return new $className($param);
+                $reflectionClass = new ReflectionClass($className);
+                $countArgs = $reflectionClass->getConstructor()->getNumberOfRequiredParameters();
+                if ($countArgs > 1) {
+                    if (!is_array($param) || count($param) < $countArgs) {
+                        throw new CM_Exception_InvalidParam("Not enough parameters", ['parameters' => $param, 'class' => $className]);
+                    }
+                    return $reflectionClass->newInstanceArgs($param);
+                }
+                return $reflectionClass->newInstance($param);
             };
         }
         $param = $this->_get($key, $default);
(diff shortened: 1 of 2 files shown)


alexispeter:
please review @tomaszdurka 

This goes a bit farther than intended, in that it also allows CM_Params to instantiate objects with more than one argument. Up until now this was only possible by including a `_class`-parameter, in which case the decoding functionality of CM_Params would use fromArray() to eagerly instantiate an object inside the CM_Params object.

tomaszdurka:
This is a bit hacky - not the solution, but general approach we have there. 
I am a bit worried with the importance of the order of the parameters. I guess if you would switch level and id in your test it would fail.
Another thing is required vs optional paramters. I think it's not needed to limit them to required only, even though this is how our models mostly look like.

Wdyt about using NamedArgs here? https://github.com/cargomedia/CM/blob/master/library/CM/Util/NamedArgs.php
I think it would be perfect for fit for what you need.

alexispeter:
please review @tomaszdurka

alexispeter:
please review @tomaszdurka

tomaszdurka:
Looks good!

PR #1789, opened 2015-06-10
Allow easier exception catching for tests

tomaszdurka:


tomaszdurka:
Hey @alexispeter

I came up with this small piece of code to make testing exceptions a bit easier and more readable. 
What do you think? Please review.

tomaszdurka:
You might ask about reasons of adding this. I basically think it's pain in the ... to add `$this->fail...` all the time in try block. Here with assertion which is run no matter if code-block throws or not we don't need to add it.

tomaszdurka:
@njam fyi

alexispeter:
It takes away the verbosity that is enforced by having to provide a meaningful message for `$this->fail()` but since we don't care about this for other assertions I guess it's ok. 

lgtm

PR #1667, opened 2015-06-15
Extract image-manipulation from CM_File_Image

alexispeter:
`CM_File_Image::resizeSpecific()` uses Imagick's `thumbnailImage()`.
This function strips all exif data - also "Orientation".

Since Imagick cannot *write* changed exif information the only two solutions are:
- Use `resizeImage()` instead of `thumbnailImage()` (keeps exif data, but file is bigger)
- Rotate image data according to exif information manually.

See also http://stackoverflow.com/questions/7489742/php-read-exif-data-and-adjust-orientation

Example images:
- [rotated](https://cloud.githubusercontent.com/assets/360800/6355830/4387c938-bc59-11e4-8b79-e2dde71e921c.jpg)
- [not rotated](https://cloud.githubusercontent.com/assets/360800/6355833/4dfaf6f6-bc59-11e4-8a7f-307fdae47432.jpg)


alexispeter:
please review @tomaszdurka

library/CM/File/Image.php
```diff
@@ -100,6 +100,37 @@ public function rotate($angle, $formatNew = null, CM_File $fileNew = null) {
...
+                $this->rotate(180);
+                $this->setOrientation(1);
+                break;
+            case Imagick::ORIENTATION_LEFTTOP: // 270° flipped
+            case Imagick::ORIENTATION_RIGHTTOP: // 270°
+                $this->rotate(90);
+                $this->setOrientation(1);
+                break;
+            case Imagick::ORIENTATION_RIGHTBOTTOM: // 90° flipped
+            case Imagick::ORIENTATION_LEFTBOTTOM: // 90°
```
tomaszdurka: Hmm, not exactly sure what comments mean here, but why we rotate 90 degrees rotated image by 270 degrees? Is it different direction, why then not comment it the same way?

library/CM/File/Image.php
```diff
@@ -142,6 +173,24 @@ public function getFormat() {
         }
     }
 
+    /**
+     * @return int
+     */
+    public function getOrientation() {
```
tomaszdurka: We should probably make this non-public as it's only used internally and in tests and should not be used outside.

library/CM/File/Image.php
```diff
@@ -142,6 +173,24 @@ public function getFormat() {
...
+        return $this->_getImagick()->getImageOrientation();
+    }
+
+    /**
+     * @param int $orientation
+     *
+     * Works only on images where the meta-data already contains orientation-information
+     */
+    public function setOrientation($orientation) {
```
tomaszdurka: Same as above.

library/CM/File/Image.php
```diff
@@ -100,6 +100,37 @@ public function rotate($angle, $formatNew = null, CM_File $fileNew = null) {
         $this->_writeImagick($imagick, $format, $fileNew);
     }
 
+    public function rotateByExif() {
```
tomaszdurka: This method should/could be protected too?

library/CM/File/Image.php
alexispeter: I think it could and should be used outside in certain cases.

library/CM/File/Image.php
alexispeter: I think it could and should be used outside in certain cases.

library/CM/File/Image.php
alexispeter: This method has to be public.

alexispeter:
please review @tomaszdurka

tomaszdurka:
Ok I agree that `rotateByExif` should be public as we are calling this only for some images - not all (at least now). 
However set/getOrientation() should not be accessible in our public api. Reason for these changes is to make public api as simple as possible. In our case orientation is only concept of imagick not our `CM_File_Image`.
Currenly your only outside usage resides in SK tests checking if it really cleared the exif - this should be only tested in CM. We should get away with calling only `rotateByExif` in our outside repos. Then `set/get` should be used only internally and in tests. Do you have any other required usages?

-----------------

Name `rotateByExif` suggests it will use exif (probably even as parameter), but not clear it.
What about renaming it to `clearAndNegateExifOrientation`?

njam:
I see you're using `_writeImagick()`. How will this affect usual uploading and processing of new images? There might be quite a few HTTP requests involved.

njam:
As discussed let's split File and Image, so we can modify the image in memory:
```php
$file = new CM_File('orig.png');
$image = new CM_Image_Image($file->read());
$imageSmall = $image->clone()->resize()->rotate();
$file2->write($imageSmall->getBlob());
```

alexispeter:
please review @tomaszdurka

library/CM/Image/Image.php
```diff
@@ -9,95 +9,158 @@ class CM_File_Image extends CM_File {
...
             }
-        }, $format);
+        });
+        return $this;
+    }
+
+    /**
+     * @return $this
+     */
+    public function rotateByExif() {
```
tomaszdurka: Still name is bit unclear. You really need to know implementation to understand what it does. What about `clearAndNegateExifOrientation`?

library/CM/Image/Image.php
```diff
@@ -9,95 +9,158 @@ class CM_File_Image extends CM_File {
...
+        switch ($orientation) {
+            case Imagick::ORIENTATION_TOPRIGHT: // flipped
+            case Imagick::ORIENTATION_UNDEFINED: // undefined
+            case Imagick::ORIENTATION_TOPLEFT: // normal
+                break;
+            case Imagick::ORIENTATION_BOTTOMLEFT: // 180° flipped
+            case Imagick::ORIENTATION_BOTTOMRIGHT: // 180°
+                $this->rotate(-180);
+                $this->_imagick->setImageOrientation(1);
+                break;
```
tomaszdurka: I looked into implementation of the rotate (and other methods) and noticed that we are using rotation within `_invokeOnEveryFrame` - which makes sense.
Question:
Should orientation be also seton every frame? I guess not... it's probably generic for whole image. I ask as it's imagick instance and each frame also has imagick instance.... should we have any concern?

tomaszdurka:
Review is done. Please change name of pull-request.

Change looks good. Was suprised there are so little files to be changed. I guess it's moslty in dependant projects. 

Additional thought: It would be nice if we could separate resize and crop methods. That would involve moving some calculations outside of the class though. @alexispeter @njam wdyt?

library/CM/Image/Image.php
alexispeter: Exif is not actually supported by .gif, neither by .png for that matter. So we don't need to worry about that.

njam: There is EXIF stored in a PNG in some other meta fields afaik?

alexispeter: PNG can indeed add XMP-tags, but the usage of orientation is not specified so I wouldn't rely on it.

alexispeter:
If resize() keeps it's `$square` argument, you don't have to move any calculations outside.

alexispeter:
a91b1e8 also fixes a bug that was introduced a while ago, where resizeSpecific() would cut off only one side instead of symmetrically cutting off from both sides before resizing when calling resize() with `square`= true. 

Using `$image->resize(200, 200, true);` on the following image would result in:

im8
#### Before
im8-resized-1
Notice how the stalk doesn't lie in the center of the picture and the bright spot in the bottom-right corner of the original picture is contained in the resized picture.
#### After
im8-resized-2

alexispeter:
please review @tomaszdurka 

@njam you wanted to have a look as well regarding the bugfix

tomaszdurka:
We should add script to redo all the malformed copies in dependant projects.

library/CM/Image/Image.php
```diff
@@ -0,0 +1,385 @@
...
+    public function stripProfileData() {
+        $this->_imagick->stripImage();
+        return $this;
+    }
+
+    /**
+     * @return $this
+     * @throws CM_Exception_Invalid
+     */
+    public function validateImage() {
```
tomaszdurka: Rename to validate

library/CM/Image/Image.php
```diff
@@ -0,0 +1,385 @@
...
+     * @throws CM_Exception
+     * @throws CM_Exception_Invalid
+     */
+    public function setCompressionQuality($quality) {
+        $quality = (int) $quality;
+        if ($quality < 1 || $quality > 100) {
+            throw new CM_Exception_Invalid('Invalid compression quality `' . $quality . '`, should be between 1-100.');
+        }
+        if (true !== $this->_imagick->setImageCompressionQuality($quality)) {
+            throw new CM_Exception('Cannot set compression quality to `' . $quality . '`.');
```
tomaszdurka: Does imagick has some exceptions/errors on it's own?

library/CM/Image/Image.php
```diff
@@ -0,0 +1,385 @@
...
+    }
+
+    /**
+     * @param int       $widthMax
+     * @param int       $heightMax
+     * @param bool|null $square
+     * @return $this
+     * @throws CM_Exception
+     */
+    public function resize($widthMax, $heightMax, $square = null) {
```
tomaszdurka: What about calling this `resizeInto($width, $height)`? :panda_face:

library/CM/Image/Image.php
```diff
@@ -0,0 +1,385 @@
...
+        $square = (boolean) $square;
+
+        $width = $this->getWidth();
+        $height = $this->getHeight();
+        $dimensions = self::calculateDimensions($width, $height, $widthMax, $heightMax, $square);
+        $widthResize = $dimensions['width'];
+        $heightResize = $dimensions['height'];
+
+        if ($square) {
```
tomaszdurka: I guess square logic could be simplified to:
```php
if ($square) {
  $size = min($width, $height);
  $this->crop($width, $height);
}
```

library/CM/Image/Image.php
tomaszdurka: This could be also separate issue. Thought it's worth doing now.

tomaszdurka:
I could probably talk about signatures of some methods, but for now except small in-code comments let's say it's good :)

library/CM/Image/Image.php
alexispeter: I like this idea.

alexispeter:
please review @tomaszdurka

library/CM/Image/Image.php
alexispeter: But let's keep it for a potential follow-up

library/CM/Image/Image.php
tomaszdurka: Does it?

library/CM/Image/Image.php
tomaszdurka: What about this one?

library/CM/Image/Image.php
alexispeter: It does.

tomaszdurka: You could do a bit effort and read between the lines.
From docu I read:

```
Return Values:
Returns TRUE on success.

Errors/Exceptions:
Throws ImagickException on error.
```
So what is actually a return false...? Is this really an error or it's occuring when compression quality is the same? We have no test to cover this.

alexispeter: As you can see, reading the documentation only raises more questions. 

I have no idea when and IF this function returns a false. Maybe the true is not really necessary because there is either true or exception and never a false? False is never mentioned in the documentation after all.

alexispeter: Should we just remove this if (true !==)  and surround if with try/catch to be safe?

njam: From experience I recommend not trusting php-imagick docu.
```sh
$ php -r '$i = new Imagick(); var_dump($i->setImageCompressionQuality(999));'
PHP Fatal error:  Uncaught exception 'ImagickException' with message 'Can not process empty Imagick object' in Command line code:1
```
```sh
$ php -r '$i = new Imagick("/home/vagrant/CM/tests/data/img/test.jpg"); var_dump($i->setImageCompressionQuality(999));'
bool(true)
```
```sh
$ php -r '$i = new Imagick("/home/vagrant/CM/tests/data/img/test.jpg"); var_dump($i->setImageCompressionQuality("jo"));'
PHP Warning:  Imagick::setimagecompressionquality() expects parameter 1 to be long, string given in Command line code on line 1
```

tomaszdurka: Oh, it was not really that I thought we should ignore false value. I thought that maybe false value is returned when e.g. image is not compressed because it's already properly compressed and we would throw with no reason.

I guess the non-true return value is used when instead of exception thrown imagick is trigerring a warning (etc). And as we have warnings as exceptions we can't reproduce non-true value.

```php
vagrant@www:~$ php -r 'error_reporting(E_ALL^E_WARNING); $i = new Imagick("/home/vagrant/CM/tests/data/img/test.jpg"); var_dump($i->setImageCompressionQuality("jo"));' 
NULL
```

I would still keep throwing on non-true.

tomaszdurka:
Looks good to me.

library/CM/Image/Image.php
alexispeter: The thing is, it says returns true on success on a lot of other methods that we use. We don't handle false return there either. I haven't been able to produce false returns either, only ImagickException when testing the method.

tomaszdurka: Have you tried my example? It returns null and issues a warning. Remember some users don't throw on warnings.

alexispeter: As discussed, let's remove the true check since error_reporting is set on framework-wide level.

tomaszdurka:
Looks good.

alexispeter:
tagged https://github.com/cargomedia/CM/releases/tag/1.132.0

PR #1792, opened 2015-06-16
GA MP: Send along "hostname"

fauvel:
Currently we don't send a hostname to *GA Measurement Protocol* when tracking hits. This is problematic because we then aren't able to filter traffic by "hostname" in views.

It looks like we can send it along as `dh`:
https://developers.google.com/analytics/devguides/collection/protocol/v1/parameters#dh

Also note they now have a "debugging" endpoint:
https://developers.google.com/analytics/devguides/collection/protocol/v1/validating-hits

@fauvel maybe something for you?


fauvel:
@njam please review

njam:
lgtm

fauvel:
Tagged 1.122.5

PR #1794, opened 2015-06-16
Add I18n phrase

fvovan:
@njam, @tomaszdurka please review

### DIFF
--- a/library/CM/I18n.php
+++ b/library/CM/I18n.php
@@ -2,15 +2,15 @@
 
 class CM_I18n {
 
-    protected $string, $params;
+    protected $phrase, $variables;
 
     /**
-     * @param string $string
-     * @param array  $params
+     * @param string $phrase
+     * @param array  $variables
      */
-    public function __construct($string, array $params) {
-        $this->string = $string;
-        $this->params = $params;
+    public function __construct($phrase, array $variables) {
+        $this->phrase = $phrase;
+        $this->variables = $variables;
     }
 
     /**
@@ -18,6 +18,6 @@ class CM_I18n {
      * @return string
      */
     public function translate(CM_Frontend_Render $render) {
-        return $render->getTranslation($this->string, $this->params);
+        return $render->getTranslation($this->phrase, $this->variables);
     }
 }
--- a/library/CM/I18n.php
+++ b/library/CM/I18n.php
@@ -1,0 +1,23 @@
+<?php
+
+class CM_I18n {
+
+    protected $string, $params;
+
+    /**
+     * @param string $string
+     * @param array  $params
+     */
+    public function __construct($string, array $params) {
+        $this->string = $string;
+        $this->params = $params;
+    }
+
+    /**
+     * @param CM_Frontend_Render $render
+     * @return string
+     */
+    public function translate(CM_Frontend_Render $render) {
+        return $render->getTranslation($this->string, $this->params);
+    }
+}
--- a/library/CM/I18n/Phrase.php
+++ b/library/CM/I18n/Phrase.php
@@ -2,14 +2,18 @@
 
 class CM_I18n_Phrase {
 
-    protected $phrase, $variables;
+    /** @var $phrase string */
+    protected $phrase;
+
+    /** @var $variables string[] */
+    protected $variables;
 
     /**
      * @param string $phrase
      * @param array  $variables
      */
     public function __construct($phrase, array $variables) {
-        $this->phrase = $phrase;
+        $this->phrase = (string) $phrase;
         $this->variables = $variables;
     }
 
--- a/library/CM/I18n/Phrase.php
+++ b/library/CM/I18n/Phrase.php
@@ -2,19 +2,19 @@
 
 class CM_I18n_Phrase {
 
-    /** @var $phrase string */
-    protected $phrase;
+    /** @var $_phrase string */
+    protected $_phrase;
 
-    /** @var $variables string[] */
-    protected $variables;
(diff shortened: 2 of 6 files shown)


tomaszdurka:
Looks good, couple of things though.

- We should fine a bit better name for this class :)
- You don't need that docu code about instance type in translate smarty helper. Having `if (... instanceof... )` will automatically consider it as instance inside that if.
- Let's rename `$string`, `$params` to `$phrase`, `$variables` as is in several places.

Cheers!

fvovan:
what about 'Localized' ?

njam:
Isn't "I18n" more of a general concept?

In general we're trying to move common files into "module" directories at the root level of "library". So maybe `CM_Translate_Phrase` or `CM_I18n_Phrase`?

fvovan:
I opt for CM_Translate_Phrase 
and what is corresponding module for it?  Or should I create new one with only single class?

njam:
There's no module for this topic yet.
@tomaszdurka what's your opinion?

library/CM/SmartyPlugins/function.translateVariable.php
```diff
@@ -7,5 +7,10 @@ function smarty_function_translateVariable($params, Smarty_Internal_Template $te
     $key = $params['key'];
     unset($params['key']);
 
-    return $render->getTranslation($key, $params);
+    if ($key instanceof CM_I18n) {
```
tomaszdurka: Add check about `$params` size and throw to let the developer know that those params will be totally ignored.

library/CM/I18n.php
```diff
@@ -0,0 +1,23 @@
+<?php
+
+class CM_I18n {
```
tomaszdurka: CM_I18n is great name for module containing all the language/translation related classes.
It's not the best moment to move all those classes now.

Let's only rename this class to CM_I18n_Phrase and make it first one being properly placed in the new module.

tomaszdurka:
@fvovan Review is done, thanks!

library/CM/I18n/Phrase.php
```diff
@@ -0,0 +1,23 @@
+<?php
+
+class CM_I18n_Phrase {
+
+    protected $phrase, $variables;
```
tomaszdurka: Add docu.

tests/library/CM/SmartyPlugins/function.translateVariableTest.php
```diff
@@ -0,0 +1,27 @@
...
+
+require_once CM_Util::getModulePath('CM') . 'library/CM/SmartyPlugins/prefilter.translate.php';
+
+class smarty_function_translateVariableTest extends CMTest_TestCase {
+
+    /**
+     * @expectedException        InvalidArgumentException
+     * @expectedExceptionMessage Passed params will be ignored as you provided CM_I18n_Phrase object
+     */
+    public function testTranslateObject() {
```
tomaszdurka: rename to testTranslatePhrase

library/CM/I18n/Phrase.php
```diff
@@ -0,0 +1,23 @@
...
+
+    /**
+     * @param string $phrase
+     * @param array  $variables
+     */
+    public function __construct($phrase, array $variables) {
+        $this->phrase = $phrase;
```
tomaszdurka: Cast to string.

library/CM/I18n/Phrase.php
tomaszdurka: Probably use `string[]` for variables.

tests/library/CM/SmartyPlugins/function.translateVariableTest.php
```diff
@@ -0,0 +1,27 @@
...
+            $this->assertSame(['bar' => 3], $params);
+        });
+        /** @var CM_Frontend_Render $render */
+
+        $object = new CM_I18n_Phrase('Bar value is {$bar}', ['bar' => 3]);
+        $render->parseTemplateContent('{translateVariable key=$foo}', ['foo' => $object]);
+        $this->assertSame(1, $getTranslationMethod->getCallCount());
+
+        $render->parseTemplateContent('{translateVariable key=$foo more=one}', ['foo' => $object]);
+        $this->assertSame(0, $getTranslationMethod->getCallCount());
```
tomaszdurka: Adjust to use `$this->catchException()` as discussed.

fvovan:
@njam, please doublecheck

njam:
- CM_I18n_Phrase: Non-public members should be prefixed with `_`
- CM_I18n_Phrase::__construct: Docu for $variables could be `string[]`
- CM_I18n_Phrase: Add simple unit test
- smarty_function_translateVariable: How about also creating an `CM_I18n_Phrase` instance if we don't pass one? Might be simpler in the end?

fvovan:
I added tests for CM_I18n, but now one assertion is duplicated with function.translateVariableTest.php
On the one hand in smarty tests we should check only correct CM_I18n_Phrase state during translate invocation.
On the other hand we can't do it because its' properties are protected.
@njam what do you think?

njam:
- CM_I18n_PhraseTest: formatting

Rest looks good. I think having duplication in tests is okay.

tomaszdurka:
Looks good, merging.

tomaszdurka:
@fvovan To use in in external, dependant projects you need to tag it. Could you do it? If you have any problems please let me know.

fvovan:
@tomaszdurka,  it's done
1.126.0

PR #1795, opened 2015-06-17
Click Feedback

christopheschwyzer:
- Buttons
- Menu
- Other Interactions, Links?

See Google approach: https://www.google.com/design/spec/animation/responsive-interaction.html#responsive-interaction-surface-reaction

To be seen e.g.  [Google Music](https://play.google.com/music/listen?u=0#/now), new Inbox

Also http://material-ui.com/#/components/buttons


christopheschwyzer:
Open issues: 
- Click decorators block each other?!
- Works only for buttons for now

@tomaszdurka @vogdb @njam One of you could help me out to pair-finish this?

vogdb:
Me ready!

christopheschwyzer:
Still not sure how to handle cases outside menu & buttons...

tomaszdurka:
You mean you are not sure how to intercept them or how to make them visually attractive? :)

vogdb:
Ok. The problem with `clickFeedback` was caused by the behaviour of jQuery `$.event.special.click.handle`. It is called for every event handler that is bound to the element. Here is the demonstration http://jsbin.com/yekutatiwe/edit?html,js,output. The current implementation of click decorators does not take in account such situation. Hence the problem.

njam:
Tomasz wrote an alternative implementation for "click feedback" which is independent of "click decorators". Just binding to `mousedown`. I think that's how it should be.
@tomaszdurka where is that code?

tomaszdurka:
Is this still valid? As we're using mousedown on document now?

njam:
Ah it's in the PR, sorry :)

tomaszdurka:
@njam https://github.com/cargomedia/CM/pull/1795/files#diff-51afe0ff8489e70844437f7f10985a82

vogdb:
We will see. I haven't reviewed it yet.

client-vendor/after-body/jquery.clickDecorators/jquery.clickDecorators.js
```diff
@@ -9,8 +10,8 @@
 		var before = [], after = [], i;
 
 		_.each($.clickDecorators, function(decorator, name) {
-			if ($this.data('click-' + name)) {
-				if (decorator.before) {
+      if (decorator.isApplicable($this)) {
+        if (decorator.before) {
```
tomaszdurka: Inconsistent styling - probably you need to use library's one. Right?

njam: This is our code

vogdb:
The new approach does not seem to work on Safari desktop and iOS.

vogdb:
Oh, sorry. iOS only it seems.

christopheschwyzer:
It's _fastClick_ :unamused:

christopheschwyzer:
Possible solutions: 
- drop fastClick (Chrome removes the delay already, plus, with the new feedback the delay doesn't feel bad anymore)
- listen for _click_ event instead
@vogdb @tomaszdurka  @njam other ideas?

njam:
Hm looks like we removed the "mouseup"/"mousedown" synthetization code we patched FastClick with? Probably because "select2" worked without it..
Original patch: https://github.com/ftlabs/fastclick/pull/308/files
File history: https://github.com/cargomedia/CM/commits/master/client-vendor/after-body/fastclick/fastclick.js

But I think listening on `click` should do the trick. In case of "fastclick" it will happen at the same time as `mousedown` anyway.

christopheschwyzer:
@vogdb please review

njam:
Would still be good to bind on *both* events so the effect is faster on non-iOS browsers. Can we bind both?

vogdb:
@christopheschwyzer should I postpone until you try both events ?

njam:
We will probably need to somehow memorize that an event was already processed, to avoid triggering the effect twice if both events are fired.
Can we set a property on the `Event` instance?

vogdb:
We can. Also what if we fix the previous solution with decorators? The one we refused in favor of separate plugin ? Would it be better ?

christopheschwyzer:
Not sure I can provide to a solution, but I'd be happy to pair on it...

vogdb:
I'm overtaking this issue for a bit to fix some crazy event hush push.

vogdb:
> Hm looks like we removed the "mouseup"/"mousedown" synthetization code we patched FastClick with? Probably because "select2" worked without it..

Yes. See #1558.

> Can we set a property on the Event instance?

We can not set it on Event instance but we can set a temp private variable on event's target. Such trick should do the job.

njam:
Investigates this with Aleksei. There are two problems:
- **Click decorator** only works with 1 event handler. If there are more handlers the result is not as expected. In case of "confirmed" we will only use the first handler. In case of "spinner" we will probably add multiple spinners. This use case should be fairly uncommon in real life. We will ignore it for now.
- **Click feedback** adds a click handler that registers itself as a next handler of **click decorator**. So we end up with above problem. To work around it we will try *not* using jQuery here, but directly use `addEventListener()`. Low level, old school.

vogdb:
@christopheschwyzer please give me the rights to your CM. Can't push the changes =(

vogdb:
btw @christopheschwyzer your changes are lgtm.

christopheschwyzer:
@vogdb can you try again?

vogdb:
Did it. @tomaszdurka or @njam please review my changes (last 2 commits).

client-vendor/after-body/jquery.clickDecorators/decorators/20-confirmed.js
```diff
@@ -3,45 +3,68 @@
...
-			if ($this.hasClass('confirmClick')) {
-				$this.data('clickConfirmed.deactivate')();
-			} else {
-				activateButton();
-				event.preventDefault();
-				event.stopImmediatePropagation();
-			}
-		}
-	};
+  var ConfirmationMonitor = function() {
```
tomaszdurka: @vogdb wdyt about putting all these methods into decorator object? Possibly as _private members.

I think if we don't store any state on this class, it's more confusing than helping to introduce this class.

tomaszdurka: Add docu for methods.

client-vendor/after-body/jquery.clickFeedback/jquery.clickFeedback.js
```diff
@@ -0,0 +1,46 @@
+/*
+ * Author: CM
+ * Dependencies: jquery.transit.js
+ */
+(function($) {
+
+  function isApplicable(elem) {
+    do {
+      if (elem.classList.contains('clickFeedback')) {
```
tomaszdurka: So we cannot use jquery closest here?

client-vendor/after-body/jquery.clickFeedback/jquery.clickFeedback.js
```diff
@@ -0,0 +1,46 @@
...
+        return true;
+      }
+      elem = elem.parentNode;
+    } while (elem.parentNode);
+    return false;
+  }
+
+  document.addEventListener('click', function(event) {
```
tomaszdurka: Didn't we use mousedown also for shorter response time?

tomaszdurka:
Review done.

Apart from inline comments:
- There are a lot of formatting changes.
- I didn't carefully look into css changes.
- Personally I don't like that we add css classes and have fixed isApplicable behaviour at the same time. We don't make click-feedback optional. This is probably the case for a lot more vendor code we use anyway. Just thought this could/should be optional for cm-framework users. Especially that we define effect here in CM framework too.

client-vendor/after-body/jquery.clickDecorators/decorators/20-confirmed.js
vogdb: The problem is to access the decorator object. The methods of this object are called with a DOM element as a context, so there is no link to the decorator object inside of its methods `before` and `after`. It means that we must have a variable/link outside of the decorator object. That decision led me to `ConfirmationMonitor`. I can make it an object instead of a class. It does not play that much of a role. Should I change it to an object?

client-vendor/after-body/jquery.clickFeedback/jquery.clickFeedback.js
vogdb: It would require from us to synthesise `mousedown` once again in fastclick cause iOS does not produce such event. The synth of `mousedown` produces its own problems. Currently tests show that delay is acceptable.

vogdb:
@tomaszdurka please review.

njam:
- 20-confirmed.js: How about storing a full "class" object in the element's data attribute (like `ConfirmationMonitor`)? And then move `isActivated` and `deactivate` to it?

vogdb:
@njam please look.

njam:
- 20-confirmed.js: constructor: docu

Rest look good!

vogdb:
@njam please look

njam:
Looks good - I only checked the decorator changes!

vogdb:
The rest was reviewed by me. It's lgtm.

PR #1796, opened 2015-06-17
Remove code styling in CM_Frontend_JavascriptContainer::compile()

dlondero:
Unfortunately this `CM_Frontend_JavascriptContainer::compile()` is not styling only pure js code but also free text which could contain chars matching the pattern of the `preg_replace`. We agreed on removing it.


dlondero:
@tomaszdurka please review.

In `CM_Frontend_JavascriptContainerTest::testCompile()` should be ok the remove one case of the assertion as it is not expected anymore. Still I think `CM_Http_Response_View_AbstractTest:: testReloadComponent()` and `CM_Http_Response_View_AbstractTest::testReloadComponentAdditionalParams()` should not be deleted because they're testing something else but honestly I have no clue from where this `;;` is coming after `registerChild`.

tomaszdurka:
This double `;;` is coming from couple of things together.
We have so called Javascript Container which you can append lines of JS into. Those lines are stored in array and then compile implodes all of them using `;\n` combination.
We try to push all of our lines with semicolon in the end, but possibly sometimes we could miss it.

In the end there is code like this:
```php
implode(";\n", array("foo();", "bar();");
```
resulting in: `foo();;\nbar();`

------------------------------------------------------------

Probably it would be better to remove those from the end if we are adding them anyway.
I think we should remove ";" from implode param and check if it's fine everywhere.
Problem is that on dev-box with all the code lines not-minified we will have no problems with code like:
```js
foo()
bar()
```
but when it will be minified it would could have problems
```js
foo()bar()
```

I would still go for it and remove the `;` from implode. Wdyt?

tomaszdurka:
Ok, I've actually tested how uglify treats all this non-semicolon ending lines and it's all fine.

Let's remove `;` from the implode then.

dlondero:
@tomaszdurka please review

tomaszdurka:
Looks good.

PR #1799, opened 2015-06-18
Keep array index order in provision script loader

tomaszdurka:
In `CM_Provision_Loader::_getScriptList()` we should keep the original array's order for items with the same run-level.

We could implement a custom callback for ordering that considers the index from the original array.

### DIFF
--- a/library/CM/Provision/Loader.php
+++ b/library/CM/Provision/Loader.php
@@ -67,7 +67,7 @@ class CM_Provision_Loader {
     protected function _getScriptList() {
         $scriptList = $this->_scriptList;
         $runLevelList = \Functional\invoke($scriptList, 'getRunLevel');
-        array_multisort($runLevelList, $scriptList);
+        array_multisort($runLevelList, array_keys($scriptList), $scriptList);
         return $scriptList;
     }
 }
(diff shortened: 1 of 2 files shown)


tomaszdurka:
I couldn't really reproduce this problem. Scripts were sorted correctly without the adding index-based sorting. 
Any ideas how to test this, how this could happen? I trust CM_Config retrieval of the script classes is fine.

@alexispeter please review.

alexispeter:
could reproduce it with
```php
$scriptList[] = 'SK_ExampleData_Locations';
$scriptList[] = 'SK_ExampleData_Languages';
$scriptList[] = 'SK_ExampleData_PaymentProviders';

$runLevels = [5, 5, 5];
array_multisort($runLevels, $scriptList);

var_dump($runLevels);
var_dump($scriptList);
```
It sorts it alphabetically using the scriptClass if the runlevel is the same.

alexispeter:
If you give the script-classes custom names in the test, you should be able to do a regression-test that doesn't work without the changes.

dlondero:
@tomaszdurka I tested this with the same scenario as yesterday and it's working as expected! Thanks :)

tomaszdurka:
As discussed with @alexispeter multisort should work well. It's still not possible for me to reproduce it using $instances of provision script as is in getScriptList, but we let it go and merge as is.

PR #1798, opened 2015-06-18
Fix CM_Component_Example::loadExample for subclasses

vogdb:
Client-side `CM_Example_Component::loadExample` should load current class example instead of hardcoded `CM_Component_Example`.

Use `this._class` instead.


vogdb:
@tomaszdurka please review.

tomaszdurka:
Looks good.

PR #1802, opened 2015-06-18
Breaking: Add tracking service for adagnit.io

fauvel:


fauvel:
@njam please review

njam:
lgtm!
Did you test it in a browser, that the JS works well?

njam:
Maybe `trackLogin()` and `trackSignUp()` can be called from `trackAction()` by looking at the action's attributes?

njam:
Shouldn't we track `Site::getHost()` to be sure it's unique (some sites have the same name)?

Rest looks good!

njam:
lgtm

fauvel:
@njam please review (tests must be adjusted)

fauvel:
@njam please review

fauvel:
@njam please review

fauvel:
@njam I've fixed a bug with virtual paths due to the fact that we reordered tracking and page rendering: Pages cannot overwrite the page path any more by calling `trackPageView()` explicitly in `prepareResponse()`, since this would be overwritten again in `_processPage()` now. I've opted for a getter to declare virtual paths on the page class.

fauvel:
@njam please review

njam:
lgtm

njam:
@fauvel moving all this logic out to a different class had some downsides. I now opted to just introduce a class to hold the result of processing (`ProcessingResult`).
Please have a look! What do you think?

fauvel:
Perfect, it's a very nice trade-off between encapsulating the rendering logic and adding some more code complexity... I like it! The code looks good, too.

PR #1803, opened 2015-06-18
Add special handler for CancellationError in window.onerror.

vogdb:


vogdb:
@njam please review.

vogdb:
I couldn't reproduce the problem on the dev version and couldn't logically find where it comes from. That is why I've decided to extend the existing handler of unexpected errors.

njam:
@tomaszdurka can you review this?

client-vendor/after-body/00-jserror/jserror.js
```diff
@@ -26,7 +26,10 @@
 			if (window.onerror) {
 				this.onerrorBackup = window.onerror;
 			}
-			window.onerror = function(message, fileUrl, fileLine) {
+			window.onerror = function(message, fileUrl, fileLine, fileColumn, error) {
+        if (error instanceof Promise.CancellationError) {
+          message = '###' + message + ' :' + fileColumn + ' ' + error.stack
+        }
```
tomaszdurka: Formatting.
Why do you format this message in this special way with hashes in front? 
Also error.stack is multiline right? The it's pretty strange first line is with message/filecolumn and other below.
Do you have example message for this?

vogdb:
@tomaszdurka please review.

vogdb:
@tomaszdurka please review.

tomaszdurka:
Well, if we have no other choice....

tomaszdurka:
Looks good!

PR #1805, opened 2015-06-20
Don't load JS files individually in debug mode

njam:
@tomaszdurka @vogdb

I would suggest to remove the individual loading of all library files in debug mode.
Doing so has some disadvantages like that the browser loads *a lot* of files one by one, which takes especially long on mobile devices.
The advantage of loading them individually is only that the line numbers match the ones in the original file. I don't think that is very important.
What do you think?

Related:
https://github.com/cargomedia/puppet-packages/blob/master/modules/cm/manifests/vhost.pp#L74

### DIFF
--- a/library/CM/Http/Response/Resource/Javascript/Library.php
+++ b/library/CM/Http/Response/Resource/Javascript/Library.php
@@ -4,13 +4,7 @@ class CM_Http_Response_Resource_Javascript_Library extends CM_Http_Response_Reso
 
     protected function _process() {
         if ($this->getRequest()->getPath() === '/library.js') {
-            $query = $this->getRequest()->getQuery();
-            $skipLibraries = !empty($query['debug']);
-            if ($skipLibraries) {
-                $this->_setAsset(new CM_Asset_Javascript_Internal($this->getSite()));
-            } else {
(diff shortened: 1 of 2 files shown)


tomaszdurka:
The idea behind it was to use ide breakpoints, debug code in there. Never fully worked. 

What about keeping this functionality and simply produce same as on production? Otherwise we can skip it.

njam:
> What about keeping this functionality and simply produce same as on production? 

You mean to keep the code, but disable it?

tomaszdurka:
Yes.

njam:
I would remove it then - it adds some complexity to the puppet modules, which will be tricky to understand in the future if we don't use it anymore?

tomaszdurka:
Let it be. 

Sidenote: Btw I think I have issue connected with all those vendor js/css somewhere opened.

njam:
You mean this one about assets? https://github.com/cargomedia/CM/pull/1557

njam:
@vogdb please let us know what you think - if it can/should be removed.

vogdb:
Sync between lines is not important. Almost never used it. Plenty of other ways to sync them for me. Lets do it.

njam:
...the proper way to do this would probably be to create a *source map*.

PR #1806, opened 2015-06-21
Use uncommon value separator for select2

njam:
Otherwise if any value contains the separator (",") it will be split by that value.
Via https://github.com/denkmal/denkmal.org/issues/473

@vogdb please review


njam:
@vogdb please review

vogdb:
lgtm. Btw I've restarted the build cause it was failed.

PR #1808, opened 2015-06-21
Remove message about "IE compatibility mode"

njam:
@christopheschwyzer I'd suggest to remove this. The "IE 9" was outdated anyway, and I think IE doesn't trigger "compatibility mode" anymore?


christopheschwyzer:
Agree, lgtm

PR #1755, opened 2015-06-23
Ajax/Promise "throttler" to avoid multiple requests

vogdb:
We often times want to avoid multiple concurrent ajax requests of the same time.

Usually we do:
```js
var requestList = {};

/**
 * @returns Promise
 */
function doSomething()  {
 if (requestList['something']) {
  return requestList['something'];
 }
 return requestList['something'] = ajax();
}
```

It would be nice to wrap this into a small script, so we don't need to declare `requestList` every time, plus have a nicer interface. Preferably it would would work as kind of like a "Promise throttling":
```js
/**
 * @returns Promise
 */
function doSomething() {
 return promiseThrottler('something', function() {
  return ajax(); // returns Promise
 });
}
```

@vogdb @tomaszdurka thoughts?
Related to https://github.com/cargomedia/CM/issues/1754

### DIFF
--- a/client-vendor/after-body/promise-throttler/promise-throttler.js
+++ b/client-vendor/after-body/promise-throttler/promise-throttler.js
@@ -7,17 +7,27 @@
   var storage = {};
 
   /**
-   * @param {String} name
-   * @param {Promise} promise
-   * @returns {Promise}
+   * @callback PromiseThrottled
+   * @param {...Object|String|Number} any number of optional params
+   * @return {Promise}
    */
-  function promiseThrottler(name, promise) {
-    if (!storage[name]) {
-      storage[name] = promise.finally(function() {
-        delete storage[name];
-      });
-    }
-    return storage[name];
+
+  /**
+   * @param {PromiseThrottled} fn
+   * @param {Boolean} cancel
+   * @returns {PromiseThrottled}
+   */
+  function promiseThrottler(fn, cancel) {
+    var promise;
+    return function() {
+      if (cancel && promise && promise.isPending()) {
+        promise.cancel();
+      }
+      if (!promise || !promise.isPending()) {
+        promise = fn.apply(null, arguments);
+      }
+      return promise;
+    };
   }
 
   global.promiseThrottler = promiseThrottler;
--- a/client-vendor/after-body/promise-throttler/promise-throttler.js
+++ b/client-vendor/after-body/promise-throttler/promise-throttler.js
@@ -1,0 +1,20 @@
+/*
+ * Author: CM
+ */
+
+(function(global) {
+
+  var storage = {};
+
+  function promiseThrottler(name, promise) {
+    if (!storage[name]) {
+      storage[name] = promise.finally(function() {
+        delete storage[name];
+      });
+    }
+    return storage[name];
+  }
+
+  global.promiseThrottler = promiseThrottler;
+
+})(window);
--- a/client-vendor/after-body/promise-throttler/promise-throttler.js
+++ b/client-vendor/after-body/promise-throttler/promise-throttler.js
@@ -19,10 +19,11 @@
   function promiseThrottler(fn, options) {
     options = _.defaults(options, {cancel: false});
     var promise;
-    
+
     return function() {
-      if (options.cancel && promise && promise.isPending()) {
+      if (options.cancel && promise && promise.isPending() && promise.isCancellable()) {
         promise.cancel();
+        promise = null;
       }
       if (!promise || !promise.isPending()) {
         promise = fn.apply(null, arguments);
--- a/client-vendor/after-body/promise-throttler/promise-throttler.js
+++ b/client-vendor/after-body/promise-throttler/promise-throttler.js
@@ -4,8 +4,6 @@
 
 (function(global) {
 
-  var storage = {};
-
   /**
    * @callback PromiseThrottled
    * @param {...Object|String|Number} any number of optional params
@@ -14,13 +12,16 @@
 
   /**
    * @param {PromiseThrottled} fn
-   * @param {Boolean} cancel
+   * @param {Object|null} options
+   * @param {Boolean} options.cancel Whether to cancel the previous promise if it is still running.
    * @returns {PromiseThrottled}
    */
-  function promiseThrottler(fn, cancel) {
+  function promiseThrottler(fn, options) {
+    options = _.defaults(options, {cancel: false});
     var promise;
+    
     return function() {
-      if (cancel && promise && promise.isPending()) {
+      if (options.cancel && promise && promise.isPending()) {
         promise.cancel();
       }
       if (!promise || !promise.isPending()) {
--- a/client-vendor/after-body/promise-throttler/promise-throttler.js
+++ b/client-vendor/after-body/promise-throttler/promise-throttler.js
@@ -6,6 +6,11 @@
 
   var storage = {};
 
+  /**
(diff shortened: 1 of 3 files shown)


vogdb:
@njam @tomaszdurka please look.

njam:
Simple enough! Add some docu to the function?

vogdb:
Yes!

tomaszdurka:
Looks good, what about making it not self-executed. Then store instance of it somehwere in cm app instance? It's probably the same in the end.

njam:
Use amd.js? :P

vogdb:
If you store it in cm app then you will not be able to access it from a global scope. Is it ok or should it be available from both global and cm app?

tomaszdurka:
I am generally against global state which we have plenty with jquery plugins etc. I would rather make library witout self-execution code and create instance of it somwhere, it can be global too...

njam:
But then it's a CM plugin, right? Or how would you access it without having a reference on the global context to it?

vogdb:
So, what is the resolution of this discussion ?

tomaszdurka:
Just against global state. I thought that `PromiseThrottler` would be a class defined within same file you've created. It would create this class in global context. We could then create instance of this class in global or cm context.

Maybe it's overkill as cm app is a global singleton anyway.

tomaszdurka:
If I did not convince any of you guys, let stay with current solution.

njam:
Hm, so you would suggest?:
```js
var ThrottlerClass = window.ThrottlerClass;
var throttler = new ThrottlerClass();
throttler.throttleFunction(promise);
```

I don't see the reason to have this class for this purpose, isn't this mainly a utility function? I would rather see:
```js
var throttler = window.throttler;
throttler.throttleFunction(promise);
```
or even (current):
```js
var throttleFunction = window.throttleFunction;
throttleFunction(promise);
```

wdyt?

vogdb:
I like the latter one!

tomaszdurka:
Ok I agree this will be the easier, kinda cleaner solution and that with not-auto-install implementation I am for reasons which we never gonna use. I just don't like that we have single global instance of something with state. Most probably there will be no use case for using multiple of these. Then I am ok with your implementation.

I guess somehting which would explain my point more is:
```js

// vendor file, doesn't influence global state
var extendWithThrottle = function(global) {
  global.throttle = function(name, promise) {
  }
};

// cm (ready?)
extendWithThrottle(window);
```

But as said, maybe it's my bad approach to those javascript vendors.

njam:
True!

Maybe the function should store state only in itself? So like the underscore functions - return a new function that wraps the old one and is throttled?

vogdb:
How can we store state only in itself? Consider this example http://jsbin.com/revajiribi/edit?html,js,output.

njam:
You would need to use always the same returned function from 1 function call. Agree it's a bit inconvenient.. It works good if you use it like this:
```js
$('.foo').on('click', _.throttle(something());
```
It also has the advantage that you don't really need to care about the "key"/"id".

Would this pattern fit our use cases?

vogdb:
I'm not sure what exactly are the use cases that we are trying to cover here.

njam:
Anywhere where we're running AJAX requests, but want max. 1 request to be in-flight at a time. Usually we store the *current request* on the view instance.

But I can see now that in most of these use cases we usually `abort()` the previous request. I think it would be good to support both use cases:
- A: Cancel previous promise and immediately start new one
- B: Ignore new promise if old one is still "pending"

How could we support both use cases?

vogdb:
Ok. I see. Both use cases can be supported by an additional param to the function or the second variation of promiseThrottler that cancels the previous request.

vogdb:
Ok. The variant without cancellation http://jsfiddle.net/4hpzcpgz/. The variant with cancellation http://jsfiddle.net/ne41wyxb/1/. @njam wdyt ?

njam:
I think it looks good.
Did you check our usages and how would could apply such a pattern there? (e.g. search for regexp `_.*request:` in files matching `*.js`)

vogdb:
For some cases it can be difficult to apply. An example:
- CM_FormField_Suggest::ready
```js
      query: function(options) {
        if (this._request) {
          this._request.cancel();
        }
        this._request = field.ajax('getSuggestions', {'term': options.term, 'options': field.getOptions()})
          .then(function(results) {
            options.callback({
              results: results
            });
          });
      },
```
  You can see that every new promise is different from another because it is applied to a different `options.term`. So, for example I don't see an option to use a throttled promise of a term `Aus` for a term `Austra`. It is the situation where the existence of some constant key for those type of promises will solve the problem.

The above example is not the one. We have 4 of such places where we deal with requests to the same url but different arguments. As the result we have this problem. It can be solved but the solution is very very vulnerable:

Save the throttled promise outside of the callback listener and use the argument through a shared variable.
```js
var savedThrottledPromise = promiseThrottler(function(){
  return promise(shared.arg);
});
```

inside callback set the argument and call the throttled promise. Crazy!
```js
shared.arg = 5;
return savedThrottledPromise;
```

My suggestion is to have at least the second version of `promiseThrottler` where you can specify the corresponding key for the promise. The implementation that I've pushed here.

njam:
Can't we just pass the arguments through the throttler (using `.apply()`)?
```js
    var query = promiseThrottler(function(options) {
      return field.ajax('getSuggestions', {'term': options.term, 'options': field.getOptions()})
        .then(function(results) {
          options.callback({
            results: results
          });
        });
    });
```

Should be useable like this:
```
query({term: "foo", options: {}});
```

Then pass it to the plugin:
```js
    this._$input.select2({
      [...]
      query: query,
```

vogdb:
You are absolutely right. As always I couldn't see another option with `apply`. @njam @tomaszdurka how do you like the new version? I've pushed the updated CM_FormField_Suggest to show you the example.

njam:
- promise-throttler.js: Remove `storage`
- promise-throttler.js: How about changing boolean `cancel` to object `options`. Possibly we can add things like `leading: false` and `trailing: true` (from underscore) - http://underscorejs.org/#throttle
- Suggest.js: Remove `_request` declaration

Rest looks very good

njam:
How about defining those functions inline instead of in `ready()`?
E.g.:
```js
CM_View_Abstract.extend({
 _pageRequest: promiseThrottler(...)
});
```

vogdb:
It can be done but would require to pass a link to the context as an argument to the function that we throttle.
For example now we have
```js
  var layout = this;
  this._pageRequest = promiseThrottler(function(path) {
    //...here we use layout
```
we can replace it with:
```js
  this._pageRequest = promiseThrottler(function(layout, path) {
    //...here we use layout
```

is it ok ?

njam:
I see! Let's not do it then.
How about moving it to `initialize()` instead of `ready()`? Seems the better place - it's available directly after instantiation and not only once the "dom" is ready.

vogdb:
ok

vogdb:
Should all of the usages be in `initialize` or just `CM_Layout_Abstract` ?

njam:
I think all usages when we set up "code" for the object. In `ready` I would only put things related to the DOM (events, DOM etc).

njam:
- Layout/Abstract.js: _pageRequest: Rename to `_loadPageThrottled?
- Layout/Abstract.js: _pageRequest: Adjust docu

njam:
lgtm

tomaszdurka:
There is follow up in #1838

PR #1814, opened 2015-06-24
Allow passing objects as parameters to cm.getUrl()

njam:
This will allow to pass params like `{foo: {bar: 12}}` as parameters and result in `?foo[bar]=12`.
I need this for some URLs in sk. 

No existing usages are affected.

@vogdb please review
@christopheschwyzer fyi

### DIFF
--- a/library/CM/App.js
+++ b/library/CM/App.js
@@ -126,7 +126,7 @@ var CM_App = CM_Class_Abstract.extend({
     params = params || null;
     relative = relative || false;
     if (params) {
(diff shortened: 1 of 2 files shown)


vogdb:
The changes are lgtm but shouldn't we have the same logic on back-end? I mean when url is generated on back-end side.

njam:
@vogdb good point!
It's already working - I added a test - please review!

vogdb:
lgtm!

PR #1812, opened 2015-06-25
Layout::loadPage: Error catching too broad?

vogdb:
The ajax request in `loadPage()` previously would handle *ajax* error in an error callback:
```js
    this._pageRequest = this.ajaxModal('loadPage', {path: path}, {
      success: function(response) {
[...]
      },
      error: function(msg, type, isPublic) {
[...]
      }
    });
```
If anything went wrong in the `success` handler it would throw an unhandled error and would eventually be logged back to our server as a "JS error".

With the new Promise-based implementation the code looks like this:
```js
    this._pageRequest = this.ajaxModal('loadPage', {path: path})
      .then(function(response) {
[...]
      })
      .catch(function(error) {
[...]
      })
```
Which means that an error in `then` will be catched in `catch` and *not* be logged to the server.

I guess it's good to set the layout into an "error" state in case of *any* errors. Should we additionally throw again to log it to the server?
@vogdb @tomaszdurka wdyt?
cc @fauvel

### DIFF
--- a/library/CM/App.js
+++ b/library/CM/App.js
@@ -281,10 +281,13 @@ var CM_App = CM_Class_Abstract.extend({
     },
     /**
      * @param {Object} content
+     * @param {String} content.msg
+     * @param {String} content.type
+     * @param {Boolean} content.isPublic
      * @return {Error}
      */
     create: function(content) {
-      var error = new Error();
+      var error = new Error(content.msg);
       return _.extend(error, content);
     }
   },
@@ -678,7 +681,7 @@ var CM_App = CM_Class_Abstract.extend({
         e.preventDefault();
         var error = e.originalEvent.detail.reason;
         if (!(error instanceof Promise.CancellationError)) {
-          cm.error.trigger(error.msg, error.type, error.isPublic);
+          throw error;
         }
       });
       $(window).focus(function() {
--- a/library/CM/Layout/Abstract.js
+++ b/library/CM/Layout/Abstract.js
@@ -79,6 +79,7 @@ var CM_Layout_Abstract = CM_View_Abstract.extend({
         if (!(error instanceof Promise.CancellationError)) {
           layout._$pagePlaceholder.addClass('error').html('<pre>' + error.msg + '</pre>');
           layout._onPageError();
+          throw error;
         }
       }).finally(function() {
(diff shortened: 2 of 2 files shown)


vogdb:
@njam please look.

njam:
Also no warning icon somewhere in the status bar?

vogdb:
Do you mean to return back this code?
```js
  cm.error.trigger(error.msg, error.type, error.isPublic);
```

njam:
I meant whether some browsers will maybe show some warning icon in their status bar on uncaught errors? IE did that in the past.

We should be 100% sure that this doesn't affect end users. Did you google about this in general? What's "good practice"?

vogdb:
It's a good practice to catch errors with `window.onerror`.
https://danlimerick.wordpress.com/2014/01/18/how-to-catch-javascript-errors-with-window-onerror-even-on-chrome-and-firefox/
http://stackoverflow.com/questions/5913978/cryptic-script-error-reported-in-javascript-in-chrome-and-firefox
Considering the fact that we already took care of CDN scripts. I see that all of them have `crossorigin="anonymous"`. So everything is good now. It does not lick any errors outside. All of them are caught.

Also I've retested throwing error on the devices. No warning icons or statuses. Everything is smooth and hidden from user.

njam:
In `jserror.js` we have the parameter `suppressErrors` which makes the error handler to return `true`. But we disable this behaviour.
What effect does it have? Should we change it to `true` then?

vogdb:
> What effect does it have?

When function `window.onerror` returns `true` then it means that the default browser handler for unexpected errors won't handle this error. On practice in 99% cases it means that the error won't appear in the console which is ok cause when we need, we put it in the console on our own by `jserror.log(message, fileUrl, fileLine);`.

> Should we change it to true then?

I'm not sure that it is needed because it's still good to have a default browser handler to handle errors. As my testing showed, the supported browsers default behaviour is to print the error in the console, nothing more, so it's hidden from user. If we want to set it to `true` then we must do it only on production. Dev version should be as verbose as possible.

njam:
Agree, this change set is done then? lgtm.

vogdb:
> Agree, this change set is done then?

Yes

PR #1817, opened 2015-06-25
Indication for lazy loading images

christopheschwyzer:
Currently the space used for lazily loaded images is just white. On white background one doesn't see that something should load in that place.

I would suggest to make those boxes background light grey.

@christopheschwyzer @vogdb wdyt? can we do it with CSS?

### DIFF
--- a/layout/default/css/contentPlaceholder.less
+++ b/layout/default/css/contentPlaceholder.less
@@ -15,6 +15,7 @@
     left: 0;
     width: 100%;
     height: 100%;
+    background-color: @colorBgBody;
   }
 
   &.stretch {


christopheschwyzer:
@njam please review
@vogdb fyi

njam:
lgtm

vogdb:
@christopheschwyzer @njam Why did you decide to put background color on the actual images and not stubs? If image is transient then it has grey background.
In other words why did you choose `.contentPlaceholder-size + *` and not `.contentPlaceholder-size` ?

njam:
Background color should only be visible *until* the image is loaded. Can you do a PR to change that behaviour if it's not correct now?

vogdb:
ok

vogdb:
Hmm. Here comes the problem. Images are not the only content of `contentPlaceholder`. There can be also complicated div for video preview and potentially any content but basically only images can be affected by this grey background I guess. Not so easy as it seemed in the beginning.

christopheschwyzer:
@vogdb Is the current solution making problems? Otherwise I don't see a reason to act here.

vogdb:
@christopheschwyzer yes yes but I've already overreacted! :goberserk:

PR #1819, opened 2015-06-25
Add a TTL to CM_Queue

fauvel:


fauvel:
@njam please review

njam:
> it's not so clear on first sight what the TTL applies to
the whole queue or the individual entry
maybe add a note to CM_Queue::setTtl?

fauvel:
It's the whole queue (key)

njam:
Yes, just meant that for other developer this might not be clear from the signature/docu. Should we add an explanation?

fauvel:
@njam please review

PR #1820, opened 2015-06-25
Add TTL to the Adagnit tracking queue

fauvel:
Based on #1819

### DIFF
--- a/library/CMService/Adagnit/Client.php
+++ b/library/CMService/Adagnit/Client.php
@@ -148,13 +148,6 @@ EOD;
     protected function _pushEvent(CM_Model_User $user, $eventType, array $data = null) {
         $trackingQueue = $this->_getTrackingQueue($user);
         $trackingQueue->push(['eventType' => $eventType, 'data' => $data]);
-        $this->_setTtl($trackingQueue);
-    }
-
-    /**
-     * @param CM_Queue $trackingQueue
-     */
-    protected function _setTtl(CM_Queue $trackingQueue) {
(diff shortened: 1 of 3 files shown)


fauvel:
@njam please review

njam:
- CMService/Adagnit/Client.php: Constructor: docu
- CMService/Adagnit/Client.php: Add a function to add items to queue? With parameter $eventType and $data
- CMService/Adagnit/Client.php: Now we set the TTL when *getting* the queue and again when adding items to the queue. Is that needed?
- CMService/Adagnit/Client.php: IMO we can set a good default as TTL (24h?) and not make this configurable? I don't think it makes sense to change this ever?

fauvel:
> Now we set the TTL when getting the queue and again when adding items to the queue. Is that needed?

No, I've just forgotten to remove the code from the queue getter.

> IMO we can set a good default as TTL (24h?) and not make this configurable? I don't think it makes sense to change this ever?

As for all other caching instances, it definitely makes sense to make the ttl configurable imo. You would choose different values for different apps running with a different load and on a different infrastructure – the framework must adapt to the available resources, not the other way round.

njam:
- CM/Redis/Client.php: Remove docu comment about expiration (this function handles *any* type of keys)

>  it definitely makes sense to make the ttl configurable imo.

Agree.
Rest looks good.

fauvel:
@njam last review please...

njam:
lgtm

fauvel:
Tagged 1.127.3

PR #1821, opened 2015-06-25
Location formField's "detectLocation" doesn't work

vogdb:
Due to Promises for ajax?

The formField `CM_FormField_Location` has a button "Get current location". It doesn't work anymore. You can test it for example on the "example" or "search" page.

@vogdb can you check?

### DIFF
--- a/library/CM/FormField/Location.js
+++ b/library/CM/FormField/Location.js
@@ -46,19 +46,21 @@ var CM_FormField_Location = CM_FormField_SuggestOne.extend({
     this.$('.detect-location').addClass('waiting');
 
     var self = this;
-    var deferred = $.Deferred();
-    navigator.geolocation.getCurrentPosition(deferred.resolve, deferred.reject);
 
-    deferred.then(function(position) {
-      self._lookupCoordinates(position.coords.latitude, position.coords.longitude);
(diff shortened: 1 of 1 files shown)


vogdb:
Yes! @njam or @tomaszdurka please review.

njam:
- _lookupCoordinates: Adjust docu
- detectLocation: Add docu
- detectLocation: Why are you wrapping `navigator.geolocation.getCurrentPosition()` into a new Promise? Couldn't we just use the result of that function?
- detectLocation: Let's stick to a convention regarding indentation of "then/catch/finally"? I would suggest to put them each onto a new line:
```js
getPromise()
.then(function(){
 //
})
.catch(function(){
 //
})
```
wdyt?

vogdb:
> - detectLocation: Why are you wrapping `navigator.geolocation.getCurrentPosition()` into a new Promise? Couldn't we just use the result of that function?

Honestly I can't think of a way to do it. Please share your solution.

vogdb:
@njam please look.

njam:
lgtm!

Might there be other cases like this? Did you search for "Deferred"?

vogdb:
good point. Can I use this PR for them?

PR #1822, opened 2015-06-26
GA: Client-side event tracking with a Redis queue

fauvel:


fauvel:
@njam please review. I've implemented a queue for virtual page views, too.

njam:
CMService_GoogleAnalytics_Client:
- I still think it's weird that you *need* to configure a TTL here. Agree it should be configurable, but I think the default of *no TTL* is not a good default (because of the load it puts on redis). How about we default to 86400?
- trackPageView: Move popping of queue into separate function?
- trackPageView: Using the data array's items as arguments for these functions seems a bit edgy. (It's not easy to read and understand, and prone to failing on changes because not very strict). How about just extracting elements from the array and passing it to the function normally?

fauvel:
@njam please review!

njam:
lgtm!

njam:
...one thing: shouldn't what's now called "event" be called "hit". This is the generic type of tracking in GA. Otherwise we have two things called "event".

fauvel:
@njam please review

njam:
lgtm, tests succeed?

fauvel:
Yep, tagged 1.127.6

PR #1786, opened 2015-06-29
Model default data is not validated during commit

fauvel:
Data field values are only validated when setting the value. 
It's possible to commit invalid (empty) value when commiting without `_set`.

### DIFF
--- a/library/CM/Model/Abstract.php
+++ b/library/CM/Model/Abstract.php
@@ -42,7 +42,6 @@ abstract class CM_Model_Abstract extends CM_Class_Abstract
             $this->_id = self::_castIdRaw($id);
         }
         if (null !== $data) {
-            $this->_validateFields($data);
             $this->_setData($data);
         }
         foreach ($this->_getAssets() as $asset) {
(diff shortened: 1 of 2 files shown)


fauvel:
@tomaszdurka please review.

tomaszdurka:
Hmm, I could have poorly describe my pull-request. 

Actually refactoring you did doesn't cover problem I've encountered.
I meant that if you create empty model object and do not run any kind of setData on it it won't validate. Even if you commit.

Example code:
```php

$model = new CM_Model_Location_City();
$model->commit();

$row = CM_Db_Db::select($model->getTableName(), '*', ['id' => $model->getId()])->fetch();
var_dump($row);
```
Above could dumps:
```
array(7) {
  'id' =>
  string(6) "367989"
  'stateId' =>
  NULL
  'countryId' =>
  string(1) "0"
  'name' =>
  string(0) ""
  'lat' =>
  NULL
  'lon' =>
  NULL
  '_maxmind' =>
  NULL
}
```
`countryId` is `0` while schema marked it as non-optional model: `'countryId' => array('type' => 'CM_Model_Location_Country'),`

fauvel:
@tomaszdurka thanks for clarifying. Please review!

fauvel:
@tomaszdurka I've covered the use case you had in mind, finally :) Please review!

tomaszdurka:
Hey @fauvel can you take a look at tests?

fauvel:
Timeout on Travis

library/CM/Model/Abstract.php
```diff
@@ -436,12 +434,20 @@ protected function _getSchemaData($data = null) {
...
-                $schema->validateField($key, $value);
+            if ($checkMissingFields) {
+                foreach ($schema->getFieldNames() as $key) {
+                    $value = isset($data[$key]) ? $data[$key] : null;
+                    $schema->validateField($key, $value);
+                }
+            } else {
+                foreach ($data as $key => $value) {
+                    $schema->validateField($key, $value);
+                }
```
tomaszdurka: What about checking all the fileds always?
I guess those which are optional should stay valid, all the others should be invalid if empty.

fauvel: When creating a new model and setting single fields initially, you obviously don't want to do that...

fauvel:
@tomaszdurka good to merge?

tomaszdurka:
Let me check again quickly.

fauvel:
*bump* :)

PR #1823, opened 2015-06-26
Fix for circle clickFeedback

christopheschwyzer:
Follow up: https://github.com/cargomedia/CM/pull/1815
Fixes Safari issue. See https://gist.github.com/adamcbrewer/5859738

PR #1824, opened 2015-06-29
Fix for clickFeedback

christopheschwyzer:
follow up https://github.com/cargomedia/CM/pull/1823


christopheschwyzer:
@vogdb please review

christopheschwyzer:
Tested this in FF, Chrome, Safari, iOS Safari, IE 10, Windows Phone 8
On Android 4.1 Stock the circle feedback doesn't work - I think we can ignore it as it isn't too bad.

vogdb:
Perfecto!

PR #1825, opened 2015-06-29
Pass GA userId as a string

fauvel:


fauvel:
@njam please review. We are setting the `userId` twice, once with `ga("create")` and once with `ga("set", "userId")`. The first one was passed as an integer, the second one as a string... It should always be a string according to the docs. Maybe this is related to our tracking issues?

njam:
lgtm

fauvel:
Tagged 1.127.9

PR #1829, opened 2015-07-03
GA tracking: escaping

fauvel:
Our GA client doesn't correctly escape custom fields, pageviews and possibly other values.
This seems to be the reason for the missing data when doing AdWords linking.

@fauvel could you check all the places in `getJs()`?

Also when investigating I noticed that when using MP for tracking events the numbers in our database match GA's numbers almost 100%. But when using frontend tracking we have a ~5% discrepancy.
How about switching back to MP, as this was not the reason for the wrong tracking? To see if everything works then - it would give us more precise data!


fauvel:
@njam please review

njam:
lgtm!

fauvel:
Tagged 1.128.0

PR #1830, opened 2015-07-03
Fix form event communication.

vogdb:
We shouldn't throw on failed submit cause form state is delivered by `cm.event` mechanism and not promise mechanism.


vogdb:
@njam please review.

njam:
lgtm

njam:
lgtm

PR #1831, opened 2015-07-07
Introduce _getEnvironment on abstract Response

tomaszdurka:
Introduce getter for environment on the abstract response.
Method would be copy of internal code from getRender (on response).

```php
/**
 * @return CM_Frontend_Environment
 */
protected function _getEnvironment() {
    $location = $this->getRequest()->getLocation();
    $currency = null !== $location ? CM_Model_Currency::findByLocation($location) : null;
    return new CM_Frontend_Environment($this->getSite(), $this->getRequest()->getViewer(), $this->getRequest()->getLanguage(), null, null, $location, $currency);
}
```

It would be useful in cases where we want environment without Render. This probably leads to idea of moving CM_Frontend_Render out of the Frontend namespace...


fauvel:
lgtm!

PR #1832, opened 2015-07-07
Get rid of cm.error

vogdb:
- Replace `triggerThrow` with normal `throw`
- Introduce configurable global error handler:
 - cm.error.ready() [calls setErrorHandler()]
 - cm.error.setErrorHandler(Function)
- Remove `Form._stopErrorPropagation`


njam:
- Would it make sense to have our own `Error` class (with `msg` and other fields we use) to have the fields documented? Then maybe also `cm.error.create()` is not needed because we can just use `new CM_Exception(msg)`?
- setGlobalHandler: Add docu
- _globalHandler: Add type for param docu
- CM/FormField/Abstract.js: Throw an error if `$container` is empty (for developers), maybe check it at the beginning?
- CM/FormField/Location.js: Use `CM_FormField_Abstract.error()` (which will be shown directly below the field)?

vogdb:
> - Would it make sense to have our own `Error` class (with `msg` and other fields we use) to have the fields documented? Then maybe also `cm.error.create()` is not needed because we can just use `new CM_Exception(msg)`?

Agree. Where I can declare such a class?

njam:
When catching exceptions and expecting it is a `CM_Exception` (having type, message etc) we need to make sure it is indeed a `CM_Exception` right (e.g. in `Example.js`)? I understand we can use bluebird for that (special `catch()` version)?

Rest looks very good!

njam:
Could we now easily get rid of `error.type` and introduce actual classes for errors?
E.g. this:
```js
      .catch(CM_Exception, function(error) {
        if ('CM_Exception_FormFieldValidation' == error.type) {
          self.error(error.message);
        } else if (value == self.getValue()) {
          throw error;
        }
      });
```
could then be:
```js
          self.error(error.message);
      });
```

wdyt?

njam:
Exception.js:
- CM_Exception: Why do we set `temp.name`?
- factory: I would throw plain `CM_Exception` here (and use `CM_Exception_Nonexistent` only for missing models and pages, like in php)
- factory: Exception msg: "No such exception X" instead of "No such exception as X"
- How about assigning new exception explicitly to `global` (instead of doing it in `extend()`) - I think it might be a bit clearer - wdyt?

@tomaszdurka maybe you wanna take a look here too?

vogdb:
> - CM_Exception: Why do we set `temp.name`?

Ahm , just a precaution. We can remove it.

njam:
Should we introduce separate files for each exception? Like we do for views etc?

library/CM/Exception.js
```diff
@@ -0,0 +1,62 @@
+(function(global) {
```
tomaszdurka: Use something like `scope` instead of `global`. `global` is a specific scope which you don't know inside this function.

library/CM/Component/Example.js
```diff
@@ -86,8 +86,8 @@ var CM_Component_Example = CM_Component_Abstract.extend({
 
   error_500_text_callback: function() {
     var self = this;
-    this.ajax('error', {status: 500, text: 'Errortext'}).catch(function(error) {
-      self.error('callback( type:' + error.type + ', msg:' + error.msg + ' )');
+    this.ajax('error', {status: 500, text: 'Errortext'}).catch(CM_Exception, function(error) {
```
tomaszdurka: Is this something promise specific or clear bluebird implementation? Question is - should we use features like these?

library/CM/App.js
```diff
@@ -223,72 +224,30 @@ var CM_App = CM_Class_Abstract.extend({
...
-        cm.window.hint(msg);
-      } else {
-        if (type) {
-          msg = type + ': ' + msg;
-        }
-        cm.window.hint(msg);
-        if (window.console && console.error) {
-          console.error('Error: ' + msg);
+
+    ready: function() {
```
tomaszdurka: I understand `ready` is some convention we have for cm app submodules, but I would rather go with 'cm.error.addGlobalHandlers()`, wdyt?

tomaszdurka:
I like the changes, looks good, couple of small notes.

vogdb:
> Should we introduce separate files for each exception? Like we do for views etc?

@njam I'v tried. Didn't work out. `library.js` couldn't be constructed after I've created a file `CM/Exception/AuthRequired.js` with the content:
```js
  window.CM_Exception_AuthRequired = CM_Exception.extend('CM_Exception_AuthRequired');
```
Error:
```
GET /library.js 
1436287090.js:1 Uncaught ReferenceError: cm is not defined(anonymous function) @ 1436287090.js:1
blogpost?blogpost=104:6197 Uncaught ReferenceError: cm is not defined
```

library/CM/Component/Example.js
vogdb: It's a bluebird's feature. In its defence I can tell that `.catch` is a bluebird's feature too. I guess we can use it while there is no standard about `catch`.

library/CM/App.js
vogdb: Consistency! We must be consistent.

njam:
lgtm, @tomaszdurka agree?

tomaszdurka:
Yes it is ok. I am againt this kind of consistency (`.ready`). Would be better if names were more informative. Let's leave it though.

njam:
- Global handler should only display (hint) errors of type `CM_Exception` - otherwise *all* JS errors are shown to the user!

PR #1836, opened 2015-07-10
Add ability to use 'replace' SQL operator in CM_Model_StorageAdapter_Database

fvovan:
We need for mechanism of atomic db changes to prevent race condition errors.
So we decided to implement one for Mysql adapter only yet, add new ```$options``` param to ```create()``` method and one available option will be 'useReplace'
As far as I understand ```create``` is used only from ```commit()``` method, so there are two options 
1. To pass '$useReplace' param directly to ```commit()```
2. To create whole model with this additional param (e.g pass to constructor).

@tomaszdurka , @njam  WDYT ?

### DIFF
--- a/library/CM/Model/Abstract.php
+++ b/library/CM/Model/Abstract.php
@@ -52,31 +52,43 @@ abstract class CM_Model_Abstract extends CM_Class_Abstract
     }
 
     /**
-     * @param array $options|null
+     * @param array $createOptions |null
      * @throws CM_Exception_Invalid
+     * @throws CM_Exception_InvalidParam
      * @throws CM_Exception_Nonexistent
+     * @throws CM_Exception_NotImplemented
      */
-    public function commit(array $options = null) {
+    public function commit(array $createOptions = null) {
         $persistence = $this->_getPersistence();
         if (!$persistence) {
             throw new CM_Exception_Invalid('Cannot create model without persistence');
         }
+        $type = $this->getType();
+
         if ($this->hasIdRaw()) {
             $dataSchema = $this->_getSchemaData();
             if (!empty($dataSchema)) {
-                $persistence->save($this->getType(), $this->getIdRaw(), $dataSchema);
+                $persistence->save($type, $this->getIdRaw(), $dataSchema);
             }
 
             if ($cache = $this->_getCache()) {
-                $cache->save($this->getType(), $this->getIdRaw(), $this->_getData());
+                $cache->save($type, $this->getIdRaw(), $this->_getData());
             }
             $this->_onChange();
         } else {
-            $this->_id = self::_castIdRaw($persistence->create($this->getType(), $this->_getSchemaData(), $options));
+            if (null !== $createOptions) {
+                $this->_validateCreateOptions($createOptions);
+                if (!empty($createOptions['useReplace']) && !($persistence instanceof CM_Model_StorageAdapter_ReplaceableInterface)) {
+                    throw new CM_Exception_NotImplemented('Param `useReplace` is not allowed for this adapter');
+                }
+            }
+
+            $idRaw = $persistence->create($type, $this->_getSchemaData(), $createOptions);
+            $this->_id = self::_castIdRaw($idRaw);
 
             if ($cache = $this->_getCache()) {
                 $this->_loadAssets(true);
-                $cache->save($this->getType(), $this->getIdRaw(), $this->_getData());
+                $cache->save($type, $this->getIdRaw(), $this->_getData());
             }
             $this->_onChange();
             $this->_changeContainingCacheables();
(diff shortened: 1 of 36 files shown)


fvovan:
It's ready. I had to adjust some tests to add testing for replace operator.
@njam please review

njam:

library/CM/Model/Abstract.php
```diff
@@ -67,7 +72,7 @@ public function commit() {
             }
             $this->_onChange();
         } else {
-            $this->_id = self::_castIdRaw($persistence->create($this->getType(), $this->_getSchemaData()));
+            $this->_id = self::_castIdRaw($persistence->create($this->getType(), $this->_getSchemaData(), $options));
```
tomaszdurka: Could you extract this into two lines? I think it will be bit more visible. Same for `save()`.

library/CM/Model/Abstract.php
```diff
@@ -51,7 +51,12 @@ public function __construct($id = null) {
         $this->_getData(); // Make sure data can be loaded
     }
 
-    public function commit() {
+    /**
+     * @param array $options|null
+     * @throws CM_Exception_Invalid
+     * @throws CM_Exception_Nonexistent
+     */
+    public function commit(array $options = null) {
```
tomaszdurka: Passing along these options down to create seems a bit dodgy. They would need to be called `$createOptions` then.

I rather propose to check and extract those params before using them in create method.

tomaszdurka: Or what about simple flag for now ? (instead of options)

tomaszdurka:
Concept generally looks ok.

One thing though :) I am totally against passing arrays like this especially 2 levels down.

So either I would:
- validate array and recreate it while passing to `create` method *or*
- pass boolean replace

Additionally what about:
1. Adding interface for replace method
2. Implement that interface on database storage adapter
3. Read `replace` flag and check if storage adapter implements replace interface (throw otherwise)

@fvovan @njam wdyt?

fvovan:
I opt for validating array. Otherwise it would have few 'null' params when we decide to add more params. 
But may be we will never add it...

And why we need to recreate it? If we validate array, we can assume it's safe.
Interface sounds good, maybe it is a bit overkill as well as $options array, because this interface will have only 1 implementation.

What should I extract for ```save()```? i see that only $this->getType can be extracted there.

fvovan:
btw, should we throw during validation or just silently omit  $createOptions?

fvovan:
added Replaceable Interface and params validation
@tomaszdurka , @njam please review

tomaszdurka:
If we keep `replace` as one of this *bloody* :) options array we need to validate it also inside create method. Otherwise you can pass there random things. If we keep current approach it's probably wise to move logic (validation, interface check) down into adapter... maybe. But then we end up with params not validated on abstract level. We could move it to abstract adapter.

What about keeping it super-simple for now with `replace` as flag? I know having options might be easier for adding extra options later on - as there will be no need to convert boolean usages into arrays. But all these options will be mostly for different things and most-likely never used in combination. 
It will simplify code so much:
- no need for validation
- no array accessors (riddle for ide)
- no mystical options bag
- simple boolean and interface

@fvovan what are advantages of options?

fvovan:
@tomaszdurka advantages are obvious , much more convenient usage later without passing " null, null, true ". 
I don't see a problem passing 'random things'. Anyway it will be just ignored and moreover ```create()``` now is used only from Model's ```commit``` 
Actually I  don't mind to implement any of these options, and even go java way and create new class for createParams and encapsulate all validations there.

fvovan:
now It's with single boolean param,
@tomaszdurka , @njam wdyt ?

tomaszdurka:
I wouldn't say they are obvious. After having some experience with frameworks based on arrays of arrays I can tell signature with less arguments is not really simpler. It seems like that on beginning but once code grows and you pass those options down it becomes nightmare. I would rather go "java way" than keeping the array. It's also hard to document associative array parameters in php. 

Passing random things doesn't mean they won't do hard, but imagine having some parameter which influences how the application works. Once you'll remove it feature you won't see any kind of exception that there is in invalid (not-affecting) param passed.

---------------------

Regarding the code:
- Currently you've added interface for replace functionality, but replace is actually parameter for `create` method, even for those adapters which does not implement `replace` (mongodb). Let's make it on top of Adapter class not tangled into it.
- Let's use interface for methods, implement `replace` on Replaceable interface.
- Keep create as simple as is.


Like this?
```php
if ($useReplace) {
  if (!$persistence instanceof CM_Model_StorageAdapter_ReplaceableInterface) {
    throw new CM_Exception_NotImplemented('Param `useReplace` is not allowed for this adapter');
  }
  $idRaw = $persistence->replace($type, $this->_getSchemaData());
} else {
  $idRaw = $persistence->create($type, $this->_getSchemaData());
}
$this->_id = self::_castIdRaw($idRaw);
```

fvovan:
@tomaszdurka , it is ready

fvovan:
@tomaszdurka, could you review it?
I merged some changes related to model validation

tomaszdurka:
Looks good, just one note. We have convention to use only `null` for default values for parameters. 
Can you change `$replace` to ` = null`? Thanks

PR #1837, opened 2015-07-13
Promise behaviour of Form.submit()

vogdb:
`CM_Form_Abstract.submit()` returns a Promise which behaves like this:
- Client side validation errors: resolve
- Server side validation errors: resolve
- Successful submit: resolve

Why do we not reject on validation errors?
I have a usage where I submit a form directly and use the returned Promise to control the further control flow. In that case I would expect that the promise only resolves on successful submit. 
wdyt?

### DIFF
--- a/library/CM/Form/Abstract.js
+++ b/library/CM/Form/Abstract.js
@@ -45,7 +45,9 @@ var CM_Form_Abstract = CM_View_Abstract.extend({
       $btn.on(event, {action: name}, function(event) {
         event.preventDefault();
         event.stopPropagation();
-        return handler.submit(event.data.action).catch(function(error){});
+        return handler.submit(event.data.action).catch(function(error) {
+          //error is already handled and displayed in `submit`
+        });
       });
     }, this);
 
--- a/library/CM/Form/Abstract.js
+++ b/library/CM/Form/Abstract.js
@@ -191,7 +191,6 @@ var CM_Form_Abstract = CM_View_Abstract.extend({
               }
             }
 
-            handler.trigger('error error.' + action.name);
             throw cm.error.create('error error.' + action.name);
           }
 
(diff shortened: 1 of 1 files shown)


vogdb:
@njam smth like this ? Don't know what to do with them, once we caught them on the click level.

njam:
- Don't we call `handler.trigger()` twice now?
- Didn't we remove `cm.error.create` - just use `throw new CM_Exception('Form validation failed')`?
- Maybe introduce a specific error `CM_Exception_FormFieldValidation`, and *only* catch that in the click case?
 - Just add a one-line comment into the catch that the `submit()` already handles and displays errors?

vogdb:
> - Didn't we remove `cm.error.create` - just use `throw new CM_Exception('Form validation failed')`?

Yes, we did but this issue is not yet merged, so I can't use CM_Exception.

> - Maybe introduce a specific error `CM_Exception_FormFieldValidation`, and *only* catch that in the click case?

Yes but only after CM_Exception is introduced.

>  - Just add a one-line comment into the catch that the `submit()` already handles and displays errors?

ok.

njam:
ok let's wait for https://github.com/cargomedia/CM/pull/1832

vogdb:
@njam wdyt now ?

njam:
- CM_Exception_FormFieldValidation: Should the constructor then accept an `errorList` array argument?
- CM/Form/Abstract.js: Add space after `//` in comment

vogdb:
> - CM_Exception_FormFieldValidation: Should the constructor then accept an `errorList` array argument?

Not needed. CM_Exception_FormFieldValidation <= CM_Exception <= Error. So the `errorList->toString()` will be called to set the message of Error. For now it's fine.

vogdb:
btw @njam please review.

njam:
ok, lgtm

PR #1827, opened 2015-07-13
Fix pseudo-element css selectors

christopheschwyzer:
Should use double colon for pseudo-elements: E.g. `::after` instead of `:after`.

Interesting article http://bitsofco.de/2015/pseudo-classes-pseudo-elements-and-colon-notation/?utm_source=CSS-Weekly&utm_campaign=Issue-169&utm_medium=email


christopheschwyzer:
@vogdb please review

vogdb:
looks very good but also we need to work on `scrollShadow` plugin.

christopheschwyzer:
@vogdb oh you're right - missed that one ;) please verify

vogdb:
lgtm

PR #1838, opened 2015-07-13
Add namespaces for throttled promises

tomaszdurka:


vogdb:
the new option that we discussed will be added in #1839.

vogdb:
I think that `throttleFunction` is not a correct name cause we throttle here promise not function. Promise is passed by function because we can't pass it directly. It's a javascript's restriction. So, I still suggest to call it `throttlePromise` or `promiseThrottled` or smth close to it. Also I want to propose to have only one function and merge the current `throttle` into the current `throttleFunction`. We can make `namespace` to be part of options or a third argument to the function.

tomaszdurka:
Hmm, I think we should rework it a bit too.

Now we have 2 functions:
- `throttleFunction` wraps function returning promise and "throttles" it internally
- `throttle` wraps function passed, stores it with certain key and executes it. It also assumes function passed returns Promise.

1. I agree we could have just one function (`throttle`) which would do all. What about simply not exposing `throttleFunction` method public.
2. I would suggest to not include Promise in the name of the method, but rather detect if return of callback is Promise and if it's not simply pass it to newly created, resolved Promise. This way it would be very universal function returning a Promise.

@vogdb wdyt?

vogdb:
> 1. I agree we could have just one function (`throttle`) which would do all. What about simply not exposing `throttleFunction` method public.

How do you plan to use `throttle` for the cases of `throttleFunction`. Is my proposition to move `certain key` to be the third argument ok to you?

> 2. I would suggest to not include Promise in the name of the method, but rather detect if return of callback is Promise and if it's not simply pass it to newly created, resolved Promise. This way it would be very universal function returning a Promise.

Do we have a real need for this? I think we should not create a super universal perfect function for all cases that we might need. Lets just create a good one for the current needs. I don't see the point in wrapping a simple function into a resolved promise.

vogdb:
As agreed I've tried to make use of `_.memoize(promiseThrottler())`.
```js
  var resolve5 = _.memoize(promiseThrottler(function(arg) {
    return new Promise(function(resolve, reject) {
      setTimeout(function() {
        console.log('### resolved ' + arg);
        resolve(arg)
      }, 1000);
    });
  }));

  resolve5(5);
  resolve5(5);
  resolve5(5);
  setTimeout(function(){resolve5(5)}, 1300);
  setTimeout(function(){resolve5(5)}, 3000);
```

The above example shows that `resolve5` resolves only ones but should resolve 3 times. It does not do that because memoize cache stays forever and is not reset after the first promise is resolved.
@tomaszdurka your thoughts.
btw don't be afraid we can always revert my recent changes and restore your version of `throttle`.

tomaszdurka:
Ah, true.
This actually caches result of function - not the passed function itself. We could figure that out after seeing it ony accepts function. We need some simple caching/storage mechanism which accepts any type of value.

vogdb:
Agree! Shall we revert here and keep `throttle` and `throttleFunction` or smth similar to them?

vogdb:
@tomaszdurka please look.
Also wdyt about smth like this:
```js
  promiseThrottler.clearKey = function(key) {
    if (key) {
      if (storage[key]) {
        delete storage[key];
      }
    } else {
      _.each(storage, function(key) {
        delete storage[key];
      })
    }
  };
```

tomaszdurka:
Fine with me! :)

tomaszdurka:
Good job! :panda_face:

PR #1840, opened 2015-07-15
JS error handling when Internet down

vogdb:


njam:
How about "No Internet connection"?
Should be added to translations!

vogdb:
Do I need to add it to `de.php` ?

njam:
Yes! "Keine Internetverbindung"

Btw you can pass only 1 value to `setTranslation()`, and it will be used for "phrase" and "value":
`setTranslation('foo')` instead of `setTranslation('foo', 'foo')`
Maybe you can change it for the whole file so it's clear?

njam:
lgtm

PR #1842, opened 2015-07-16
Layout.loadPage() should return Promise

vogdb:
@vogdb wdyt?


vogdb:
@njam please look.

njam:
Thanks :) lgtm

Issue #1437, opened 2014-10-08
Notify user of new app version during active session


njam:
Problem: when a new application version is deployed the client would need to reload his browser to get the new client code.

I see three options:
- A: Reload automatically on new version (compare client with server version).
 - Downside: Will lose unsaved changes, might be too often
- B: Notify user about new version, hint to reload.
 - Downside: Might be too often, will irritate users
- C: Hint to reload on uncatched errors
- D: When detecting a new version set a flag on the *router*, so that is triggers a hard-reload on the next link clicked, instead of an ajax-navigation.

@mariansollmann @tomaszdurka wdyt?

njam:
Closing in favor of https://github.com/cargomedia/CM/issues/1813

PR #1813, opened 2015-07-17
Reload Page on new Application Release

vogdb:
Idea **D** via https://github.com/cargomedia/CM/issues/1437

1. Any ajax/form response from the server to the client could include the server's `releaseStamp`. We could return it from `CM_Http_Response_View_Abstract::_process()`.
2. If the client detects that its `releaseStamp` is different from what the server has it will set a flag `shouldReload`.
3. When the client navigates to a different page and the `shouldReload` flag is set then the page will be hard-reloaded (instead of using ajax navigation).

@tomaszdurka @dlondero @vogdb wdyt?

#### Steps
* [x] Return releaseStamp in ajax-requests https://github.com/cargomedia/CM/pull/1841
* [ ] ...

### DIFF
--- a/library/CM/App.js
+++ b/library/CM/App.js
@@ -12,12 +12,6 @@ var CM_App = CM_Class_Abstract.extend({
   /** @type Object **/
   options: {},
 
-  /** @type {Boolean} **/
-  _shouldReload: false,
-
-  /** @type {Number|Null} **/
-  _deployVersion: null,
-
   ready: function() {
     this.error.ready();
     this.dom.ready();
@@ -27,12 +21,11 @@ var CM_App = CM_Class_Abstract.extend({
     this.router.ready();
   },
 
-  updateDeployVersion: function(version) {
-    if (this._deployVersion && this._deployVersion != version) {
-      this._shouldReload = true;
-    } else {
-      this._deployVersion = version;
-    }
+  /**
+   * @returns {Number}
+   */
+  getDeployVersion: function() {
+    return cm.options.deployVersion;
   },
 
   /**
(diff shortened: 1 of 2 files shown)


vogdb:
@njam please look

njam:
To keep it symmetrics with server side part I would suggest:
- Add `CM_App.getDeployVersion()`: `return cm.options.deployVersion;`
- _shouldReload: Move this into `cm.router`?
- updateDeployVersion: Will be simpler - inline?
- Mechanism to hard-reload page will be better placed in `cm.router.route()` instead of `Layout.loadPage()` (because it covers more cases, and there's already the field `forceReload`)

vogdb:
@njam please look.

njam:
With the underscore in `_shouldReload` we indicate that it should be used internally only. How about adding some setter? `cm.router.enableShouldReload()` - naming suggestions welcome.

vogdb:
@njam wdyt about `forceReload` ? btw the builds are keep failing...

njam:
sgtm

Why did you replace `this._shouldReload`  with `cm.router._shouldReload`?

vogdb:
I'm still not sure about this replacement. Such replacement saves us from a wrong `this` context and I saw that we use `cm` instead of `this` in a lot of places. So I've decided to be consistent!

njam:
Ok, but further down in `route()` we use `this` again.
I would stick to `this` unless it's a public API?

vogdb:
@njam please look.

njam:
lgtm!

(haven't tested - but guess you've done that?)

vogdb:
Yep.

PR #1846, opened 2015-07-20
CM_Layout_Abstract: Make sure "_timeoutLoading" is always cleared

vogdb:
In case of `Promise.CancellationError` it will not be cancelled.
Maybe put it into a `finally` block?


vogdb:
> Maybe put it into a `finally` block?

I think it's better to cancel this timeout as soon as possible cause some operations in `then` are heavy, so the timeout can clash with them and make a mess.

@njam please review.

njam:
lgtm

PR #1847, opened 2015-07-20
Debug composer on travis

njam:


njam:
@vogdb please review

Probably not a good idea to update to composer master on every release :)

vogdb:
oh, lgtm!

PR #1272, opened 2014-06-10
Breaking: Split FormField->validate

cm-jan:
Split `FormField->validate` into:

```
mixed function parseUserInput(string $userInput);
void function validate(mixed $value);
```

`CM_Form_Abstract->process` would contain something like:

``` php
try {
    $value = $field->parseUserInput($fieldValue);
    $field->validate($environment, $value);
    $formData[$fieldName] = $value;
} catch (CM_Exception_FormFieldValidation $e) {
    $response->addError($e->getMessagePublic($response->getRender()), $fieldName);
}
```

PR #1844, opened 2015-07-20
Change serialization for location formField

vogdb:
For: https://github.com/cargomedia/sk/pull/11

Currently we serialize location objects as `$level.$id` in `CM_FormField_Location`.
We should change it to `toArray()`-serialization, which is an array `{level: $level, id: $id, _class: $className}`.

The goal is to get json objects on `change` in JS:
```js
locationFormField.on('change', function(value) {
 // value is a JSON object with "level" and "id"
});
```

@vogdb you wanna do it? Is it possible like this?

### DIFF
--- a/library/CM/FormField/Location.js
+++ b/library/CM/FormField/Location.js
@@ -33,7 +33,7 @@ var CM_FormField_Location = CM_FormField_SuggestOne.extend({
       var distanceEnabled = false;
       var items = this.getValue();
       if (items.length > 0) {
-        distanceEnabled = items[0].level >= this.getOption("distanceLevelMin");
+        distanceEnabled = JSON.parse(items[0].id).level >= this.getOption("distanceLevelMin");
       }
       this.getDistanceField().$("input").prop("disabled", !distanceEnabled);
     }
--- a/library/CM/FormField/Location.js
+++ b/library/CM/FormField/Location.js
@@ -76,13 +76,5 @@ var CM_FormField_Location = CM_FormField_SuggestOne.extend({
    */
   _lookupCoordinates: function(lat, lon) {
     return this.ajax('getSuggestionByCoordinates', {lat: lat, lon: lon, levelMin: this.getOption('levelMin'), levelMax: this.getOption('levelMax')});
-  },
-
-  _getSelect2Options: function() {
-    return {
-      id: function(data) {
-        return JSON.stringify(_.pick(data, 'id', 'level'));
-      }
-    }
   }
 });
--- a/library/CM/FormField/Location.js
+++ b/library/CM/FormField/Location.js
@@ -31,31 +31,31 @@ var CM_FormField_Location = CM_FormField_SuggestOne.extend({
   updateDistanceField: function() {
     if (this.getDistanceField()) {
       var distanceEnabled = false;
-      var items = this.getValue();
-      if (items.length > 0) {
-        distanceEnabled = items[0].id.level >= this.getOption("distanceLevelMin");
+      var value = this.getValue();
+      if (value) {
+        distanceEnabled = value.id.level >= this.getOption("distanceLevelMin");
       }
       this.getDistanceField().$("input").prop("disabled", !distanceEnabled);
     }
   },
 
   /**
-   * @return {Array}
+   * @return {Object}
    */
   getValue: function() {
     var value = CM_FormField_SuggestOne.prototype.getValue.call(this);
-    if (_.isString(value[0].id)) {
-      value[0].id = JSON.parse(value[0].id);
+    if (_.isString(value.id)) {
+      value.id = JSON.parse(value.id);
     }
     return value;
   },
 
   /**
-   * @param {Array} value
+   * @param {Object} value
    */
   setValue: function(value) {
-    if (_.isObject(value[0].id)) {
-      value[0].id = JSON.stringify(value[0].id);
+    if (_.isObject(value.id)) {
+      value.id = JSON.stringify(value.id);
     }
     return CM_FormField_SuggestOne.prototype.setValue.call(this, value);
   },
--- a/library/CM/FormField/Location.js
+++ b/library/CM/FormField/Location.js
@@ -9,6 +9,13 @@ var CM_FormField_Location = CM_FormField_SuggestOne.extend({
     'click .detectLocation': 'detectLocation'
   },
 
+  /** @type {Object} */
+  _select2Options: {
+    id: function(data) {
+      return JSON.stringify(_.pick(data, 'id', 'level'));
+    }
+  },
+
   ready: function() {
     CM_FormField_SuggestOne.prototype.ready.call(this);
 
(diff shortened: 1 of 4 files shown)


vogdb:
@njam it's just a draft. please look.

library/CM/FormField/Location.php
```diff
@@ -18,12 +18,11 @@ public function getSuggestion($location, CM_Frontend_Render $render) {
         for ($level = $location->getLevel(); $level >= CM_Model_Location::LEVEL_COUNTRY; $level--) {
             $names[] = $location->getName($level);
         }
-        return array(
-            'id'   => $location->getLevel() . '.' . $location->getId(),
-            'name' => implode(', ', array_filter($names)),
-            'img'  => $render->getUrlResource('layout',
-                    'img/flags/' . strtolower($location->getAbbreviation(CM_Model_Location::LEVEL_COUNTRY)) . '.png'),
-        );
+        $result = $location->toArray();
```
njam: Let's not use `toArray()` here, it's rather random that the two share a common format.
The format here is defined by `CM_FormField_Suggest::getSuggestion()`, so it should be defined here exactly as needed.

library/CM/FormField/Location.php
```diff
@@ -34,11 +33,12 @@ public function getSuggestion($location, CM_Frontend_Render $render) {
      */
     public function validate(CM_Frontend_Environment $environment, $userInput) {
         $value = parent::validate($environment, $userInput);
-        if (!preg_match('/^(\d+)\.(\d+)$/', $value, $matches)) {
```
njam: Now we could replace this most of this function with `CM_Model_Location::fromArray()`?

library/CM/FormField/Suggest.js
```diff
@@ -15,7 +15,7 @@ var CM_FormField_Suggest = CM_FormField_Abstract.extend({
     var prePopulate = this._$input.data('pre-populate');
 
     this._$input.removeClass('textinput');
-    this._$input.select2({
+    this._$input.select2(_.extend({
```
njam: Is this the best/only way of doing this? Extending internals of a jquery plugin seems a bit intrusive..
Can we rather plug in that transformation in other places?

library/CM/FormField/Location.php
vogdb: It will not save too much code for us cause the errors that are raised by `CM_Model_Location::fromArray()` are not `CM_Exception_FormFieldValidation`. So, we can't actually remove our validation checks like `if (!is_int($id) ...` or `if ($level < $this->_options['levelMin'] ...` because they throw `CM_Exception_FormFieldValidation` which is well displayed in UI under the field. `CM_Model_Location::fromArray()` in its turn throws `CM_Exception_Invalid`.

library/CM/FormField/Suggest.js
vogdb: But we don't extend internals here. I only extend `select2` options cause they can't be change after the plugin is instantiated. I need somehow change `id` format of select2 in `Location` field but select2 is instantiated in `Suggest` field. That is why I need to pass this additional option from `Location` to `Suggest`.

library/CM/FormField/Location.php
njam: I see.. I would still do it. Because the validation we currently do is also only partial (ID might be invalid..).

Anyway I think we don't need to throw `CM_Exception_FormFieldValidation` here. This exception should be thrown for errors that we expect to happen (i.e. user puts some wrong data), so that we can tell the user what he did wrong. In this case though the user is not able to put such invalid data into the form field in normal operation (because the field only allows to select locations within the "levelMin" and "levelMax".
So if such data is sent to the server it means we either have an application bug, or a weird client. In these cases I think throwing some other exception is even better, as it will be logged?

library/CM/FormField/Location.js
```diff
@@ -9,6 +9,13 @@ var CM_FormField_Location = CM_FormField_SuggestOne.extend({
     'click .detectLocation': 'detectLocation'
   },
 
+  /** @type {Object} */
+  _select2Options: {
+    id: function(data) {
+      return JSON.stringify(_.pick(data, 'id', 'level'));
```
njam: Why pick only `id` and `level` - can't we just pass the output of toArray() to fromArray()?

library/CM/FormField/Suggest.js
njam: Oh get it now!
Maybe rename to `_getSelect2Options()`?

library/CM/FormField/Location.js
vogdb: Hmm but data is not the output of `toArray`. It is the output of `getSuggestion` that contains also extra redundant fields like `name` and `img`. Of course we can skip picking and simple pass `data`. It will work too.

library/CM/FormField/Location.php
vogdb: Got it. Makes sense.

vogdb: The only thing that does not make sense here is that this code is placed inside `validate` method. We don't throw Validation errors inside `validate` >_<.

vogdb:
@njam please review.

library/CM/FormField/Location.php
njam: Agree!
This is a big TODO - to split `validate()` into `parseUserInput()` and `validate()` - https://github.com/cargomedia/cm/pull/1272

Let's use `CM_Params::jsonDecode()` instead of `json_decode()` - it has additional error checking!

njam:
Could we instead put the whole json string of the location into the `id` field?
```
'id' => '{level: $level, id: $id, _class: $className}'
```

vogdb:
I guess we can. Should I try ?

vogdb:
@njam please review.

njam:
- CM/FormField/Location.php: validate: We should check if the level is within the configured boundaries (using `Location::getLevel()`). Otherwise hackers could send a *country* as their signup location while we want to restrict it to *city* or *zip*.
- CM/FormField/Location.js: Needs to be adjusted I guess?
- CM/FormField/Location.js: I guess the `id` is not automatically json-parsed on the client side - so we have a string and not an object available?

vogdb:
> - CM/FormField/Location.js: I guess the `id` is not automatically json-parsed on the client side - so we have a string and not an object available?

Yep. It is not parsed. If we automatically parse it and store as an object then we have no control over its value that is sent to the server. It means that the `id` will be sent as an Object. In javascript terms it means that the result of `toString` method on this Object will be sent. And that is `[Object object]`. Unfortunately I don't see an option of specifying the format of the sent because the format can be defined only in `CM_Form_Abstract::submit`.

Also please look at the changes.

njam:
Maybe we should override `getValue()` and `setValue()` to do the JSON encoding/decoding?

vogdb:
> Maybe we should override `getValue()` and `setValue()` to do the JSON encoding/decoding?

That's the thing. We can't change them on back-end because we use `validate` actually to set value. We can't do it either on front-end cause the values to submit are taken not by `get/setValue` but by `CM_Form_Abstract::getData` which uses `this.$().serializeArray();` to get the data to submit.

njam:
But on the frontend it would simplify the usage in `updateDistanceField()` and in the future for the entityQuery (using `getValue()`)?

vogdb:
Should I also change return type of `getValue` from `[Object]` to `Object` ?

njam:
Yes! But that should probably be done in `CM_FormField_SuggestOne` already?

vogdb:
ok.

vogdb:
@njam please look

njam:
lgtm!
I trust you tested this.

vogdb:
Yes!

PR #1848, opened 2015-07-21
Upgrade composer to alpha10

njam:
cc @dlondero @tomaszdurka

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -5,7 +5,7 @@
     "bin-dir": "bin/"
   },
   "require": {
-    "composer/composer": "1.0.0-alpha8",
+    "composer/composer": "1.0.0-alpha10",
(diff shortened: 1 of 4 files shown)


njam:
@tomaszdurka please review

`Composer\Config::get('vendor-dir')` now returns an absolute path by default (based on a `$baseDir` you pass to its constructor).
I've opted to retrieve the relative path nonetheless (using a flag) to keep our code mostly untouched.

tomaszdurka:
So there is a flag so config returns relative paths? Ok.
What about making `createComposer`'s `$baseDir` non optional?

njam:
agree, please double check!

tomaszdurka:
Looks good!

njam:
Tagged 1.133.1

PR #1852, opened 2015-07-21
ClipSlide late call problem.

vogdb:


vogdb:
@njam please review.

njam:
- I'm not sure it's good to use *underscore* in jquery plugins like these. Especially for this small use case it might be preferable to reduce dependencies. wdyt?
- I trust you tested this and it works in Firefox?

vogdb:
> - I'm not sure it's good to use *underscore* in jquery plugins like these. Especially for this small use case it might be preferable to reduce dependencies. wdyt?

Removed it.

> - I trust you tested this and it works in Firefox?

Nope but did it now. I can assure it wasn't only Firefox problem.

@njam please review.

njam:
lgtm

PR #1853, opened 2015-07-21
Use docker based travis build

njam:
Depends on having mongodb 2.6 on Travis: https://github.com/travis-ci/travis-ci/issues/2246

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,5 @@
 language: php
+sudo: false
 
 php:
   - 5.4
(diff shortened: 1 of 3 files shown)


njam:
Main problem here is that we have mongodb 2.4 on Travis which gives:
```
1) CM_MongoDb_ClientTest::testFind
MongoResultException: localhost:27017: unrecognized field "cursor
/home/travis/build/cargomedia/cm/vendor/lstrojny/functional-php/src/Functional/Map.php:40
/home/travis/build/cargomedia/cm/tests/library/CM/MongoDb/ClientTest.php:160
/home/travis/build/cargomedia/cm/tests/helpers/CMTest/library/CMTest/TestCase.php:23
```

Code in question:
```php
        $result = $mongoDb->find($collectionName, ['groupId' => 1], ['_id' => 0, 'foo' => 1], [['$unwind' => '$foo']]);
        $actual = \Functional\map($result, function ($val) {
            return $val;
        });
```

@alexispeter @kris-lab do you see a way to work around that? What's the reasoning for this `map()`?

kris-lab:
Not sure tbh! Maybe @alexispeter has some ideas?

alexispeter:
The map() is just a way to get an array containing the items the cursor returns. Using `iterator_to_array()` instead would work just as well.

alexispeter:
I'm confused. Don't these changes basically "downgrade" from 2.6 to 2.4 because you removed the manual installation?

This `unrecognized field "cursor` error occurs when the client requests a cursor for the result of an aggregation instead of the full set, which was only added in server version 2.6

njam:
Yes, I wanted to try using Travis' new Docker infrastructure (which is much faster). It doesn't allow to use `sudo` so installing mongodb from packages would be tricky.

Do you see a reasonable way of being mongodb 2.4 compatible? Or should we wait until travis updates mongodb to 2.6?

alexispeter:
Not unless you want to forgo cursors on aggregation-operations (we have quite a few).

njam:
Ok, let's wait for Travis upgrading mongodb then.

Waiting on: https://github.com/travis-ci/travis-ci/issues/2246

fauvel:
The PR is on its way for #1864: https://github.com/travis-ci/apt-package-whitelist/pull/597

njam:
Closing in favor of https://github.com/cargomedia/cm/pull/1864

PR #1851, opened 2015-07-22
Use CM_I18N_Phrase in CM_Exception

fvovan:
Current constructor of `CM_Exception`:
```php
    /**
     * @param string|null $message
     * @param array|null  $metaInfo
     * @param array|null  $options
     */
    public function __construct($message = null, array $metaInfo = null, array $options = null) {
        $this->_metaInfo = null !== $metaInfo ? $metaInfo : array();
        $this->_messagePublic = isset($options['messagePublic']) ? (string) $options['messagePublic'] : null;
        $this->_messagePublicVariables = isset($options['messagePublicVariables']) ? (array) $options['messagePublicVariables'] : null;
        if (isset($options['severity'])) {
            $this->setSeverity($options['severity']);
        }
        parent::__construct($message);
    }
```

Here we could now use `CM_I18n_Phrase` for `messagePublic` and `messagePublicVariables`. @fvovan wanna look into this when you have some time?

cc @tomaszdurka

### DIFF
--- a/library/CM/Action/Abstract.php
+++ b/library/CM/Action/Abstract.php
@@ -181,7 +181,9 @@ abstract class CM_Action_Abstract extends CM_Class_Abstract implements CM_ArrayC
 
         $isAllowed = call_user_func_array(array($this, '_isAllowed'), $arguments);
         if (false === $isAllowed) {
-            throw new CM_Exception_NotAllowed('Action not allowed', null, null, new CM_I18n_Phrase('The content you tried to interact with has become private.'));
+            throw new CM_Exception_NotAllowed('Action not allowed', null, null, [
+                'messagePublic' => new CM_I18n_Phrase('The content you tried to interact with has become private.'),
+            ]);
         }
         if (true !== $isAllowed) {
             $this->_checkActionLimits();
--- a/library/CM/Action/Abstract.php
+++ b/library/CM/Action/Abstract.php
@@ -181,7 +181,7 @@ abstract class CM_Action_Abstract extends CM_Class_Abstract implements CM_ArrayC
 
         $isAllowed = call_user_func_array(array($this, '_isAllowed'), $arguments);
         if (false === $isAllowed) {
-            throw new CM_Exception_NotAllowed('Action not allowed', null, array('messagePublic' => 'The content you tried to interact with has become private.'));
+            throw new CM_Exception_NotAllowed('Action not allowed', new CM_I18n_Phrase('The content you tried to interact with has become private.'));
         }
         if (true !== $isAllowed) {
             $this->_checkActionLimits();
--- a/library/CM/Action/Abstract.php
+++ b/library/CM/Action/Abstract.php
@@ -181,7 +181,7 @@ abstract class CM_Action_Abstract extends CM_Class_Abstract implements CM_ArrayC
 
         $isAllowed = call_user_func_array(array($this, '_isAllowed'), $arguments);
         if (false === $isAllowed) {
-            throw new CM_Exception_NotAllowed('Action not allowed', new CM_I18n_Phrase('The content you tried to interact with has become private.'));
+            throw new CM_Exception_NotAllowed('Action not allowed', null, null, new CM_I18n_Phrase('The content you tried to interact with has become private.'));
         }
         if (true !== $isAllowed) {
             $this->_checkActionLimits();
--- a/library/CM/Captcha.php
+++ b/library/CM/Captcha.php
@@ -14,7 +14,7 @@ class CM_Captcha {
         $this->_id = (int) $id;
         $this->_text = CM_Db_Db::select('cm_captcha', 'number', array('captcha_id' => $this->getId()))->fetchColumn();
         if (!$this->_text) {
-            throw new CM_Exception_Nonexistent('Invalid captcha id `' . $id . '`', null, array('severity' => CM_Exception::WARN));
+            throw new CM_Exception_Nonexistent('Invalid captcha id `' . $id . '`', null, CM_Exception::WARN);
         }
         $this->_fontPath = CM_Util::getModulePath('CM') . 'resources/font/comicsans.ttf';
     }
--- a/library/CM/Captcha.php
+++ b/library/CM/Captcha.php
@@ -14,7 +14,7 @@ class CM_Captcha {
         $this->_id = (int) $id;
         $this->_text = CM_Db_Db::select('cm_captcha', 'number', array('captcha_id' => $this->getId()))->fetchColumn();
         if (!$this->_text) {
-            throw new CM_Exception_Nonexistent('Invalid captcha id `' . $id . '`', null, CM_Exception::WARN);
+            throw new CM_Exception_Nonexistent('Invalid captcha id `' . $id . '`', CM_Exception::WARN);
         }
         $this->_fontPath = CM_Util::getModulePath('CM') . 'resources/font/comicsans.ttf';
     }
--- a/library/CM/Clockwork/Manager.php
+++ b/library/CM/Clockwork/Manager.php
@@ -45,7 +45,7 @@ class CM_Clockwork_Manager {
             return $event->getName() == $eventName;
         });
         if ($duplicateEventName) {
-            throw new CM_Exception('Duplicate event-name', null, null, ['name' => $eventName]);
+            throw new CM_Exception('Duplicate event-name', null, ['name' => $eventName]);
         }
         $this->_events[] = $event;
     }
@@ -99,7 +99,7 @@ class CM_Clockwork_Manager {
     protected function _getRunningEvent($identifier) {
         $eventName = array_search($identifier, \Functional\pluck($this->_eventsRunning, 'identifier'));
         if (false === $eventName) {
-            throw new CM_Exception('Could not find event', null, null, ['identifier' => $identifier]);
+            throw new CM_Exception('Could not find event', null, ['identifier' => $identifier]);
         }
         return $this->_eventsRunning[$eventName]['event'];
     }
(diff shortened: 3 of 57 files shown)


fvovan:
it's ready
@tomaszdurka , please review.

njam:
- _messagePublicPhrase: Let's call it just `_messagePublic` (as we usually don't include the variable type in its name
- We should have the possibility to pass a `Phrase` instance as `messagePublic` to the constructor. We could also keep supporting the old scheme, but if it's not too much hassle I would refactor usages, wdyt?

fvovan:
Yes, I will adjust it, and change all occurrences of its constructor

fvovan:
But what if we move this param in constructor from $options array to dedicated parameter.
Then we will be able to use type hinting, but not add  isset && isIstanceOf checkings?

tomaszdurka:
I actually thought about replacing string/array system with Phrase. Not adding extra option there. This would require usages adjustments.

What do you guys think about reworking constructor of exceptions? Make it more explicit, accept specific parameters instead of array-bag of them. Plus add some methods for setting additonal, optional values. 
I think we could play bit more with exceptions in certain places (like HttpResponse). It's ok to create exception, set couple of things and then throw it. We could still keep simple syntax for casual exceptions (no metaInfo).
```php
public function __construct($message, CM_I18n_Phrase $messagePublic = null, $severity = null);
public function setMessagePublic(CM_I18n_Phrase $messagePublic);
public function setMetaInfo(array $metaInfo);
public function setSeverity($severity);
```

Instead of:
```php
throw new CM_Exception_Invalid('Not found', ['request' => $this->_request], ['messagePublic' => 'Not found']);
```
do this:
```php
$exception = new CM_Exception_Invalid('Not found', new CM_I18n_Phrase('Not found'));
$exception->setMetaInfo(['request' => $this->_request]);
throw $exception;
```

tomaszdurka:
*Side note*

Any of you has experience with using `SplEnum`? Unfortunately this needs additional php extension.
```php
<?php

class CM_ExceptionSeverity extends SplEnum {

    const WARN = 1;
    const ERROR = 2;
    const FATAL = 3;
}
$exception =  new CM_Exception_Invalid($message);
$exception->setSeverity(new CM_ExceptionSeverity(CM_ExceptionSeverity::ERROR));
```
Probably less readible, but more strict and safe.

njam:
I would vote for keeping $severity *before* $messagePublic - I think it's used more often

tomaszdurka:
I don't mind putting `severity` before `messagePublic`. I agree it is used more frequently. My motivation was to have singature more human friendly.

As I prefer:
```php
new CM_Exception('Not found', new CM_I18n_Phrase('Not found'));
new CM_Exception('Not found', new CM_I18n_Phrase('Not found'), CM_Exception::WARN);
new CM_Exception('Not found', null, CM_Exception::WARN);
```
than
```php
new CM_Exception('Not found', null, new CM_I18n_Phrase('Not found'));
new CM_Exception('Not found', CM_Exception::WARN, new CM_I18n_Phrase('Not found'));
new CM_Exception('Not found', CM_Exception::WARN);
```
Especially cause 2 first use cases :panda_face:

fvovan:
As for me I would prefer grouping arguments together too.
But why we can't  pass those additional arguments ($metaInfo and $options, or  even only $metaInfo since $options will be divided) to constructor too?

fvovan:
should I change CM_Exception_FormFieldValidation (and possibly other childs) as well?

tomaszdurka:
Yes, `CM_Exception_FormFieldValidation` should probably accept `messagePublic` as `Phrase` only.

alexispeter:
I remember that we had explicit parameters before and added the array-bag when we felt the signature would become too crowded.

See https://github.com/cargomedia/cm/pull/1114

fvovan:
Ahh, that's why I found few places remained with old arguments passing.

tomaszdurka:
@alexispeter I agree that too crowded signatures are problematic. But keeping them short while adding depth ain't good solution too. It's very hard to determine what you can pass down there. That's why I suggest to move out public message and severity out of options array. Especially when messagePublic is single argument (Phrase) now.

Other thing is which arguments should be in constructor. Imo most important for exceptions are message and maybe severity. Other things like `messagePublic` and `metaInfo` are optional. Maybe we should remove both of these (also messagePublic) from constructor and have only setters for them.
I suggested to keep `messagePublic` in constructor as it's used quite often, as is severity. It turned out that `$metaInfo` is very rarely used.

alexispeter:
Agree on public message stuff, but meta-info is used in a lot of places.

tomaszdurka:
I agree it's used quite often. 
With this buggy regexp (`new CM_Excep[^(]+\([^,\n]+, [$a\[]`) I got 61 usages.
'messagePublic' found only 24 times.

@alexispeter You would keep both in constructor signature or move messagePublic out into setter?

I have nothing against:
```php
public function __construct($message, $severity = null, array $metaInfo = null);
public function setMessagePublic(CM_I18n_Phrase $messagePublic);
public function setMetaInfo(array $metaInfo);
public function setSeverity($severity);
```
On the other hand not always amount of usages should decide about signature.

alexispeter:
Honestly, I have nothing against a constructor signature that contains all possible settings for exceptions. Especially since we can replace `messsagePublic` and `messagePublicVariables` with `phrase`. Even without that, 5 arguments is not really that much.

tomaszdurka:
I thought for a while, why we are having this `$messagePublic` at all. If we would pass `Phrase` as `$message` we could simply not translate it and get proper non-public message. Then we could customize public messages with translations.
We would still need a way to set exception to be public (`setPublic($state)`).

Deeper topic. Wdyt?

fvovan:
I'am not sure if I should implement it before discuss would been finished, but I did.
All params are passed to constructor.
@tomaszdurka 
> we would pass Phrase as $message we could simply not translate it and get proper non-public
> message. Then we could customize public messages with translations.
> We would still need a way to set exception to be public (setPublic($state)).

I'am not sure. There are too many occurrences of new CM_Exception and its childs. And already in few places we use dynamic invocation (like ```new $exception()``` etc) and it is not covered with tests. We can overcomplicate it now.
And actually I think that setters should be used as little as possible especially if object is supposed to be immutable (like any Exception is)

alexispeter:
> I thought for a while, why we are having this `$messagePublic` at all. If we would pass `Phrase` as `$message` we could simply not translate it and get proper non-public message. Then we could customize public messages with translations.

So basically you would use an exception message `A`, which is meaningful to developers, as a language-key to get a sentence `B` which is meaningful to the user? I don't think this is a good idea, this could be very confusing for translators.

fvovan:
Will we make some decision about constructor signature? 
Then I could  adjust it, or, if it's okay, change sk part.
as a reminder currently it is 
```public function __construct($message = null, CM_I18n_Phrase $messagePublic = null, $severity = null, array $metaInfo = null)```

**Comment by njam** [comment]
And master has:
```php
CM_Exception:
 public function __construct($message = null, array $metaInfo = null, array $options = null)

CM_InputStream_InvalidValueException:
 public function __construct($message = null)

CM_Exception_FormFieldValidation:
 public function __construct($messagePublic, array $variables = null)

CM_Exception_Blocked:
 public function __construct(CM_Model_User $user)
```

I would still prefer `array $options` just because this API should *never* change (there are lots of usages). The array version is just easier to adapt to changing needs.
@fvovan wdyt?

fvovan:
I don't mind passing it in $options array, but it is too similar to our recent discussion about ```replace()``` in CM_Model_StorageDatabase. Tried to investigate how this problem is solved in popular frameworks, but they avoid passing any custom data in exceptions. I assume  that they figure out such kind situations just catching and logging it ASAP (and probably later rethrowing), so they keep all context information inside certain module. So this is not our case.
What if we move ```$messagePublic``` to $options array (as this data is not critical), but keep $severity in the constructor signature. Then I think it is worth to create some "$options validator".
``` public function __construct($message = null, $severity = null, array $metaInfo = null, array $options) ```
But I'am not sure @tomaszdurka  will like it

**Comment by njam** [comment]
I think it's good!
```php
public function __construct($message = null, $severity = null, array $metaInfo = null, array $options = null)
```

fvovan:
@njam please review.
I am not sure about ```_validateOptions()```. Something tells me that validator should be either as simplest as possible or  it should generalised and moved to separate class (or trait). Possibly we already use such kind validators (in schema or somewhere else).

njam:
- CM_Exception: How about instead of using a validator just using a setter?:
```php
if (isset($options['foo']) {
 $this->setFoo(isset($options['foo']);
}
```
(we would not throw on superfluous options anymore, but I think that would be okay?)

@alexispeter do you wanna have a look at this change as well? wdyt?

alexispeter:
Honestly I preferred the old solution, where "invalid" options are simply ignored.

fvovan:
Changed to non strict fashion.
One thing, I would make setMessagePublic setter private, because currently getMessagePublic actually returns already translated string but not CM_I18_Phrase instance.

njam:
- CM/I18n/Phrase.php: For consistency I would rather do a separate IF for the default value, à la:
```php
if (null === $foo) {
 $foo = 12;
}
```
In the sense of checking and adjusting input parameters early.

Rest looks good to me. I trust you checked all usages.

@alexispeter ok for you as well?
@fvovan could you prepare a PR in SK so that we can then quickly upgrade after this is merged?

alexispeter:
ok for me

fvovan:
it's here
https://github.com/cargomedia/sk/pull/1897/

njam:
- CM/I18n/Phrase.php: What I meant is to adjust input data first, then assign it, like:
```
if (null === $foo) {
 $foo = 12;
}
$this->_foo = $foo;
```
I think it's a bit more readable, especially if methods become longer.

PR #1854, opened 2015-07-22
Uncovered password input

christopheschwyzer:
...with _hide_ option. 

See https://youtu.be/nmKMz3Fg76M?t=14m55s

### DIFF
--- a/layout/default/Component/Example/tabs/forms.tpl
+++ b/layout/default/Component/Example/tabs/forms.tpl
@@ -1,5 +1,6 @@
 {form name="CM_Form_Example" viewer=$viewer}
 {formField name='text' label="{translate 'Text'}" placeholder="{translate 'Placeholder'}"}
+{formField name='password' label="{translate 'Password'}" placeholder="{translate 'Password'}"}
 {formField name='textarea' label="{translate 'Textarea'}" placeholder="{translate 'Placeholder'}"}
 {formField name='int' label="{translate 'Integer'}"}
 {formField name='location' label="{translate 'Location'}"}
--- a/layout/default/FormField/Password/default.less
+++ b/layout/default/FormField/Password/default.less
@@ -1,0 +1,17 @@
+& {
+  position: relative;
+}
+
+.mode-hidden {
+  display: none;
+}
+
+.textinput {
+  padding-right: @sizeButton;
+}
+
+.button {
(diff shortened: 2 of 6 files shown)


christopheschwyzer:
Current approach: 
Two input fields that you can toggle. This not be the best idea as we have two input fields with the same ID and I'm not sure if that can cause problems :/

njam:
We can't change the `type` of the input?

christopheschwyzer:
According to http://stackoverflow.com/a/1544338 it's not possible. Haven't verified it though.

njam:
This answer is about a jQuery restriction which they put in place because of IE.
I can't find this restriction in current jQuery. I imagine it works in IE9+?

christopheschwyzer:
@njam You're right. I should start to read through articles.

njam:
`togglePasswordMask` could probably be simplified into one code branch.
Maybe à la?:
```js
var showText = (type === 'password');
$input.attr('type', showText ? 'text' : 'password');
$buttonText.toggle(showText);
$buttonPassword.toggle(!showText);
```

christopheschwyzer:
Yes!

njam:
Code lgtm, haven't tested it.

Are we hiding the native UI for this in IE?
Should Nicolas do some testing?

christopheschwyzer:
Native UI is my biggest concern. I didn't dare testing it yet :grin:. Also I created a new icon that is not available yet https://github.com/cargomedia/cm/pull/1857.

christopheschwyzer:
Solved for IE!

christopheschwyzer:
One difference is that mobile phones use to start with an uppercase letter for textfelds (opposed to password fields that are lowercase by default)

njam:
http://stackoverflow.com/questions/5171764/how-do-you-turn-off-auto-capitalisation-in-html-form-fields-in-ios ?

christopheschwyzer:
yep. Should be all good now.

njam:
Cool! Code looks good.

Any more testing needed?

christopheschwyzer:
I think technically it's fine - not sure how users will embrace it ;)

njam:
@NicolasSchmutz you wanna test it and give your opinion?

NicolasSchmutz:
lvgtm!

PR #1858, opened 2015-07-23
Reduce timeout for Kickbox email validation request

fauvel:
They were just down for 10mins:
```
RuntimeException: [curl] 28: Operation timed out after 150218 milliseconds with 0 out of 0 bytes received [url] https://api.kickbox.io/v1/verify?email=example%40gmail.com in /home/example/releases/20150722121906/vendor/kickbox/kickbox/lib/Kickbox/HttpClient/HttpClient.php on line 119
```

@fauvel how about setting this timeout to 0.5 seconds or so?

cc @kris-lab @ppp0

### DIFF
--- a/library/CMService/KickBox/Client.php
+++ b/library/CMService/KickBox/Client.php
@@ -67,7 +67,7 @@ class CMService_KickBox_Client implements CM_Service_EmailVerification_ClientInt
      * @throws Exception
      */
     protected function _getResponse($email) {
-        $kickBox = new \Kickbox\Client($this->_getCode());
+        $kickBox = new \Kickbox\Client($this->_getCode(), [\Guzzle\Http\Client::CURL_OPTIONS => [CURLOPT_TIMEOUT => 7]]);
         return $kickBox->kickbox()->verify($email);
     }
 


fauvel:
@njam please review. Testing it locally, it turned out that a 5 seconds timeout would be too short – it is being reached from time to time even when KickBox is up and running. 6 seconds seems fine. Note that we already have a workaround in place for this performance issue: We are validating email addresses on-the-fly while users are filling out the sign up form, and we cache the result to validate quickly upon submit. Caching will **not** take place in case of a timeout, and that's bad for the user experience because the form submission will last longer. So it is better to have a long timeout and make sure that we load the validation response *before* the form is being submitted.

ppp0:
Any ways to implement this in a non-blocking way?

fauvel:
There are several options of course, we could for instance disable validation for 1 minute every time we have a timeout by putting a short-lived flag in the cache... This would avoid slowing down the sign up validation when KickBox is down. But is it really relevant? I mean, a 150s timeout was definitely too long, but 7 seconds are kind of ok (in the exceptional case of a downtime). Wdyt?

njam:
@fauvel 7 seconds, really?? So be it.
And if the timeout is reached we allow the email address?

@ppp0 validating the email address is required before signup, it's already happening asynchronously *before* hitting the button. What do you mean by doing it non-blocking?

ppp0:
@njam are you allowed to hit the button if the async answer is pending?

njam:
It will wait until the validation it done, because we don't want users to sign up with not-validated addresses.

@fauvel if the timeout is reached do we consider the email address to be valid?

ppp0:
Well, so then if the validation is pending and finally times out it blocks the process of signing up?

ppp0:
On the other hand, we allow users that enter a valid, but not their email address?

fauvel:
- The email form field validates the email address on-the-fly, asynchronously, before the user pushes the sign up button. As far as I know, this validation does not block the sign up form – you can submit it even while the form field validation is still pending.
- @njam Yeah, 6 seconds. I've tried it "live", too, by filling up the sign up form and finally copy-pasting an email address just before submitting it. It looks like email validation is the longest part of the process (that's why we validate on-the-fly and cache the result, so that it is already there when the form is submitted).
- If a timeout is reached, the email validation falls back to the standard one (pattern-based). But the KickBox validation result isn't being cached. This means that, when KickBox is down, the on-the-fly validation will pass after the given timeout (resp. defaults to the standard validation), and that a KickBox validation attempt will be made again when the form is being submitted. This of course delays the form processing by the duration of the timeout. But 6 seconds is fully ok imo.

njam:
lgtm then!

fauvel:
Let's have a look at the error logs afterwards then!

PR #1859, opened 2015-07-24
Add adapter pattern for logging

njam:
Two loggers so far:
- Database (CM_Log_Abstract) - Mongodb
- Sentry
- Plain file
- CM_OutputStream_Abstract

Note: currently `logException()` formats exception and sends text to logger:
```php
$log->add($formatter->formatException($exception), $metaInfo);
```

The Mongodb handler should accept a "TTL" and a "collection" configuration.
The TTL needs to be applied somewhere (setup script?).

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -18,7 +18,8 @@
     "kissmetrics/kissmetrics-php": "~0.1.0",
     "guzzlehttp/guzzle": "~4.1.7",
     "kickbox/kickbox": "~1.0.3",
-    "tomaszdurka/codegenerator": "~0.4.1"
+    "tomaszdurka/codegenerator": "~0.4.1",
+    "raven/raven": "~0.12.0"
   },
   "require-dev": {
     "phpunit/phpunit": "~4.3.5",
--- a/library/CM/Log/Context.php
+++ b/library/CM/Log/Context.php
@@ -1,8 +1,8 @@
 <?php
 
-class CM_Logging_Context {
+class CM_Log_Context {
 
-    /** @var CM_Logging_Context_ComputerInfo */
+    /** @var CM_Log_Context_ComputerInfo */
     private $_computerInfo;
 
     /** @var CM_Model_User|null */
--- a/library/CM/Log/Context/ComputerInfo.php
+++ b/library/CM/Log/Context/ComputerInfo.php
@@ -6,9 +6,6 @@ class CM_Log_Context_ComputerInfo {
     private $_fqdn;
 
     /** @var string|null */
-    private $_interpreterName;
-
-    /** @var string|null */
-    private $_interpreterVersion;
+    private $_phpVersion;
 
 }
--- a/library/CM/Log/Context/ComputerInfo.php
+++ b/library/CM/Log/Context/ComputerInfo.php
@@ -1,6 +1,6 @@
 <?php
 
-class CM_Logging_Context_ComputerInfo {
+class CM_Log_Context_ComputerInfo {
 
     /** @var string|null */
     private $_fqdn;
--- a/library/CM/Log/Handler/HandlerInterface.php
+++ b/library/CM/Log/Handler/HandlerInterface.php
@@ -1,6 +1,6 @@
 <?php
 
-class CM_Logging_Handler_HandlerInterface {
+class CM_Log_Handler_HandlerInterface {
 
     /** @var bool */
     private $_bubble;
@@ -12,10 +12,10 @@ class CM_Logging_Handler_HandlerInterface {
     }
 
     /**
-     * @param CM_Logging_Record $record
+     * @param CM_Log_Record $record
      * @return bool Whether the record was successfully handled
      */
-    public function handleRecord(CM_Logging_Record $record) {
+    public function handleRecord(CM_Log_Record $record) {
     }
 
 }
--- a/library/CM/Log/Logger.php
+++ b/library/CM/Log/Logger.php
@@ -1,22 +1,23 @@
 <?php
 
-class CM_Logging_Logger {
+class CM_Log_Logger {
 
     const DEBUG = 100;
     const INFO = 200;
     const NOTICE = 250;
     const WARNING = 300;
+    const ERROR = 400;
     const CRITICAL = 500;
     const ALERT = 550;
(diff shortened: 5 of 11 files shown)


njam:
I pushed a draft.
@vogdb wanna have a look? :)

vogdb:
looks ok for a draft.
The thing that I don't understand how are `CM_Logging_Logger::addException` and `CM_Logging_Logger::addMessage` will create a corresponding `CM_Logging_Context` for an instance of `CM_Logging_Record `. In case of Sentry this `CM_Logging_Context` is magically created by the lib. How can we set it manually? Do we need to set it manually?

njam:
Yup the `Context` part is a bit tricky. My proposition would be:
- For server side records we calculate the *context* from global state (ouch) or from some environment (App, Bootloader).
- For client side records (JS) we create a *context* based on what the AJAX requests sends to the server

fvovan:
But what stops us from using https://github.com/Seldaek/monolog/blob/master/src/Monolog/Handler/RavenHandler.php ?
@njam, I remember, on Friday you mentioned some interface incompatibility but here http://sentry.readthedocs.org/en/stable/developer/interfaces/index.html I can't find something special, that could not be done with Monolog processors.

njam:
As discussed in person:
- Nice thing about Monolog is that there are many *handlers* already written
- Problem about Monolog (and PSR-3) is that it's very vague (`context` is array or arbitrary information). So inter-handler unification is mostly nonexistent.
- Monolog Raven handler doesn't provide the functionality we want.
- Monolog core logic we need is pretty basic, we'll do it ourselves.

zazabe:
cm_logger_diagram

Please, take a look at this draft UML diagram:
- `CM_Log_Logger`: the log service, create and forward log records to log handlers
- `CM_Log_Record`: store the log level, the message and a context object   
   the exception has been removed at this level, logging is mainly about recording a message somewhere and an exception is more a specific case
- `CM_Log_Context`: store all the meta information available, it can contain global info about the application env but also specific info regarding the current log record
- `CM_Log_Handler_*`: Process log record according to the facility used to store it

The main change is about moving the exception into the context object.

fvovan:
As for me I'm not sure if is good idea to make all these PSR-3 helper methods.
At least for now. They look confusing because of existing too many entry points for Logger, and besides  presence ```error()``` and ```exception()``` methods is weird. First logs messages with ERROR level, but latter logs Exceptions (without certain level).

Probably for now it would be better to leave ```addException``` and ```addMessage``` as public and also make addRecord encapsulated.

As for moving Exception to context probably in our case it doesn't match well, because ```Context``` here is some environment information while we want Exception (or Message) to be a payload. And probably Exceptions are even more important for us AFAIU this issue was created to deal with them foremost and plain messages came later "as a bonus". 

Also I thought about 'bubbling' feature. Do we really need it? What could be it usage case? If  we set several handlers for certain logger we probably want it to log our Record to all of them? If we need only 1 handler we can instantiate separate Logger for it. Wdyt?

We should discuss what we want to pass in ```$options``` array. I assume we need a way to pass some context (for example in case of JS error), then inside ```addException()``` and ```addMessage()``` decide if we need to obtain more information and 'merge' contexts or use just passed or throw it away.

One more thing we should think about how to make certain (actually all non-critical) messages asynchronous. If we widely used it probably we don't want to block our application for a mongo insert and especially for external http request. Maybe it is good idea to create 2  instances of Raven_Client and dispatch noncritical data asynchronously (Raven can do it). MongoDB has async driver too (probably we already use it)

fvovan:
And maybe if we decide to implement Stream handler we don't need for File one (or at least it could be specific case of Stream, as it's done in Monolog)

zazabe:
@fvovan 

- Having some helpers at the Logger level seems very common to me... but I agree, we can definitely remove some levels.
- `exception` is a simple helper using a mapping between the exception severity and the log level, this method aims to be used by our exception handler and not directly... actually if we keep the exception in the context, we will be able to do `$logger->debug('my debug message', ['exception' => $e ]);` and handlers with exception support will be able to record then as any other caught exception.
- the context store all the meta information available about a log message, like the user (if any), the http request (if any), the exception (if any), etc...   
if we keep the exception in the log record, it will have no sense for all the "normal" log message...

But it's really depend on the propose of this refactoring:
1. adding an additional layer to our exceptionHandler system to easily support diff handlers
2. creating a logger system that follow as much as possible the PSR-3 recommendation and then using it for different propose: debug, info, error/exception

fvovan:
I understand what ```exception``` does, but really seems confusing when ```error``` and  ```exception`` are together but do non similar things. :)

>actually if we keep the exception in the context, we will be able to do $logger->debug('my debug message', ['exception' => $e ]); and handlers with exception support will be able to record then as any other caught exception.

what's advantages over just call addException()?
Obvious downside is we not be able to use type hinting.

If i'm not mistaken the primary purpose is Raven integration for exceptions logging, and there exception is 'a first class entity' and exception has some context, but not opposite.
Anyway it is not PSR-3, and if we just want PSR-3 it would be easier to take ready Monolog and customize Raven handler for it.

zazabe:
cm_logger_diagram_v2

- agreed that `exception` is maybe not the best name for this method
- an exception is not something contextual to a log record, agree on that, so instead we can maybe handle that by using an exception log record
- i removed some levels, they have no meaning for us and we can stay on exception severity as level (warning, error, critical) + debug/info
- about the `bubbling` concept, we need to define some kind of hierarchy for our handlers to get the bubbling order, maybe a priority level defined on each handler... but I also think that we can drop this feature and just define the level processed by each handler.

zazabe:
As discussed with @njam and @fvovan :

Bubbling we be automatically defined base on the logger service configuration, ie:
```
handlers => [A, B, C]
fallbacks => [D, E]
```
- all `handlers` will bubble, except the last one (`C`) 
- fallback handlers will not bubble
- if there's no `handlers`, available, fallback handlers will be used  
  useful during the bootstrapping, we can use a `basic` handler to log message early in the init process, then only use it as a fallback.

zazabe:
I did a new pull request: #1926 
@njam should we close this one?

njam:
Yup let's close. I adjusted the references.

PR #1866, opened 2015-07-28
Only uncover password on focus

christopheschwyzer:
Follow up https://github.com/cargomedia/cm/pull/1854

To make password managers (like 1Password) work.

### DIFF
--- a/library/CM/FormField/Password.js
+++ b/library/CM/FormField/Password.js
@@ -61,6 +61,6 @@ var CM_FormField_Password = CM_FormField_Text.extend({
    * @param {Boolean} state
    */
   _setInputTypeByState: function(state) {
-    this.getInput().attr('type', state ? 'text' : 'password');
+    this.getInput().attr('type', state ? 'text' : 'password').focus();
   }
 });
--- a/library/CM/FormField/Password.js
+++ b/library/CM/FormField/Password.js
@@ -20,9 +20,15 @@ var CM_FormField_Password = CM_FormField_Text.extend({
     this._applyDesiredPasswordVisibility();
 
     var self = this;
-    this.getForm().on('submit', function() {
-      self.togglePasswordVisibility(false);
-    });
+    var visibilityDesired;
+    this.getForm()
+      .on('submit', function() {
+        visibilityDesired = self._visibilityDesired;
+        self.togglePasswordVisibility(false);
+      })
+      .on('error', function() {
+        self.togglePasswordVisibility(visibilityDesired);
+      });
   },
 
   triggerChange: function() {
--- a/library/CM/FormField/Password.js
+++ b/library/CM/FormField/Password.js
@@ -17,9 +17,8 @@ var CM_FormField_Password = CM_FormField_Text.extend({
 
   ready: function() {
     this.enableTriggerChangeOnInput();
-    this.on('change', function() {
-      this._applyDesiredPasswordVisibility();
-    });
+    this._applyDesiredPasswordVisibility();
+
     var self = this;
     this.getForm().on('submit', function() {
       self.togglePasswordVisibility(false);
@@ -49,11 +48,7 @@ var CM_FormField_Password = CM_FormField_Text.extend({
   },
 
   _applyDesiredPasswordVisibility: function() {
-    if (0 === this.getValue().length) {
-      this._togglePasswordVisibilityInput(false);
-    } else {
-      this._togglePasswordVisibilityInput(this._visibilityDesired);
-    }
+    this._togglePasswordVisibilityInput(this._visibilityDesired);
   },
 
   /**
--- a/library/CM/FormField/Password.js
+++ b/library/CM/FormField/Password.js
@@ -7,6 +7,8 @@ var CM_FormField_Password = CM_FormField_Text.extend({
 
   _visibilityDesired: true,
 
+  _valueLast: '',
+
   events: {
     'click .togglePasswordVisibility': function() {
       this.togglePasswordVisibility();
@@ -22,6 +24,15 @@ var CM_FormField_Password = CM_FormField_Text.extend({
     this.getForm().on('submit', function() {
       self.togglePasswordVisibility(false);
     });
+  },
+
+  triggerChange: function() {
+    var valueCurrent = this.getInput().val();
+    //it means we that 'paste' event has occurred.
+    if (valueCurrent.length - this._valueLast.length > 1) {
+      this.togglePasswordVisibility(false);
+    }
+    CM_FormField_Text.prototype.triggerChange.apply(this, arguments);
   },
 
   /**
--- a/library/CM/FormField/Password.js
+++ b/library/CM/FormField/Password.js
@@ -5,7 +5,7 @@
 var CM_FormField_Password = CM_FormField_Text.extend({
   _class: 'CM_FormField_Password',
 
-  _visibilityDesired: true,
+  _visibilityDesired: false,
 
   _valueLast: '',
 
@@ -17,7 +17,7 @@ var CM_FormField_Password = CM_FormField_Text.extend({
 
   ready: function() {
     this.enableTriggerChangeOnInput();
-    this._applyDesiredPasswordVisibility();
+    this.togglePasswordVisibility(this._visibilityDesired);
 
     var self = this;
     var visibilityDesired;
--- a/library/CM/FormField/Password.js
+++ b/library/CM/FormField/Password.js
@@ -3,52 +3,52 @@
  * @extends CM_FormField_Text
  */
 var CM_FormField_Password = CM_FormField_Text.extend({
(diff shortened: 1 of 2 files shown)


christopheschwyzer:
@njam @vogdb maybe there is a better way to do this?!

njam:
- Let's not pass an `event` instance from the event handler to the function (as it makes the function unusable from other places where we don't have an event).
 - Rather I would put *some* logic into the `focus` event handler, à la `if (length === 0) { togglePasswordMask(); }`
- Don't we have the problem now that on every `focus` event we toggle the mode if there is some text?

christopheschwyzer:
done w/ @njam 
@vogdb wanna have a look?

njam:
@christopheschwyzer I think it would be good to hide the password on "paste" as you suggested.
-> https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/onpaste

vogdb:
looks ok.

christopheschwyzer:
I added 
```
      'paste input': function() {
            this.togglePasswordVisibility(false);
        }
```
Now when I paste some text from clipboard it works fine - when I use 1PW it doesn't switch to password. Any ideas?

vogdb:
It can actually type the password, not paste it =) It just does it real quick cause it's a program.

njam:
Oh, then that's not an option  :/
Either we ignore it, or we detect input speed?

vogdb:
For that we need to be sure that 1PW types it and not sets it directly smth like this: `input.value = value`. Anyway we don't want to depend on its inner workings. Detection of speed is not an option.

njam:
That would trigger our code as well (because we're using `propertychange` in `CM_FormField_Text.enableTriggerChangeOnInput()`).
We could detect that the text changes from 0 to 2 characters in less than X milliseconds, then switch to password mode?

vogdb:
Is it ok that now browsers' password managers do not offer to save password on our Login and SignUp forms?

njam:
Yes, I noticed that too :(
But I think with this pull request it should be fixed, as we're using `type="password"` again by default, can you confirm?

vogdb:
hm, didn't work for me. I don't see why it is suppose to work. When you start to type, input becomes of `type = "text"` and does not return back to `password` before submitting.

vogdb:
btw why we have a new formatting here? 4 spaces instead of 2? Did we change our default formatting?

vogdb:
I pushed here a small change to enable Password Manager. Verified it on the latest Chrome. Works.

christopheschwyzer:
So you change back to password on submit? Good idea!

vogdb:
Found an interesting bug on Android. The first entered symbol goes forward on the right side of the cursor. It's like this symbol was already in the input field before we had started to type.

Also have a question. Do we need to change input type to `text` in case of submit error? Now we change it to `password` before submit.

vogdb:
@christopheschwyzer @njam  please look at the change.

christopheschwyzer:
> Also have a question. Do we need to change input type to text in case of submit error? Now we change it to password before submit.

This will not be possible b/c error is only possible after submit?!

vogdb:
My last commit stores the status of `visibilityDesired` on `submit` and restores it on `error`.

christopheschwyzer:
Ah I see, smart. Didn't notice your commit. Sounds reasonable.

njam:
@tomaszdurka could you review this? Aleksei, maybe give Tomasz an intro on the topic?

tomaszdurka:
Isn't Aleksei on holidays this week? I am aware of the issue. 

*I am generally against uncovering password, but well, you guys probably got this topic covered right* :panda_face: Shouldn't this be only for registration?

library/CM/FormField/Password.js
```diff
@@ -5,18 +5,62 @@
...
+      this.togglePasswordVisibility(false);
+    }
+    CM_FormField_Text.prototype.triggerChange.apply(this, arguments);
+  },
+
+  /**
+   * @param {Boolean} [state]
+   */
+  togglePasswordVisibility: function(state) {
+    if ('undefined' === typeof state) {
```
tomaszdurka: There is underscore function for this too: `_.isUndefined()`

library/CM/FormField/Password.js
```diff
@@ -5,18 +5,62 @@
...
+  },
 
-    var showText = ($input.attr('type') === 'password');
-    $input.attr('type', showText ? 'text' : 'password');
-    $buttonText.toggle(showText);
-    $buttonPassword.toggle(!showText);
+  /**
+   * @param {Boolean} state
+   */
+  _togglePasswordVisibilityInput: function(state) {
```
tomaszdurka: Maybe call this `_setInputTypeByState`?

vogdb:
> Shouldn't this be only for registration?

It's hard for me to answer this question. I guess guys thought about it.

vogdb:
@tomaszdurka please review

christopheschwyzer:
@tomaszdurka Can you elaborate why you think it's not a good idea?

vogdb:
@christopheschwyzer lets finish this issue without Tomasz. It might take a while otherwise.

tomaszdurka:
@vogdb Let's say it's good.

@christopheschwyzer I could ask why you guys actually introduced this feature in the first place :smile_cat: Is this a trend on other platforms? 
Personally I feel that password is something very private (even for the non-important service you signed up for fun) and exposing it to anyone/anything except your mind feels unsecure. Especially in times when privacy is so important and it's lack might feel unsafe and dangerous. When I see password revealed - my mind somehow translates it to plain text passwords stored with no real security. 
Is this really only me thinking this way? :panda_face:

christopheschwyzer:
@tomaszdurka I agree on the first feeling that plain text is insecure. But then you still have the option to hide it if you're uncomfortable. It's a very rare situation that you have somebody looking over your shoulder while you type in your password. From UX perspective I'd say you help a lot of people from misspelling the pw, especially on signup, where it is crucial. About the _trend_ - I see more and more  "Reveal password" buttons, and I think it's only the first step. You can think of us as trendsetters in that regard. ;)

vogdb:
@christopheschwyzer Are you ok with the changes? Mergeable?

christopheschwyzer:
Doing some final testing right now.

christopheschwyzer:
@vogdb Works fine to input the pw. I noticed that 1PW doesn't ask you to store the pw the first time you input it (b/c it's not a pw input). Can we change it to be a pw type before/while we submit?

vogdb:
hm, we do it before submit in this PR.

christopheschwyzer:
True. So we change it too late somehow? Is there another event 1PW might use?

tomaszdurka:
Let's cover by default - solves all the problems :panda_face:

vogdb:
> So we change it too late somehow?

That seems to be the truth. If you hide password field then 1PW works fine. It does not matter when you do it. After or before password is filled. The only thing that matters is that it has `password` type before submit =(.

vogdb:
Changing it to `password` by default will not fix the problem completely cause when user changes the type to `text` before submit 1PW won't work either.

NicolasSchmutz:
TBH I would not bother too much about 1PW (although I love it too). 
I don't think our users use such a tool. I see more of an improvement for mobile users if it is not covered by default...

vogdb:
We can divide this behavior and make it uncover for small displays and cover for big displays. Nevertheless the 1PW problem exists.

vogdb:
Hm, now Chrome password manager does not work anymore. Firefox still works. I'm afraid that this feature makes unstable usage of password managers. And I think no one will argue that a big portion of users relies on password managers.

vogdb:
If password field was rendered from the beginning as the password and we don't change it's type in `ready` of CM_FormField_Password then browsers managers work. 1PW still does not. So, I'll propose to change the default visibility to hidden cause it fixes password managers problems.

vogdb:
@NicolasSchmutz @christopheschwyzer I need your opinions here.

christopheschwyzer:
I see the benefits don't make up for the caused problems :/ Let's switch the behavior after all. Password by default. Possibility to uncover. Switch to PW on submit and ignoring that it won't work 100% on all Browsers. @tomaszdurka @vogdb @NicolasSchmutz agree?

NicolasSchmutz:
sgmt

vogdb:

christopheschwyzer:
lgtm. @tomaszdurka please verify

vogdb:
@tomaszdurka please purify

tomaszdurka:
Sounds looks good.

christopheschwyzer:
Small details. Maybe we should re-focus the field after changing the type. @vogdb wdyt?

vogdb:
yep, that will be good

vogdb:

christopheschwyzer:
lgtm!

PR #1869, opened 2015-07-30
ContentPlaceholder stretch on Android devices.

vogdb:


vogdb:
somehow setting `max-height: 100%` on `.contentPlaceholder-size + *` does not work =( but that seems to be the problem.

christopheschwyzer:
closed as not necessary.

Issue #1652, opened 2015-02-21
Version 2.0 roadmap


tomaszdurka:
### Replace all global singletons with services
- [x] Convert CM_Memcache_Client into a service #1654
- [x] Convert CM_Elasticsearch_Client into a service #1653
- [ ] Convert Caches into services #1655
- [x] Convert CM_Adprovider into a service #1658
- [ ] Convert Gearman into JobManager service #1659
- [x] Convert CM_Option into a service #1661
- [x] Convert CMService_Newrelic into a service #1663
- [x] Convert CM_Debug into a service #1664
- [x] Convert CM_Stream_Video into a service #1665
- [x] Convert CM_Stream_Message into a service #1666
- [ ] Convert CM_Render into a service
- [ ] Convert Mail Delivery into a service
- [ ] Introduce "CodeRepository" service
- [ ] Introduce "ModelRepository" service
- [x] Introduce CM_Util_NamedArgs for easier services configuration #1662
### Update scripts
- [ ] convert update script to classes - use ProvisionScript class (_consider naming udpates scripts, consider suffixing numbered update scripts_) - see https://github.com/cargomedia/CM/issues/588 and #1870
### Remove global state

Replace singleton accessors with dependency injection.
- [ ] Remove global constants
- [ ] Remove global service manager
  - [ ] Store service manager on cli manager, pass manager to commands / runnables
  - [ ] Remove CM_ServiceManager singleton usages _(test cases, CM_File, Update scripts, etc)_
- [ ] Move boostraping from CM_Bootloader to CM_Application
  _Consider making CM_Application an abstract class (make CommandManager and HttpHandler its children)._
### Helpers
- introduce `CM_Schema_Definition`, `CM_Schema_Field`
- make `CM_Model_Schema_Definition` extend `CM_Schema_Definition`
- add `CM_Model_Schema_Definition::getSchema()`
- add `CM_Form_Abstract->registerFieldsFromSchemaDefinition(CM_Schema_Definition $schema)`
### New project creation notes
- upgrade web entry point
- nginx configuration doesn't support `modules/**/*.js` files loading
- add creation of the layout
- separate creation of the web application and cli application?
- create tables, collections from schemas instead of sql/json files
- improve creation of schema 
  - atm you need to know all the types to create proper schema - more oop?
  - introduce `Schema_Field`?
- add `CM_PagingSource_Model`
- convert CM_FormAction to CM_Form `action*` method
# 

This is a proposal of major framework version release. 
[...]

PR #1870, opened 2015-07-31
Simplify procedure of running "online" update-scripts

alexispeter:
The goal is to simplify app-deployment and subsequent running of "online-scripts".
Currently we add a `return;` statement to the beginning of update-scripts to skip them during deployment and manually edit and run the file on the server to run it.

New procedure is to return a `CM_Provision_UpdateScript` in the update-script with blocking set to true, which will cause the regular update-routine called with `bin/cm db run-udpates` to skip the script.
The script can then be run using `bin/dm db run-update <version>` after deployment has finished.

##### Example
```php
<?php

$script = new CM_Provision_UpdateScript(function () {
    echo "Running blocking updatescript\n";
}, true);

return $script;
```

please have a look @njam

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -22,7 +22,7 @@
   },
   "require-dev": {
     "phpunit/phpunit": "~4.3.5",
-    "tomaszdurka/mocka": "~0.10.0"
+    "tomaszdurka/mocka": "~0.11.0"
   },
   "autoload": {
     "psr-0": {
--- a/library/CM/App.php
+++ b/library/CM/App.php
@@ -107,11 +107,13 @@ class CM_App implements CM_Service_ManagerAwareInterface {
     }
 
     /**
+     * @param boolean|null $skipBlocking
      * @param Closure|null $callbackBefore fn($version)
-     * @param Closure|null $callbackAfter fn($version)
+     * @param Closure|null $callbackAfter  fn($version)
      * @return int Number of version bumps
      */
-    public function runUpdateScripts(Closure $callbackBefore = null, Closure $callbackAfter = null) {
+    public function runUpdateScripts($skipBlocking = null, Closure $callbackBefore = null, Closure $callbackAfter = null) {
+        $skipBlocking = (boolean) $skipBlocking;
         CM_Cache_Shared::getInstance()->flush();
         CM_Cache_Local::getInstance()->flush();
         $versionBumps = 0;
@@ -119,7 +121,7 @@ class CM_App implements CM_Service_ManagerAwareInterface {
             $version = $versionStart = $this->getVersion($namespace);
             while (true) {
                 $version++;
-                if (!$this->runUpdateScript($namespace, $version, $callbackBefore, $callbackAfter)) {
+                if (!$this->runUpdateScript($namespace, $version, $skipBlocking, $callbackBefore, $callbackAfter)) {
                     $version--;
                     break;
                 }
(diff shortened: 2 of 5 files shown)


njam:
This is related to https://github.com/cargomedia/cm/issues/1652 ("Update scripts").

@tomaszdurka can you look into this with Alexis?

tomaszdurka:
I like the idea, as you see of #1652 I was thinking about this a bit before.
Your implementation is very simple and allows usage of current system without any problem. What I would see different though are the non-blocking imlementation and update script classes.

First thoughts about update scripts as classes came when we did provision scripts. I would actually re-use that concept (extend?) with `load`, `unload`, `isLoaded` interface. This would solve problems of checking if specific script has been loaded (with non numeric names), plus would allow rollback in some cases.

I can see now the blocking flag is more like `runBeforeRelease`. With this approach we could probably have 2 cli commands for running before-release-script and after-release-scripts and orchestrate this outside in pulsar recipes. I would try to avoid that. Imo we have complicated update scripts enough already. 
What about using other tools we have already implemented - like gearman jobs. This could work for running non-blocking tasks. Update script would simply queue a code into gearman which would be picked up after release (once job-manager is started again).

@alexispeter @njam wdyt?

alexispeter:
I am a bit unsure about reusing the provision-scripts concepts.
I generally like the idea of relying on isLoaded() instead of app-version to tell if a script has run or not, but there are a lot of cases where that wouldn't be possible. In those cases we would rely on option-based scripts, which I highly discourage as it would clutter up our internal config for one-time scripts.
On alternative would be to introduce another type of option-based solution which does not rely on using types, which would rely on namespace and numeric name.

Regarding automated execution of after-release scripts, I would prefer your suggestion to use a separate cli-command run by pulsar. Using gearman is not practical, since a lot of after-release scripts output progress- and other information which the person responsible for the release should be able to see.

tomaszdurka:
Just to be sure - I meant to base implementation on `CM_Provision_Script_Abstract` and CM_Provision_Script_UnloadableInterface`, not on `CM_Provision_Script_OptionBased` one.
Not sure about internals (implementation of the abstract methods) yet.

tomaszdurka:
Regarding the output I think we can output command into some file/db log (per script maybe?).

I can see another problem with changes which e.g. require percona toolkit opertaions before. These need to be run on db machine and there is no way to do that from application.

alexispeter:
> I can see another problem with changes which e.g. require percona toolkit operations before. These need to be run on db machine and there is no way to do that from application.

That is just something we have to live with. 

> First thoughts about update scripts as classes came when we did provision scripts. I would actually re-use that concept (extend?) with load, unload, isLoaded interface. This would solve problems of checking if specific script has been loaded (with non numeric names), plus would allow rollback in some cases.

So you would discontinue using numerical names and always check for every update-script if it should be run or not? What about the order of execution? Usually order is probably not a problem when doing a regular release, but there could definitely be problems when doing a release that involves a lot of update-scripts that somehow depend on one another.

alexispeter:
Regarding output of gearman-workers. We would need either a remote file, since you don't know which worker-machine will pick the job up, or use the db. I dislike both.

tomaszdurka:
Ok I agree that gearman-workers might be a bit complicated in terms of output and monitoring for now. Additionally sometimes we want to run extra manual tasks before we run the script (e.g. percona-toolkit on databsase machines). This way it would not be possible.

------------------------------------------

With current implementation you use single class and instantiate it for each update script. You still determine if it was run based on the script name (and database). Old, more procedural update scripts are still acceptable.

How about creating class for each update script and use proposed before `CM_Provision_Script_Abstract` as parent class of it? It would require a little change of the run-update-scripts mechanism and do not accept normal file scripts anymore - which imo is something we need to do soon anyway. Each script would be class implementing `load` `shouldBeLoaded` and possibly `unload` for rollback.

-------------------------------------------------------

Regarding the implementation of ignoring some scripts (running later) I think it's our specific feature which doesn't need to be implemented with such explicit feature named "blocking".
Other approach (with extending `CM_Provision_Script_Abstract`) would use (ideal to me) `shouldBeLoaded` method to determine was the script already loaded. We could also easily override this method to not run some of the scripts.

alexispeter:
I just think that one class per update-script is overkill. Couldn't we achieve the same with just one single class?

Also, I think there could be quite a few problems with the order of update-scripts (big releases with a lot of inter-dependent scripts) and old scripts whose shouldBeLoaded suddenly becomes true again by the actions of some later update-script.

I think it's important to consider the differences between setup-scripts and update-scripts when defining the interface. Setup-scripts can always assume the same initial state (the data that the script is going to generate does not exist yet), whereas   update-scripts can operate on existing data.

tomaszdurka:
I think having a class is not that bad as you assume. 
With single class implementation you need to instantiate it in the script and return it. It uses not so clean behaviour of require which I would love to avoid (remove). Also the update script is heavily identified by the filename making it part for the script. E.g. in order to load, run, test update script from outside you would need to use update-script cli mechanism to search and load it.

Class per update script would use psr standards for autoloading so there would be no magic in connecting update script with actual filename.

--------

Regarding the `shouldBeLoaded` suddenly becoming true again. 
I understand that is in case when e.g. `shouldBeLoaded` actually checks if some table has a column, if yes it removes it? Yes that could be problematic. I would use very similiar mechanism as we do now. 

Or possibly (probably better) we could use this `shouldBeLoaded` for this kind of checks (if column exists) and additional simple one (list containing loaded scripts) on top of it in update-scripts-loaded (cli).

PR #1871, opened 2015-08-04
Empty translation keys in admin

fvovan:
Probably using CM_I18n_Prase in CM_Exceptions leads to adding empty translation keys to database table. Need to add additional checkings to its constructor

### DIFF
--- a/library/CM/Component/Example.php
+++ b/library/CM/Component/Example.php
@@ -42,9 +42,11 @@ class CM_Component_Example extends CM_Component_Abstract {
         if (!in_array($exception, array('CM_Exception', 'CM_Exception_AuthRequired'), true)) {
             $exception = 'CM_Exception';
         }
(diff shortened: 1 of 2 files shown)


fvovan:
@alexispeter , please review

alexispeter:
* Do a comparison with empty string. Using `empty()` will also return true if $phrase is `0`.
* use CM_Exception_Invalid rather than CM_Exception_InvalidParam. InvalidParam is reserved for exceptions relating to parameters supplied by the client and lead to a 404 http response.

NicolasSchmutz:
@fvovan is this released? I can still see the empty key...

fvovan:
Not yet. will make new CM tag and following sk PR now

alexispeter:
The empty key is still here, but the bug that lead to it is fixed.

NicolasSchmutz:
ok thanks!

PR #1872, opened 2015-08-08
Set minimum-scale=1

njam:
This will enable GPU rasterization in Chrome
See https://www.chromium.org/developers/design-documents/chromium-graphics/how-to-get-gpu-rasterization

@christopheschwyzer or @vogdb could you try how it affects the rendering behaviour of our feed page? I tried on a Samsung S6 and am not sure if it's really better..

### DIFF
--- a/layout/default/Layout/Abstract/default.tpl
+++ b/layout/default/Layout/Abstract/default.tpl
@@ -6,7 +6,7 @@
     <meta http-equiv="X-UA-Compatible" content="IE=edge; requiresActiveX=true">
     {if isset($pageDescription)}<meta name="description" content="{$pageDescription|escape}">{/if}
     {if isset($pageKeywords)}<meta name="keywords" content="{$pageKeywords|escape}">{/if}
-    <meta name="viewport" content="width=device-width, initial-scale=1, maximum-scale=1, user-scalable=no, minimal-ui">
+    <meta name="viewport" content="width=device-width, initial-scale=1, maximum-scale=1, minimum-scale=1, user-scalable=no, minimal-ui">
     <meta name="apple-mobile-web-app-capable" content="yes">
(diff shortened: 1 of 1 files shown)


vogdb:
S6 device and my Sony M2 have GPU rasterisation enabled by default, so they don't need this `meta` to work.

S6


Sony M2

vogdb:
And I didn't see any difference between the one with enabling settings and the opposite.

christopheschwyzer:
screen shot 2015-08-10 at 15 19 37
@vogdb Can you check with enabled Monitor. GPU Raster should be green (on)? I think by default it isn't without `minimal-scale` set.

vogdb:
You are right it's disabled on Desktop but I was talking about mobile devices cause I thought that our primary concern are them. Should we consider also Desktop?

christopheschwyzer:
@vogdb oh sorry, all good!

vogdb:
Hm but what about Desktop. Do we want to enable it there?

christopheschwyzer:
I would add the meta tag like suggested in this PR. Doesn't do harm either way.

njam:
Currently it only has an effect on Chrome for Android. You can check with the FPS flag for example.

Ok let's do this then.

vogdb:
True. This `meta` does not enable GPU rasterization on Desktop. The interesting feature about this rasterization is that when I've forced it on Desktop, it had started to devour memory as hell on scroll down.

Forced GPU on Desktop
screen shot 2015-08-10 at 21 54 55


Default GPU
screen shot 2015-08-10 at 21 56 53

njam:
Note this is only *GPU* memory. Probably if it's disabled it goes to different memory.

vogdb:
Ok!

PR #1873, opened 2015-08-12
CM_FormField_Site could have ambiguous labels

dlondero:
In `CM_FormField_Site::_initialize()` we use `$site->getName()` to set labels for the options. It could happen that two sites could have the same name set in the configuration and this would lead to two options with the same label.

Would be good to use `$site->getHost()` instead.

/cc @njam @fvovan


dlondero:
@fauvel please review

fauvel:
lgtm

PR #1875, opened 2015-08-13
Add CM_Exception_Invalid to the front-end

fauvel:


fauvel:
@tomaszdurka please review

njam:
Copying @fauvel
> We ran into a front-end error instead (Uncaught CM_Exception: No such exception CM_Exception_Invalid)

This comes from `CM_Exception.factory()` which is called from `CM_App.ajax()`:
```js
        if (response.error) {
          reject(new (CM_Exception.factory(response.error.type))(response.error.msg, response.error.isPublic));
        } else {
          resolve(response.success);
        }
```

Basically this means that any error that we "forward" to the client (as configured with  `exceptionsToCatch` in the config files) needs to be available on the client side as well? @vogdb can you confirm? How should we handle this? Create all JS files? Or use some automagic mechanism to create instances on the fly?

vogdb:
> `exceptionsToCatch` in the config files) needs to be available on the client side as well? @vogdb can you confirm?

I confirm.

> How should we handle this? Create all JS files? Or use some automagic mechanism to create instances on the fly?

I would vote for automagic mechanism ^____^ and removal of already configured exceptions from the client side but let me check whether this is possible.

vogdb:
Yep. We can do magic here. Who is up for magic?

njam:
sgtm! Would still be great if we *could* implement specific child classes if desired (for those cases where extra logic is needed).

vogdb:
ok, will count those in too.

fauvel:
Should I merge this as a quick fix, though?

vogdb:
hmm, no need no need

fauvel:
Well I need it right now... for the next release (within 1 hour)?

njam:
agree we can merge this. @vogdb can you open a followup?

vogdb:
noooo

fauvel:
@vogdb please delete all existing JS class declarations once you have implemented the "magic"!

vogdb:
Why you have done this? I've already prepared a fix here.

vogdb:
I did write here to wait a little bit

vogdb:
=(

vogdb:
='(

njam:
sorry :no_mouth:

fauvel:
oh I'm sorry too, I didn't think you were so quick... Underestimated your magic powers :)

PR #1868, opened 2015-07-28
Fix scrollShadow issue on iOS Safari

christopheschwyzer:
Issue with `position: absolute;` and  `webkit-overflow-scrolling: touch;`
Demo, if you somebody wants to do a bug report. http://jsfiddle.net/vfz1t4tj/4/

PR #1878, opened 2015-08-14
Fix scrollShadow issue on iOS Safari II

christopheschwyzer:
follow up https://github.com/cargomedia/cm/pull/1868
As discussed, let's re-add `-webkit-overflow-scrolling: touch;` but hide the _shadows_ on iOS.

### DIFF
--- a/client-vendor/after-body/jquery.scrollShadow/jquery.scrollShadow.js
+++ b/client-vendor/after-body/jquery.scrollShadow/jquery.scrollShadow.js
@@ -25,11 +25,10 @@
         this.updateShadow();
         return;
       }
-      var self = this;
 
       var cssClass = 'scrollShadow';
 
-      if (this.iOS) {
+      if (iOS) {
         // Fix for iOS Safari: Absolute positioned elements in combination with -webkit-overflow-scrolling: touch; Demo: http://jsfiddle.net/vfz1t4tj/4/
         cssClass += ' noShadows';
       }
@@ -37,6 +36,7 @@
       this.$element.addClass(cssClass);
       this.$element.wrap('<div class="scrollShadow-wrapper"></div>');
 
+      var self = this;
       this.$element.on('scroll.scrollShadow', _.throttle(function() {
         self.updateShadow();
       }, 200));
--- a/client-vendor/after-body/jquery.scrollShadow/jquery.scrollShadow.js
+++ b/client-vendor/after-body/jquery.scrollShadow/jquery.scrollShadow.js
@@ -30,6 +30,7 @@
       var cssClass = 'scrollShadow';
 
       if (this.iOS) {
+        // Fix for iOS Safari: Absolute positioned elements in combination with -webkit-overflow-scrolling: touch; Demo: http://jsfiddle.net/vfz1t4tj/4/
         cssClass += ' noShadows';
       }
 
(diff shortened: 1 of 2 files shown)


christopheschwyzer:
@vogdb @njam not sure this is the right way to do it. Please advise.

njam:
Also add a comment describing why we're doing that, and maybe link to ticket/fiddle?

vogdb:
Hm. Are we trying to keep the layout of `scrollShadow`? Why we don't do smth like that:

```js
  $.fn.scrollShadow = function(action) {
    var iOS = !navigator.userAgent.match(/(iPad|iPhone|iPod)/i);
    if(iOS) {
      return this; // No scroll shadow at all!
    }
    ...
```

If we want to keep the layout of `scrollShadow` then I can't think of a better solution than yours.

client-vendor/after-body/jquery.scrollShadow/jquery.scrollShadow.js
```diff
@@ -24,7 +27,14 @@
       }
       var self = this;
 
-      this.$element.addClass('scrollShadow');
+      var cssClass = 'scrollShadow';
+
+      if (this.iOS) {
```
vogdb: Should it be just `iOS` ?

client-vendor/after-body/jquery.scrollShadow/jquery.scrollShadow.js
```diff
@@ -24,7 +27,14 @@
       }
       var self = this;
```
vogdb: Would be better to move `self` declaration closer to the place where it's used.

njam:
Do we even need to initialize the whole plugin in that case, or could we just abort execution in `init()` if the device is not supported?

christopheschwyzer:
Problem is that we rely on the specific markup in certain contexts (extra `div`, css classes that are targeted). I agree, it would be nicer to not initialise the plugin at all, but I don't think it's possible with some of the current use cases.

vogdb:
That's what I suspected. Then lgtm.

njam:
makes sense

client-vendor/after-body/jquery.scrollShadow/jquery.scrollShadow.js
christopheschwyzer: you mean instead of `this.iOS`?

client-vendor/after-body/jquery.scrollShadow/jquery.scrollShadow.js
christopheschwyzer: where? just before `this.$element.on…` ?

client-vendor/after-body/jquery.scrollShadow/jquery.scrollShadow.js
vogdb: Yep, cause you have declared it as `var iOS` above.

client-vendor/after-body/jquery.scrollShadow/jquery.scrollShadow.js
vogdb: yep

christopheschwyzer:
@vogdb please verfiy

vogdb:
Brilliant! lgtm

PR #1889, opened 2015-08-20
Enable custom usertext-instance in usertext smarty plugin

alexispeter:
- Adds an extensible base for list-based usertext filters with wildcard capabillities, uses logic that was previously implemented on the badwords-paging
- Allows usage of pre-configured usertext instances in usertext smarty plugin for special filter usage


alexispeter:
please review @fauvel

This approach now only applies the filter once, as opposed to applying the filter multiple times, which would not work with the spamword-highlighting filter.
As a result, the badword-filter will no longer replace badwords that are created by application of the filter itself (see changes to the unit-test), which is probably not necessary since its highly unlikely that this would ever happen.

library/CM/Usertext/Filter/ListReplace.php
```diff
@@ -0,0 +1,119 @@
...
+
+    /**
+     * @param $text
+     * @return string
+     */
+    public function replaceMatch($text) {
+        $text = (string) $text;
+        $regex = $this->_toRegex();
+        $count = preg_match_all($regex, $text);
+        if ($count) {
```
fauvel: Why this first step?

library/CM/Usertext/Filter/ListReplace.php
```diff
@@ -0,0 +1,119 @@
...
+    }
+
+    /**
+     * @param $text
+     * @return string
+     */
+    public function replaceMatch($text) {
+        $text = (string) $text;
+        $regex = $this->_toRegex();
+        $count = preg_match_all($regex, $text);
```
fauvel: Why this first step?

library/CM/Usertext/Filter/ListReplace.php
```diff
@@ -0,0 +1,119 @@
...
+    /**
+     * @param string $phrase
+     * @return string
+     */
+    protected function _transformItemToRegex($phrase) {
+        $regexp = preg_quote($phrase, '#');
+        $regexp = str_replace('\*', '[^A-Za-z]*', $regexp);
+        $regexp = str_replace('\|', '\b', $regexp);
+        $regexp = '\S*' . $regexp . '\S*';
+        return $regexp;
```
fauvel: I would add the delimiters & flags `#...#i` right here (they belong to the regex)

library/CM/Usertext/Filter/ListReplace.php
```diff
@@ -0,0 +1,119 @@
...
+    }
+
+    /**
+     * @return string
+     */
+    private function _toRegex() {
+        $cacheKey = $this->_getCacheKey();
+        $cache = CM_Cache_Shared::getInstance();
+        $phrasesRegex = null;
+        if (false == ($phrasesRegex = $cache->get($cacheKey))) {
```
fauvel: Use `===`

library/CM/Usertext/Filter/ListReplace.php
```diff
@@ -0,0 +1,119 @@
...
+        return CM_CacheConst::Usertext_Filter_ReplaceList . '_phrasesHash:' . md5(implode('', iterator_to_array($this->_phrases)));
+    }
+
+    /**
+     * @return string
+     */
+    private function _toRegex() {
+        $cacheKey = $this->_getCacheKey();
+        $cache = CM_Cache_Shared::getInstance();
+        $phrasesRegex = null;
```
fauvel: You don't need to initialize here

library/CM/Usertext/Filter/ListReplace.php
alexispeter: code from an earlier approach I forgot to take out

library/CM/Usertext/Filter/ListReplace.php
```diff
@@ -0,0 +1,119 @@
...
+
+        return false;
+    }
+
+    /**
+     * @param string $userInput
+     * @return bool
+     */
+    public function isMatch($userInput) {
+        if (preg_match($this->_toRegex(), (string) $userInput)) {
```
fauvel: What about `return (bool) preg_match()`?

tests/library/CM/Usertext/Filter/BadwordsTest.php
```diff
@@ -2,61 +2,21 @@
...
+        $this->assertSame('… bar', $filter->transform('foo bar', new CM_Frontend_Render()));
+    }
 
-        $actual = $filter->transform("hello foobar there", $render);
-        $this->assertSame("hello ${replace} there", $actual);
-        $actual = $filter->transform("hello XfoobarX there", $render);
-        $this->assertSame("hello ${replace} there", $actual);
-        $actual = $filter->transform("hello mayo.foobar.ran there", $render);
-        $this->assertSame("hello ${replace} there", $actual);
+    public function testTransformCustomtReplacement() {
```
fauvel: Typo (`testTransformCustomReplacement `)

library/CM/Usertext/Filter/ListReplace.php
```diff
@@ -0,0 +1,111 @@
...
+     */
+    public function isMatch($userInput) {
+        return (boolean) preg_match($this->_toRegex(), (string) $userInput);
+    }
+
+    /**
+     * @return string
+     */
+    private function _getCacheKey() {
+        return CM_CacheConst::Usertext_Filter_ReplaceList . '_phrasesHash:' . md5(implode('', iterator_to_array($this->_phrases)));
```
fauvel: Is that caching!? Computing the cache key is almost as expensive as computing the cached value...

alexispeter: Is caching even needed?

fauvel: Caching the source paging for sure.

alexispeter: Did a quick benchmark to be sure, calculating the regex takes much less time than the actual cache-call.

library/CM/Usertext/Filter/Badwords.php
```diff
@@ -1,12 +1,12 @@
...
-    public function transform($text, CM_Frontend_Render $render) {
-        $text = (string) $text;
-        $badwordList = new CM_Paging_ContentList_Badwords;
-        $text = $badwordList->replaceMatch($text, '…');
-
-        return $text;
+    /**
+     * @param string|null $replace
+     */
+    public function __construct($replace = null) {
```
fauvel: Maybe rename to `$replacementPattern` for clarity?

fauvel:
lgtm!

PR #1863, opened 2015-08-19
Redesign FormFields

christopheschwyzer:
see: https://github.com/cargomedia/sk/issues/1902

PR #1886, opened 2015-08-20
Extend FormFields: Date 

christopheschwyzer:
- Input masks for date input field

Some solutions:
- https://github.com/bguzmanrio/maskjs
- https://github.com/igorescobar/jQuery-Mask-Plugin
- https://github.com/RobinHerbots/jquery.inputmask (seems to be the best one)

We also need to adjust based on location https://en.wikipedia.org/wiki/Date_format_by_country

### DIFF
--- a/client-vendor/after-body/jquery.inputmask/1-inputmask.js
+++ b/client-vendor/after-body/jquery.inputmask/1-inputmask.js
@@ -1,0 +1,2638 @@
+/*
+ * Input Mask Core
+ * http://github.com/RobinHerbots/jquery.inputmask
+ * Copyright (c) 2010 -	Robin Herbots
+ * Licensed under the MIT license (http://www.opensource.org/licenses/mit-license.php)
+ * Version: 0.0.0-dev
+ */
+
+(function($) {
+	function Inputmask(options) {
+		this.el = undefined;
+		//init options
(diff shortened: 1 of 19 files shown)


christopheschwyzer:
changes from https://github.com/cargomedia/cm/pull/1863 included here.

@njam I think it would make the input faster. But it also has some downsides: We need to adjust the mask depending on the location.

njam:
Tried it here briefly: http://robinherbots.github.io/jquery.inputmask/

I'm not a big fan of this kind of input personally, also because I have the feeling it's misusing a lot of HTML features to create new features, which is usually error-prone. *But* I see that it's probably faster for typing a birth date!

Lots of opinions: http://stackoverflow.com/questions/339956/whats-the-best-ui-for-entering-date-of-birth
Not sure, not sure :)

christopheschwyzer:
I see your point. 
Maybe a light version would be good compromise: 3 separate input fields. (Auto focus the next field after input) Could be more convenient to input especially when you trigger "number" keyboard on mobiles.

But maybe it's not worth the effort. Not sure - I see that dropdowns are are bullet-proof the most obvious.

christopheschwyzer:
I will close this, as no perfect solution in sight and status quo is fine.

PR #1892, opened 2015-08-25
Fix link parsing in CM_Usertext_Markdown

alexispeter:
Incorrect link markdown causes errors à la `ErrorException: Noticed exception 'ErrorException' with message 'E_NOTICE: Undefined offset: 3'`
Caused by copy-paste error in _doAnchors_reference_callback() which called the wrong parent-method.

please review @tomaszdurka 
cc @njam


tomaszdurka:
Ok then.

PR #1893, opened 2015-08-26
Fix promise throttler queue option.

vogdb:


vogdb:
@tomaszdurka please review.

vogdb:
The problem was that `.finally` didn't actually return the result of `fn.apply(null, args)` cause `.finally` does not modify the returned result, i.e. its result is ignored. See https://github.com/petkaantonov/bluebird/blob/master/API.md#finallyfunction-handler---promise

tomaszdurka:
Looks good!

PR #1894, opened 2015-09-02
Allow/set different timeout for creating ES indexes

fvovan:


fvovan:
I think it is not tricky.
All we need to reverse params in array_merge, then we will be able to put appropriate 'timeout' keys in our configs. I wanted to create an issue in SK, but then noticed that ```local.php``` (and some other configs) are not in git, but ```*.dev``` configs seem to be not used at ```CM_Config::_init()```
Could you clarify a little bit about our config system, please?

tomaszdurka:
Keep in mind we would not like to change timeout for other operations than createIndex.

@fvovan please read carefuly. I see two ways:

1. set it in config before any operations
2. set it in config within cli or within createIndex - results in changing timeout for all future connections, doesn't change existing connections

So if you choose solution 1 it won't be tricky. Solution two requires change of existing connections - kinda tricky, plus reverting them after operation.

fvovan:
Well, if we want to use different connection settings for different operations (different operation group) and all there is no clear way to change client configuration "on the fly" (because initialization runs once in the client constructor)  what if we just create another "service" for Elastic Search CLI operations, we could inherit it from ```CM_Elasticsearch_Cluster``` and override only constructor. 
wdyt?

fvovan:
Just to illustrate what i mean,
also we need to add this service to SK configs, and possibly use it for all CLI operations. 
@tomaszdurka wdyt?

njam:
Tomasz is on holidays, so let me take over here.

I would avoid creating an additional configuration for an "elasticsearch" service with a longer timeout. The identifier "long" is quite ambiguous, plus it introduces duplication (two times the same *functional* configuration of IP address and port, but different timeouts).

How about setting the timeout in `createIndex()` on-the-fly? Afaiu this is the only place where we want this high timeout, and we always want it. So it should be enforced in this function?

As Tomasz mentioned then we also should revert it at the end of the function..
(Alternatively *maybe* we could "clone" the connection initially, change the timeout, and use it only for the  *index creation*. At the end of the function it could be closed.)

fvovan:
I think it is possible, but not sure if it will be more clear.

Probably  it’s not good idea to work with connections directly because they are encapsulated in Client class and it’s not our class then we had to inherit from it and probably to maintain compatibility.

Changing clients inside CM_Elasticsearch_Type_Abstract::createIndex() method could set this object to inconsistent state (we need to care on proper shutdown of these clients, maybe even add exception checking to make sure we put them back, generally changing shared resources state is always not good idea).

What if we lazily duplicate clients array (or even just create only few of them) inside CM_Elasticsearch_Cluster::getTypes()? Then we add boolean param to this method and if it says that we should use long timeOut (or possibly more meaning name) then another clients will be lazily initialised and passed down for further processing.

njam:
Uh oh tricky... how about setting the timeout globally? :D

fvovan:
Do you mean 'globally' for all operations, not only for CLI index operation?
> Uh oh tricky... 
Maybe I described it  overcomplicated :). Actually it is only 15 lines of code, will demonstrate it in the evening.

fvovan:
well, it's not 15 lines but nevertheless 
@njam could you have a look please?

njam:
Let's change it hardcoded to 30 (used by Tomasz).

fvovan:
Last trial :) 
The idea flashed upon me, we just can get safe using try-catch
And we even hadn't to clone anything. Just save all connection timeouts to array and then switch it back.
There is some code duplication during putting it back, but it can be refactored
If we had php 5.5 we could use single finalize statement and it could be idiomatic.
@njam wdyt ?

njam:
Not bad :)
Still a lot of boilerplate code which shouldn't be necessary.
I would rather prefer to set a timeout for the specific operation in question (using `elasticsearch-php`).

fvovan:
I increased it globally and switched argument's positions in ```array_merge()``` to allow changing it using application settings
@njam please review

njam:
lgtm!
Let's look into `elasticsearch-php`!

PR #1900, opened 2015-09-09
Throw warning if ajax method is called that doesn't exist

fvovan:
Handled in `CM_Http_Response_View_Ajax`.
Currently fails with:
```
ErrorException: E_ERROR: Call to undefined method SK_Component_Chat::ajax_getChatDataList() in /home/example/releases/20150908135405/vendor/cargomedia/cm/library/CM/Http/Response/View/Ajax.php on line 29 
```

Instead I suggest we check if the method exists, and otherwise throw a warning, because this is user input and we have no control over it.

@fvovan could you look into this?

### DIFF
--- a/library/CM/Http/Response/View/Ajax.php
+++ b/library/CM/Http/Response/View/Ajax.php
@@ -27,12 +27,10 @@ class CM_Http_Response_View_Ajax extends CM_Http_Response_View_Abstract {
 
         $this->_setStringRepresentation(get_class($view) . '::' . $ajaxMethodName);
 
-        if (method_exists($view, $ajaxMethodName)) {
-            $data = $view->$ajaxMethodName($params, $componentHandler, $this);
-        } else {
+        if (!method_exists($view, $ajaxMethodName)) {
             throw new CM_Exception_Invalid('Method not found: `' . $ajaxMethodName . '`', CM_Exception::WARN);
         }
-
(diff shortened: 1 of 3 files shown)


fvovan:
@njam please review.

njam:
- For simpler control flow let's do `if (!$notExist) { throw new Exception() }`?
- Could you add a test?

fvovan:
retest this please

fvovan:
@njam it is done. Is it okay that I added tests to CM_Http_Response_View_AbstractTest class?
There are no own tests for various View classes.

njam:
- I think the test should be in a file specifically for the Ajax response, as it's about code in that file. Can you add such a test if you agree?
- In the unit test we could use the PHPUnit exception annotations? If yes then this would make it much shorter: https://phpunit.de/manual/current/en/writing-tests-for-phpunit.html#writing-tests-for-phpunit.exceptions

fvovan:
@njam please review
I just added one more test to check another possible warining

njam:
- Response/View/AjaxTest.php: I would usually try to avoid declaring mock classes directly in the test files. We have it in some places, where a different solution would be very difficult. But I think here it could be done with mocking. I understand we just need a valid View instance. Would this work?
```php
$view = $this->getMockForAbstractClass('CM_View_Abstract');
```

fvovan:
ah, I copied it from AbstractTest.php :)

PR #1902, opened 2015-09-09
Remove background color of contentPlaceholder.

vogdb:
I changed image lazyload library from [Unveil](https://github.com/luis-almeida/unveil) to [Lazyload](https://github.com/tuupola/jquery_lazyload). Lazyload shows grey previews instead of real images by default and this behaviour can't be configured. So, I propose to remove the same functionality that is currently implemented by us. I mean:
```less
  .contentPlaceholder-size + * {
    ...
    background-color: @colorBgBody;
  }
```

@christopheschwyzer @njam ?


vogdb:
@njam please look

njam:
lgtm!

PR #1909, opened 2015-09-11
Reset offlineStamp if set on all requests

alexispeter:


alexispeter:
please review @njam

njam:
- setOfflineStamp: Docu
- getOfflineStamp: Instead of returning early, just int-cast only in an `if`?
- CM_Session::start: Let's use type-safe comparison in `if`
- Test new functionality in `CM_SessionTest`
- Test fails: CM_Model_UserTest::testOfflineDelayed

alexispeter:
please review @njam

njam:
lgtm

alexispeter:
tagged https://github.com/cargomedia/cm/releases/tag/1.145.0

PR #1906, opened 2015-09-15
Check if we should switch from Elastica to elasticsearch-php

fvovan:
See https://github.com/elastic/elasticsearch-php
More low level API, completely sufficient for us.

They also have per-operation "timeout" configuration, so we could set a higher timeout for adding documents in `CM_Elasticsearch_Type_Abstract::update()` (see https://github.com/cargomedia/cm/pull/1894).

@fvovan can you check if it would make sense?


fvovan:
Conceptually it could be implemented that way.
I'am still having some problems with mapping API, but can be reviewed just generally.
@njam, @tomaszdurka could you take a look on it, please?

Also I'am think about hiding all request parameters routines inside some kind of facade for Client.
We could put all '_' prefixed indices and document processing methods there.

njam:
SK/Usertext/Usertext.php
- _setMode: Beginning of this function is a bit difficult to read. I would first do *only* option parsing. How about `$optionsDefaults`, `$options` and `$optionsOverrides` - then merge the three? ..and move `$highlightSpamwords` down to where it's used?

fvovan:
>SK/Usertext/Usertext.php

@njam probably wrong PR

njam:
CM_Elasticsearch_Cluster:
- getTypes: Rename to `getTypeList` for consistency (see [guides](https://github.com/cargomedia/guides/tree/master/style#naming))

CM_Elasticsearch_Document:
- $_data: Use single-line docu (having "document data" there doesn't add much information, since the class is called "document" and the field "data")
- I would remove "chainable setters", just because we don't usually use them - unless there's a good use case?
- I would remove magic functions, because they are *magic* :) - again unless there's a good use case
- set: The check `is_array($this->_data)` is unnecessary, because it's always an array
- has: Same
- add: Remove this function?
- addFile, addFileContent: Remove this function?

CM_Elasticsearch_DocumentTest:
- Move testing of `create()` into separate test

CM_Elasticsearch_Type_Abstract:
- INDEX_NAME: What do you about making this an abstract function? Seems a bit dodgy with a const
- getIndex: Remove in favor of `getIndexName`
- createIndex:
 - $indicesHandler: Rename to `$indicesNamespace`? Because that's the name of the class.
 - _putIndexSettings: Write on 1 line, for readability?
 - 
- _getIndexesByAlias: Return docu could read `string[]`
- Static functions: I would make whatever doesn't *have* to be static rather non-static. It will make it easier to test it (stub it).
- How about introducing an abstraction for "Index", to move out some of the logic to there?
- convertDate: I would simplify this, now that it's in our code base. How about accepting `DateTime|int` - but not `string`?

fvovan:
> INDEX_NAME: What do you about making this an abstract function? Seems a bit dodgy with a const

I agree but can't be switched to abstract function easily, static ```updateItem($item)``` uses it and it seems to be hard to change it invocation contexts. But I can make an interface with static method.

njam:
> But I can make an interface with static method or constant.

Interface with static method sounds good? ...or just a static function on the abstract class, implemented as:
```php
abstract protected function _getIndexName() {
 throw new CM_Exception_NotImplemented();
}
```
..and then implement it on the non-abstract children.

> Actually I'd rather prefer to move it into some wrapper for ```Client``` probably ```CM_Elacticsearch_Client```, many currently  protected methods (like ```_refreshIndex(), _deleteIndex``` bulk methods etc) could be moved there. 

I just saw that most of these functions have a `$indexName` argument, so I thought they could be wrapped into a class which stores this "indexName".

fvovan:
> I just saw that most of these functions have a $indexName

That's true but actually this param can be an array. Probably either naming should be improved or it is worth to make it array strictly. And natively this operations should be able to receive arrays because client can process them in a batch. Separating methods (by single or couple index) doesn't seem to be a good idea too.
It would be pretty weird to have such method signatures inside ```Index``` class but as an option we can make ```Index``` just as helper for using all these method. So when we use 
```$index->update(..)``` actually it will delegate all request to ```client```

```
//inside index
public function update ( $params ) {
  $this->client->update($this->name, $params);
}
```
I'm not sure about how it will overcomplicate things for us.

njam:
Makes sense, moving it to a Client sounds like a good idea then! (à la `Client::updateIndex(string|array $index, $params)`)

fvovan:
did moving
@njam  please review
at the meantime i will add tests for new Client wrapper.

njam:
Looks very good so far!



CM_Elasticsearch_Alias:
- Call it `CM_Elasticsearch_AliasInterface`?

CM_Elasticsearch_Client:
- bulkDeleteDocuments: $idsDelete: Rename to `$idList`
- createIndex: Drop `$recreate` argument and just do that in usages?
- createIndex: $useSource: Set to default (`false`) at beginning of function
- createIndex: $newIndexName: Rename to just `$indexName`?
- createIndex: Get rid of "hack" - let's adjust usages
- Add tests

CM_Elasticsearch_Cluster:
- How does it related to `CM_Elasticsearch_Client`? Both have an instance of `Elasticsearch\Client` - shouldn't this be wrapped in one class only? I see we use sometimes `Cluster`, sometimes `Client` and sometimes `Elasticsearch\Client` if we need a reference to a client. Should we streamline this?

CM_Elasticsearch_Document:
- Use `[]` instead of `array()`
- get: Should it throw a specific message if a field doesn't exist?
- remove: Should it really throw if it doesn't exist?
- toArray: Remove in favor of just `toArray()`?

CM_Elasticsearch_Type_Abstract:
- Some open todos

Elasticsearch/UpdateDocumentJob.php:
- $elsticClient: Should be `$elasticsearchClient`. Maybe rename to just `$client`? (also in a few other places)

fvovan:
>toArray: Remove in favor of just toArray()?
Do you mean "in favour of  ``` getData()```" ?

njam:
Yes :)

fvovan:
retest this please

fvovan:
> CM_Elasticsearch_Cluster How does it related to CM_Elasticsearch_Client? 

I thought about it today, think it will better to not use \Elasticsearch\Client directly, so we will wrap it in Cluster's constructor and then use it everywhere, so Cluster and all Types will share the same CM_Elasticsearch_Client instance.

fvovan:
@njam please review.
sk part is ready too https://github.com/cargomedia/sk/pull/2144 and I tested it locally, but yet have no idea how to make composer vcs repo work with remote repository.
Will try to figure it out.

njam:
Overall:
- Rename `$elasticClient` to `$elasticsearchClient` - use a text search?

CM_Elasticsearch_ClientTest:
- Add docu to getters
- Add specific test for `count()`? Covering the case when `$responseCount['count']` is not set maybe?

composer.json:
- You can add needed composer packages to our [satis server](https://github.com/cargomedia/puppet-cargomedia/blob/master/modules/cargomedia/templates/satis/dist.json) to make them available without packagist in repos like SK.

CM_Elasticsearch_Client:
- createIndex: Only set default for `$useSource` if `null === $useSource`?
- createIndex: Let's not delete the index here if it exists - this might be quite unexpected? We can just do it in the usage if desired?
- search: Should it maybe also accept `string|string[]` for index and type?
- _prepareIndexNameParam: How about adding a `$minimumCount` param here for checking if we need at least 1 index for example?

CM_Elasticsearch_Cluster:
- construct: Introduce some variable(s) to make this heavy one-liner to assign the client a bit simpler
- findType: Delete because unused?

CM_Elasticsearch_Document:
- Let's remove some not so useful docu comments (like "Creates a new document"? IMO all those can be removed. wdyt?)

CM_Elasticsearch_Type_Abstract:
- createIndex: Introduce `$client` for readability?

CM_Elasticsearch_Type_AbstractTest:
- Add test for count?

Rest looks very good

fvovan:
Found very weird error in \Elasticsearch\Client. If we call ```count``` (or ```search```) method on just created index, it fails with 503 response code although response body is okay.
This behaviour is not correct, curl queries give 200 (and the same body).
It seems this 503 code is added somewhere deep in ``` \Elasticsearch\...``` class.
Don't know if it is worth to investigate is. These situations probably will never happen in our case but looks very suspicious.

fvovan:
all is done except 2 things.
1)
>_prepareIndexNameParam: How about adding a $minimumCount param here for checking if we need at least 1 index for example?

I'm not sure if it is good idea, because I think decision about further processing should be made inside it's caller. And besides, it would helps us to keep docu Exception signatures relevant.

2) 
>You can add needed composer packages to our satis server to make them available without packagist in repos like SK

No problems, but is it right place (I mean linked file) to do it? Shouldn't other CM dependencies like functional-php or guzzle be there?

njam:
> I'm not sure if it is good idea

Agree, let's keep it.

> Shouldn't other CM dependencies like functional-php or guzzle be there?

Yes you're right. Satis resolves dependencies recursively, so it will be made available automatically.

CM_Elasticsearch_Client:
- count: About this 503 quirk. I would suggest to open a ticket in the original library about this, and then link the ticket as a comment in our workaround.
- createIndex: I guess you wanted to delete that `deleteIndex()` call?
- _getCmClient: Here we should use `$this->getServiceManager()` - the idea is that tests can have a separate service environment (..and to reduce global state).

CM_Elasticsearch_ClientTest:
- testCount: Can we do it without this `ElasticClientMock`? I understand we just need a class with a `count()` method. This can be easily done with [mocka](https://github.com/tomaszdurka/mocka) for example.

fvovan:
https://github.com/elastic/elasticsearch-php/issues/303

fvovan:
@njam please review

njam:
lgtm!

fvovan:
Owner of elasticsearch repo responded 
https://github.com/elastic/elasticsearch-php/issues/303#issuecomment-147712408

njam:
@fvovan and it was only happening in test code for us? If so, how about we move this workaround (or the one he suggested) out of production into the test code?

fvovan:
Yes, it is only for test code, we can remove error handlings from client and then insert blocking code (awaiting for shard propagation) to test method

PR #1910, opened 2015-09-15
Flexible usertext options

alexispeter:
For https://github.com/cargomedia/sk/issues/2066

### DIFF
--- a/library/CM/SmartyPlugins/function.usertext.php
+++ b/library/CM/SmartyPlugins/function.usertext.php
@@ -12,21 +12,20 @@
 function smarty_function_usertext($params, Smarty_Internal_Template $template) {
     /** @var CM_Frontend_Render $render */
     $render = $template->smarty->getTemplateVars('render');
+    $options = $params;
     $text = (string) $params['text'];
+    unset($options['text']);
     $mode = (string) $params['mode'];
-    $maxLength = isset($params['maxLength']) ? (int) $params['maxLength'] : null;
-    $isMail = isset($params['isMail']) ? (bool) $params['isMail'] : null;
-    $skipAnchors = isset($params['skipAnchors']) ? (bool) $params['skipAnchors'] : null;
-    $allowBadwords = isset($params['allowBadwords']) ? (bool) $params['allowBadwords'] : null;
-    /** @var CM_Usertext_Usertext $usertext */
-    $usertext = isset($params['usertext']) ? $params['usertext'] : null;
-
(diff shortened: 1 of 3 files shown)


alexispeter:
please review @njam

njam:
CM/Usertext/Usertext.php:
- addFilterAfter: If not match is found this will be a noop. Should it maybe add to the end, or throw?
- setMode: Let's use proper multi-line array syntax for `$optionsDefault` (each element on separate line)
- setMode: How about adding an `isMail` option here, which sets the `emoticonFixedHeight` option? This would move some rather heavy logic out of the smarty plugin, and make it reusable in the usertext class. Or are there any downsides of it?

alexispeter:
> addFilterAfter: If not match is found this will be a noop. Should it maybe add to the end, or throw?

Usually, you would only use addFilterAfter() if you have to insert the filter at a very specific slot in the pipeline. Adding it to a nonexistent filter is just plain nonsense that shouldn't escape the programmer's notice. But throwing an exception to be sure is fine.

> setMode: How about adding an isMail option here, which sets the emoticonFixedHeight option? This would move some rather heavy logic out of the smarty plugin, and make it reusable in the usertext class. Or are there any downsides of it?

I moved it outside because I don't think we should be using "aliases" for certain parameter-values. 
Maybe it should be moved outside of the smarty-helper as well?

alexispeter:
please review @njam

njam:
- setMode: Docu
- setMode: Use correct multiline array notation (every element on one line, with a comma at the end)
- isMail: There should definitely be one place where we define `emoticonFixedHeight = 16`. We don't want this value spread around different files (and possibly repos). Let's leave it in the smarty function then.

Additional notes from the other ticket:
- `CM_Usertext_Usertext::transform()` uses `Render` from instance, which has to be set with another function. How about passing it to `transform()`?

alexispeter:
please review @njam

njam:
lgtm

alexispeter:
totally forgot this :D

please review @njam

njam:
Exception constructor's signature changed! You wanna pass metaInfo? I'd say it's not needed here, because it's already part of the message?

alexispeter:
Right, forgot about this absolutely terrible change, will remove it.

njam:
lgtm

alexispeter:
tagged https://github.com/cargomedia/cm/releases/tag/1.145.0

PR #1911, opened 2015-09-16
Get rid of error 'CM_Exception_Invalid: Cannot find valid class definition for view' 

fvovan:
Change exception type  (to warning) in CM_View_Abstract::factory()


fvovan:
@njam  please review

njam:
I think we should keep it an "error" in the factory, as this is core business logic and potentially used in various places. I suggest we only make it a warning if it comes directly as user input through an AJAX request.
Maybe we could move the extraction and check of the classname out of `CM_Http_Response_View_Abstract::loadComponent` into `CM_View_Abstract::ajax_loadComponent`?

fvovan:
@njam  review please

njam:
- Could you add a test for `ajax_loadComponent()` that covers the new behaviour?
- What do you think about passing the class name normally down to `loadComponent()` (instead of within `$params`), now that we extract it anways?

fvovan:
tests for ```CM_View_Abstract``` are completely absent, will create them

> What do you think about passing the class name normally down to loadComponent() (instead of within $params), now that we extract it anyways?

sounds reasonable but will require us to change SK too, because it has several usages.

njam:
True, but I think it makes sense! Let's do itè

fvovan:
@njam please review
maybe since we extracted className value from params, then we should remove it (from ```$params```)?

njam:
Yes agree, we should remove it.

- CM_CaptchaTest: removed by accident?
- CM_View_AbstractTest::testConstructor: No need to check instance of `$view` as we create it manually

fvovan:
@njam thanks!
yes , it seems I moved but not copied draft for test. I reverted it but for some reason it is shown like complete recreation

njam:
- CM/CaptchaTest.php: Something's still off, if you check the diff. Probably whitespace is different? Can you check if the IDE is configured to use the "cargomedia" code styles?

fvovan:
Old (deleted) version used CRLF for line breaks.
Reverted version was automatically converted with git to use just LF.

njam:
ok, lgtm then!

fvovan:
I will create new CM tag and prepare sk PR very soon

PR #1912, opened 2015-09-18
Allow DateTime for assertSameTime

tomaszdurka:


tomaszdurka:
@zazabe please review

PR #1914, opened 2015-09-21
Pass the last runtime to clockwork callbacks

zazabe:
A nice feature could be to have the last execution time passed to clockwork callbacks, some procedure may require it to do periodical processing since the last execution.

If the callback has never been called, we can simply set the argument as `null`.

### DIFF
--- a/library/CM/Clockwork/Event.php
+++ b/library/CM/Clockwork/Event.php
@@ -49,7 +49,7 @@ class CM_Clockwork_Event {
     /**
      * @param DateTime|null  $lastRuntime
      */
-    public function run($lastRuntime) {
+    public function run(DateTime $lastRuntime=null) {
         foreach ($this->_callbacks as $callback) {
             call_user_func($callback, $lastRuntime);
         }
--- a/library/CM/Clockwork/Event.php
+++ b/library/CM/Clockwork/Event.php
@@ -47,9 +47,9 @@ class CM_Clockwork_Event {
     }
 
     /**
-     * @param DateTime|null  $lastRuntime
+     * @param DateTime|null $lastRuntime
      */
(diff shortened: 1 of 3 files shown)


zazabe:
@njam can you review it please?

library/CM/Clockwork/Event.php
```diff
@@ -46,9 +46,12 @@ public function registerCallback($callback) {
         $this->_callbacks[] = $callback;
     }
 
-    public function run() {
+    /**
+     * @param DateTime|null  $lastRuntime
+     */
+    public function run($lastRuntime) {
```
tomaszdurka: Add type hinting, allow null.

njam:
..and apply formatting, otherwise looks good!

tomaszdurka:
@alexispeter can you please confirm this one?

tomaszdurka:
Paired with @zazabe .

alexispeter:
We explicitly set the end of execution as runTime so the interval between long-running processes stays constant.  Is there a specific reason for this change?

zazabe:
@alexispeter we have a talk with @njam and @tomaszdurka about this point, it's true that we change the  behaviour for interval event, now if a process take more time than the interval, it will be run again directly without waiting.
We did this because for some clockwork process, we may need the lastRuntime (like for mail report processing) and we thought it's a good feature to add at the clockwork level, but with this, we need to set the lastRuntime at the beginning of the process and not at the end.

Do you see any troubles we may have by introducing this change?

alexispeter:
Just asking because it wasn't mentioned.

As for possible troubles, the advantage of using "end of execution time" was that it made the maintenance daemon "sensitive" to load on the site, that was the only reason as far as I can remember. 

Probably not going to have any negative consequences.

PR #1919, opened 2015-09-23
Fix clockwork wrapper handling throwed exception twice

zazabe:


zazabe:
paired with @tomaszdurka

PR #1890, opened 2015-09-24
FB-292 Extend FormFields: Textarea

vogdb:
For all browsers use:

``` html
<div contenteditable="true"></div>
```

(Remove all HTML client-side)

For supported browsers use ([feature detection](http://stackoverflow.com/questions/10672081/how-to-detect-if-browser-supports-plaintext-only-value-in-contenteditable-para/18316972)):

``` html
<div contenteditable="plaintext-only"></div>
```

---

Upsides
- We can remove _autosize_ plugin
- We can use html (preview emoticons, etc.)

Concerns: 
- Behavior on mobile, browser support
- Form specific attributes:
  - Placeholder: http://stackoverflow.com/questions/20726174/placeholder-for-contenteditable-div
  - resizing: ignore
  - :focus: works
  -  tabindex: works?
- Validation


njam:
How do we handle newlines? I understand browsers will add `<br>` or `<p>` tags even in `plaintext-only`-mode. If that's the case I guess we need to transform those to newlines on the server?

vogdb:
You are right. Browser adds those tags for formatting. Firefox uses `<br>`, Chrome - `<div>` and `&nbsp;`

njam:
So we need both way transformation (adding html from server->client, and adding newlines from client->server).

njam:
Let's also make sure `enableTriggerChangeOnInput()` still works

njam:
First impression is very good, seems to work nicely!
@NicolasSchmutz would be great if you could do some acceptance testing on all supported devices asap, so we can be sure this is the way to go.

- Textarea/default.less: Use variable for placeholder color (see `.placeholder()`-mixin in `form.less`)
- css/form.less: Should use `@heightInputTextarea` - which should be the minimum/default height
- CM/Component/Debug.js: Maybe instead check for `contenteditable` attribute here, to not make it depending on the formfield?
- CM/FormField/Textarea.js:
 - Is this "paste" handler adopted from somewhere else? Where? It's difficult to say if it works in all cases / all browsers. Should we just acceptance test it?
 - Make sure `getValue()`/`setValue()` work
   - Maybe override `getInput()`?
 - Make sure `hasFocus()` works
 - Make sure events `blur`/`focus` are triggered
 - Make sure event `change` is triggered (see `enableTriggerChangeOnInput()`)

njam:
...oh and of course the transformation from server- to client-values and back (html to text).

NicolasSchmutz:
please see: https://github.com/cargomedia/sk/pull/2145#issuecomment-144683474

njam:
As discussed let's try doing the html->text->html transformation in PHP. Some libraries:
- https://github.com/soundasleep/html2text
- https://github.com/thephpleague/html-to-markdown
@tomaszdurka can you push to Aleksei's branch?

tomaszdurka:
Browser behaviour:
- OSX
  - Chrome, Firefox, Safari (<br> on enter and shift+enter)
- Windows
  - IE Edge (<div> on enter, <br> on shift+enter, <div><br></div> on empty lines)
  - IE11 (<p> on enter, <br> on shift+enter, <p><br></p> on empty lines)
- Android
  - Android 4.4 (<br> on enter and shift+enter)
- iOS
  - Safari (<br> on enter and shift+enter)
- Windows Phone
  - Lumia 930 (<p> on enter and shift+enter, <p><br></p> on empty lines)

vogdb:
@njam 

> - css/form.less: Should use `@heightInputTextarea` - which should be the minimum/default height

It becomes huge. 
screen shot 2015-10-02 at 17 26 35

> - Is this "paste" handler adopted from somewhere else? Where? It's difficult to say if it works in all cases / all browsers. Should we just acceptance test it?

I took it from here http://stackoverflow.com/a/19269040/495656

>  - Make sure events `blur`/`focus` are triggered

> - Make sure event `change` is triggered (see `enableTriggerChangeOnInput()`)

Couldn't test it on Windows Phone. Seems that my test device acquired some glitch that does not allow me to test anymore on it.

vogdb:
@tomaszdurka did you check that replace of `_initChangeEmitter` with `enableTriggerChangeOnInput ` works?

tomaszdurka:
I saw it's almost exactly the same behaviour as in `CM_FormField_Text` - as you see I've also adjusted that one a bit. Didn't do extensive testing yet - I will do it later though.

You have some objections? You were storing value on initial focus - we stored it on `ready` / load. I've also added `paste` event to parent implementation.

vogdb:
does not seem to work.

alexispeter:
please review @tomaszdurka

alexispeter:
please review @tomaszdurka

zazabe:
I don't want to break all the things done until now, but why not handling this on the client side with a markdown plugin like [medium.js](http://ionicabizau.github.io/medium-editor-markdown/) and using it for markdown <-> html conversion ?

Pros:
- easy to handle on the server side, just store the markdown text...
- support more formatting options out of the box (bold, italic,...)
- easy support of emoticon by adding a medium plugin
- real WYSIWYG

Cons:
- all markdown text have to be converted every time, on the client side => performance issue?

zazabe:
another nice/well designed wysiwyg editor: http://trix-editor.org

christopheschwyzer:
Trix looks interesting. Might be a solution.

vogdb:
IMO libraries like medium.js or http://trix-editor.org are too powerful and redundant for our needs. User expects the usual behaviour of input text field from fields like CommentAdd, Blogpost description, Conversation. Just look at Facebook, for example. There is no Bold, Italic and other fancy WYSIWYG functionality. So, we should not add it, even implicitly.

> - easy to handle on the server side, just store the markdown text...

Use enters plain text. Isn't it easy to store?

> - support more formatting options out of the box (bold, italic,...)

For what reason? Us to be fancy? Users will complain about new complications in their life.

> - easy support of emoticon by adding a medium plugin

Do we have a difficult support now?

> - real WYSIWYG

Again, redundant. Also opens a hole for scammers to write a very attitude messages with bold, uppercase header font.

vogdb:
I've changed `CM_Formfield_Textarea` so it always returns plain text stripped of any `contenteditable` formatting like `<br>`, `<div>` and others. Shall we remove `CM_HtmlConverter->convertHtmlToPlainText`?
@tomaszdurka, @njam ?

vogdb:
@tomaszdurka I removed `CM_HtmlConverter`. If you have any objections, please tell them.

tomaszdurka:
Ok, I don't remember this code too much. I can recall that all the browsers had different way of putting new-lines. It's much better to do that on the client-side.

layout/default/FormField/Textarea/default.tpl
```diff
@@ -1 +1 @@
-{tag el="textarea" name=$name id=$inputId content=$value|escape class="textinput {$class}" maxlength=$options.lengthMax tabindex=$tabindex placeholder=$placeholder}
+{tag el="div" contenteditable="true" name=$name id=$inputId content=$value|escape class="needsclick textarea textinput {$class}" tabindex=$tabindex placeholder=$placeholder}
```
njam: .needsclick class not needed

library/CM/App.js
```diff
@@ -520,6 +518,21 @@ var CM_App = CM_Class_Abstract.extend({
       audio.setOptions(options);
       audio.setSources(sourceList);
       return audio;
+    },
+
+    /**
+     * @param {String} text
+     */
+    pasteTextAtCursor: function(text) {
```
njam: How about moving this into FormField/Textarea.js?
I can see that this *could* be reused, but I'm also not sure if in any context. Could it be used also in a normal text field? We can just move it into Textarea, and investigate reusability later when needed, wdyt?

christopheschwyzer:
Made some progress:
> Field decreases height by 1px when gaining focus

Can't confirm?

> Newlines are duplicated. If I enter "hello \n world", the server value (and the client value) is "hello \n\n world".

Can't confirm?
screen shot 2017-03-07 at 13 00 02

I was also wondering if we should introduce a default `lengthMax` for text fields. (I understand it's now limited by the database?)

library/CM/FormField/Textarea.js
```diff
@@ -3,5 +3,119 @@
...
+  getEnabled: function() {
+    return true;
+  },
+
+  reset: function() {
+    this.setValue('');
+  },
+
+  _initPlaceholder: function() {
+    this.getInput().focusout(function() {
```
njam: Could this be moved to Backbone.View's `events` binding instead?

library/CM/FormField/Textarea.js
```diff
@@ -3,5 +3,119 @@
...
+      var $this = $(this);
+      if (!$this.text().trim().length) {
+        $this.empty();
+      }
+    });
+  },
+
+  _initPlaintextonly: function() {
+    var self = this;
+    this.getInput().on('paste', function(e) {
```
njam: Could this be moved to Backbone.View's `events` binding instead?

library/CM/FormField/Textarea.js
```diff
@@ -3,5 +3,119 @@
...
+    } else {
+      this.error(null);
+    }
+  },
+
+  _scrollToCursor: function() {
+    var $input = this.getInput();
+    var selection = window.getSelection();
+    var range = selection && selection.getRangeAt(0);
+    var textCoords = range && range.getClientRects()[0];
```
njam: Maybe rename to `selectionCoords`, it's more consistent and easier to read.

resources/translations/en.php
```diff
@@ -55,4 +55,5 @@
     $language->setTranslation('An unexpected connection problem occurred.');
     $language->setTranslation('Unable to detect location');
     $language->setTranslation('No Internet connection');
+    $language->setTranslation('Too long');
```
njam: We should also update de.php

christopheschwyzer:
@njam let me know (to pair finish this maybe)

library/CM/FormField/Textarea.js
```diff
@@ -1,7 +1,117 @@
-/**
+/**selectionCoords
```
njam: remove

library/CM/FormField/Textarea.js
```diff
@@ -1,7 +1,117 @@
...
+
+  events: {
+    'blur [contenteditable]': function() {
+      this.trigger('blur');
+    },
+    'focus [contenteditable]': function() {
+      this.trigger('focus');
+    },
+    'focusout [contenteditable]': function() {
+      this._initPlaceholder();
```
njam: This function should now be renamed (e.g. "_onFocusOut()"), and we don't need to call it from `ready()` anymore.

njam:
Notes from testing:
- The client-side value `hello<b>` will reach the server as `hello&lt;b&gt;`, which should not be.
- When cutting and pasting, and cutting and pasting a text again, suddenly the field's value will be 2 newlines, while the visible value is empty.

christopheschwyzer:
Hm still issues
- Copy/paste creates all sorts of incompatibilities. 
  - Chrome:  `A\nB\nC` becomes `<div>A</div>…` in markup -> "ABC" if retrieved with `$input.text()` or `$input[0].textContent` -> we could use `$input[0].innerText`
  - Edge 14: `A\nB\nC` (`A <br />…`) becomes `A B C`

Maybe have a look at https://github.com/guardian/scribe

njam:
Closing in favor of https://github.com/cargomedia/cm/pull/2609

Issue #1920, opened 2015-09-23
MessageStream::startSynchronization respawns regularly


njam:
Dies every ~minute and is respawned.
The original reason (exception?) seems not to be logged.

cc @zazabe @alexispeter

tomaszdurka:
- Cannot reproduce locally so far.
- Turns out there is no exception and it exits with 0.
- Not connected with the CLI, we've run startSynchronization separately and still quits.
- We tried running `redis->subscribe()` with simple dumping channel/message and it never died.

zazabe:
after a look at [the diff](https://github.com/phpredis/phpredis/compare/2.2.3...phpredis:2.2.7) between `phpredis 2.2.3` and `phpredis 2.2.7`, there's some improvements that can helps maybe:

- phpredis/phpredis/pull/493 : Allow to break out of subscribe loop   
  the guy fixed a memory leak issue

I'm not sure it will solve this issue, i think the problem is somewhere here: 
https://github.com/phpredis/phpredis/blob/2.2.7/redis.c#L5669

maybe the way we define the callback is overcomplicated, with a callback defined as a property of the redis client (`$_subscribeCallback`) which is also a method name of the same class...
https://github.com/cargomedia/cm/blob/master/library/CM/Redis/Client.php#L273

zazabe:
I created this ticket about phpredis:  [Throw an exception when the subscribe method fail at some point](https://github.com/phpredis/phpredis/issues/662)

FYI, not throwing exception seems to be a decision since the creation of this extension:
> Since the beginning of the phpredis project a design decision was made to not throw exceptions except in the case of communication errors.
https://github.com/phpredis/phpredis/issues/502#issuecomment-54200996

I will investigate https://github.com/nrk/predis to replace phpredis, there's less activity on this project but [Redis recommend it also](http://redis.io/clients#php).

njam:
@zazabe fixed by https://github.com/cargomedia/cm/pull/1923?

zazabe:
@njam I don't see any "Respawning dead child..." warning exception in the admin, so I suppose we can close this ticket, isn't it?

njam:
Very good

PR #1923, opened 2015-09-25
Breaking: Switch from phpredis client to predis

zazabe:
To avoid having more problems with https://github.com/phpredis/phpredis, we decided to switch to an alternative Redis client, https://github.com/nrk/predis.

The task is related to https://github.com/cargomedia/cm/issues/1920.

Note: the merge request is marked as "Breaking" even if our Redis client API has not changed, but we may have some issues because all cases are probably not well tested.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -18,7 +18,8 @@
     "kissmetrics/kissmetrics-php": "~0.1.0",
     "guzzlehttp/guzzle": "~4.1.7",
     "kickbox/kickbox": "~1.0.3",
-    "tomaszdurka/codegenerator": "~0.5.0"
+    "tomaszdurka/codegenerator": "~0.5.0",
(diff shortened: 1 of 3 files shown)


zazabe:
@tomaszdurka please review

zazabe:
@njam I changed the `read_write_timeout` connection parameter because an exception was raised after 1min, I tried to set it to `-1` to disable the timeout, then I subscribed to an idle channel during more than 15min without having this problem again.
Could you review my changes?

njam:
- sFlush: `sMembers()` call should also be within `multi`/`exec` for it to be atomic, and use `DEL`
- subscribe: Add comment for `@param` about "Return something else than `null` to exit"

zazabe:
@njam please review

njam:
lgtm

PR #1925, opened 2015-09-29
CM_MenuItem::compare() fails when comparing scalars with arrays

alexispeter:
Present implementation fails when comparing scalars with arrays, i.e. when a parameter is defined as scalar value on a menu-item but also has valid array-values (e.g. an object is identified by an id and type) or vice versa.
It fails with an `E_NOTICE: Array to string conversion` because array_intersect_assoc() casts values to strings to do a strict comparison.

Addendum: This obviously applies to all cases were either the parameter on the menu or the one compared to it is an array.

### DIFF
--- a/library/CM/MenuEntry.php
+++ b/library/CM/MenuEntry.php
@@ -51,6 +51,8 @@ class CM_MenuEntry {
             $paramsMatch = array_uintersect_assoc($this->getParams(), $params, function ($a, $b) {
                     if (is_array($a) && is_array($b)) {
                         return (array_diff_assoc($a, $b) === array_diff_assoc($b, $a)) ? 0 : -1;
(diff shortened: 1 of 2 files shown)


alexispeter:
please review @njam 

Doesn't distinguish values like int(0) empty string anymore, but I can't imagine a scenario where that would be a problem.

njam:
I would keep the strict comparison. For example `count=0` is different from `count=`? Or what's the reason to not do it?

alexispeter:
please review @njam

njam:
lgtm

alexispeter:
please review @njam

last commit covers cases were the menu-value is a scalar and the param-value an array or vice versa (Actually the functionality I described in this issue's description).

njam:
lgtm

PR #1926, opened 2015-09-29
Logging system refactoring

zazabe:


zazabe:
@fvovan please review

library/CM/Log/Context.php
```diff
@@ -0,0 +1,78 @@
...
+    /** @var CM_Http_Request_Abstract|null */
+    private $_httpRequest;
+
+    /** @var string[] */
+    private $_extra = [];
+
+    /**
+     * @param array|null $options ['user' => CM_Model_User|null, 'http-request' => CM_Http_Request_Abstract|null, 'extra' => []|null]
+     */
+    public function __construct(array $options = null) {
```
fvovan: If we have type for certain method argument I would pass it as separated parameter as methodName(CM_Model_User $user=null, CM_Http_Request $request=null, array $extra = null)

library/CM/Log/Handler/Abstract.php
```diff
@@ -0,0 +1,56 @@
+<?php
+
+abstract class CM_Log_Handler_Abstract implements CM_Log_Handler_HandlerInterface {
+
+    /** @var  bool */
+    protected $_bubbling = true;
+
+    /** @var int */
+    protected $_level = CM_Log_Logger::DEBUG;
```
fvovan: probably it's better to move all initialization to constructor.

library/CM/Log/Handler/Abstract.php
```diff
@@ -0,0 +1,56 @@
...
+
+    /**
+     * @param integer $level  The minimum logging level at which this handler will be triggered
+     * @param Boolean $bubble Whether the messages that are handled can bubble up the stack or not
+     */
+    public function __construct($level = CM_Log_Logger::DEBUG, $bubble = true) {
```
fvovan: We try to not use other default arguments except null. It's always possible to check them for null at the method's beginning.

library/CM/Log/Handler/Abstract.php
```diff
@@ -0,0 +1,56 @@
...
+        $this->setLevel($level);
+        $this->bubble = $bubble;
```
fvovan: ```$this->_bubbling``` I guess

library/CM/Log/Handler/Abstract.php
```diff
@@ -0,0 +1,56 @@
...
+    /** @var  bool */
+    protected $_bubbling = true;
+
+    /** @var int */
+    protected $_level = CM_Log_Logger::DEBUG;
+
+    /**
+     * @param integer $level  The minimum logging level at which this handler will be triggered
+     * @param Boolean $bubble Whether the messages that are handled can bubble up the stack or not
+     */
```
fvovan: Please use lowercase names for primitives. and 'int' instead 'integer'

library/CM/Log/Handler/Abstract.php
```diff
@@ -0,0 +1,56 @@
...
+    public function getBubble() {
+        return $this->_bubbling;
+    }
+
+    /**
+     * @param bool $bubbling
+     */
+    public function setBubble($bubbling) {
+        $this->_bubbling = $bubbling;
+    }
```
fvovan: Add casting please

library/CM/Log/Context.php
```diff
@@ -0,0 +1,78 @@
...
+        $options = array_merge([
+            'user'         => null,
+            'http-request' => null,
+            'extra'        => null,
+        ], $options ?: []);
+
+        $this->_setComputerInfo();
```
fvovan: Probably it is better to pass ```$computerInfo``` to constructor too.

zazabe:
@fvovan 
I agree about the `CM_Log_Context` constructor, also I think we can pass a context instance directly to Logger methods, like: 
```
$logger->info('foo', new CM_Log_Context($user, $httpRequest));
```
instead of an obscure `$options` argument...

fvovan:
I think passing it each time would be redundant.
In most cases context remains the same between ```log``` method calls.
And computer info is probably even more constant thing. If we decide to put Logger to CM_Service_Manager I would prefer context and computerInfo initialization being located there.
And certainly we still need ability to define custom context per log record.

library/CM/Log/Handler/Abstract.php
```diff
@@ -0,0 +1,74 @@
...
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getLevel() {
```
fvovan: Invalid PHPDoc

zazabe: http://manual.phpdoc.org/HTMLSmartyConverter/HandS/phpDocumentor/tutorial_tags.inlineinheritdoc.pkg.html
Get inspired by monolog source code...

Should I remove the `{@inheritdoc}` ?

fvovan: yes, please, it doesn't contain any useful information

library/CM/Log/Logger.php
```diff
@@ -0,0 +1,194 @@
...
+    const ERROR = 400;
+    const CRITICAL = 500;
+    const NOTSET = 999;
+
+    /**
+     * Logging levels from syslog protocol defined in RFC 5424
+     *
+     * @var array $levels Logging levels
+     */
+    protected static $levels = array(
```
fvovan: can't we do ```self::DEBUG => 'DEBUG' ``` ?
btw why we need their names as strings? If it is only for fancy exception descriptions probably it is not needed.

library/CM/Log/Logger.php
```diff
@@ -0,0 +1,194 @@
...
+            }
+        }
+    }
+
+    /**
+     * @param string     $message
+     * @param int        $level
+     * @param array|null $options
+     */
+    public function addMessage($message, $level = self::NOTSET, array $options = null) {
```
fvovan: Please use null as default argument.
And what is NOTSET? What if we just add one more attribute for default level? And probably even make this method protected as we implemented plenty of helpers for it?

library/CM/Log/Context.php
```diff
@@ -0,0 +1,54 @@
...
+    private $_computerInfo;
+
+    /** @var CM_Model_User|null */
+    private $_user;
+
+    /** @var CM_Http_Request_Abstract|null */
+    private $_httpRequest;
+
+    /** @var string[] */
+    private $_extra;
```
fvovan: Probably extra will not always be string. It's better to do just array

library/CM/Log/Handler/Abstract.php
```diff
@@ -0,0 +1,76 @@
...
+        $level = is_null($level) ? CM_Log_Logger::DEBUG : $level;
+        $bubbling = is_null($bubbling) ? true : $bubbling;
+        $this->setLevel($level);
+        $this->setBubble($bubbling);
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getBubble() {
```
fvovan: It's better to use consistent name matching for attributes and getters/setters like $_bubbling => _getBubbling() => _setBubbling()

library/CM/Log/Handler/Abstract.php
```diff
@@ -0,0 +1,76 @@
...
+     */
+    public function getLevel() {
+        return $this->_level;
+    }
+
+    /**
+     * @param $level
+     * @throws CM_Exception_Invalid
+     */
+    public function setLevel($level) {
```
fvovan: Please add type for parameter

library/CM/Log/Handler/HandlerInterface.php
```diff
@@ -0,0 +1,20 @@
...
+
+interface CM_Log_Handler_HandlerInterface {
+
+    /**
+     * @return bool
+     */
+    public function getBubble();
+
+    /**
+     * @param int
```
fvovan: It's not a param type, it's return type

library/CM/Log/Handler/Sentry.php
```diff
@@ -0,0 +1,18 @@
+<?php
+
+class CM_Log_Handler_Sentry extends CM_Log_Handler_Abstract {
+
+    private $_client;
```
fvovan: Please add phpdoc for this attribute

library/CM/Log/Logger.php
zazabe: I just get the code from monolog, I think it can be useful to convert from level code from/to level name...

library/CM/Log/Logger.php
```diff
@@ -0,0 +1,194 @@
...
+     * @param CM_Log_Handler_HandlerInterface $handler
+     */
+    public function addHandler(CM_Log_Handler_HandlerInterface $handler) {
+        /** @var CM_Log_Handler_HandlerInterface $handler */
+        foreach ($this->_handlerList as $currentHandler) {
+            $currentHandler->setBubble(true);
+        }
+        $handler->setBubble(false);
+        $this->_handlerList[] = $handler;
+    }
```
fvovan: Perhaps i misunderstand the purpose of bubbling again, but as we decided to distinguish handlers and fallbacks it's not clear why we leave $_bubbling attribute? 
we have two arrays. Each record handles every record, and if something goes wrong with logging record just dispatched to fallback. wdyt?

library/CM/Log/Logger.php
```diff
@@ -0,0 +1,194 @@
...
+    /**
+     * Gets the name of the logging level.
+     *
+     * @param  integer $level
+     * @return string
+     * @throws CM_Exception_Invalid
+     */
+    public static function getLevelName($level) {
+        if (!isset(static::$levels[$level])) {
+            throw new CM_Exception_Invalid('Level "' . $level . '" is not defined, use one of: ' . implode(', ', array_keys(static::$levels)));
```
fvovan: we use backticks for variables output within exceptions messages, https://github.com/cargomedia/guides/tree/master/style
but still not sure if we need these methods, what if we just replace them with simple validation method?

library/CM/Log/Logger.php
zazabe: - is it a strong convention at CM to disallow default attribute value different from null ?

library/CM/Log/Context.php
```diff
@@ -0,0 +1,54 @@
+<?php
+
+class CM_Log_Context {
+
+    /** @var CM_Log_Context_ComputerInfo */
```
fvovan: Should be either marked as possible null here or made mandatory in constructor.

library/CM/Log/Logger.php
fvovan: Yes, it was  rather inconvenient for me too :). 
```$level = is_null($level) ? CM_Log_Logger::NOTSET :  (int) $level;``` would be even better.
Not sure if we want to use NOTSET, I would prefer to set it to some defined default value.

Comment (zazabe) [inline library/CM/Log/Logger.php]: `CM_Log_Logger::NOTSET` is the highest level, caught by all handlers... if we set the default to `0` for example, this log will have no change to be handled...

File: library/CM/Log/Logger.php (hunk 9 above)
Comment (fvovan) [inline library/CM/Log/Logger.php]: If we really need this functionality it will never be late to add it

File: library/CM/Log/Logger.php (hunk 10 above)
Comment (fvovan) [inline library/CM/Log/Logger.php]: well, why not use  some valid level as a default value (and it could be highest level)? now it is possible to use addRecord(..., NOTSET) what is obviously is not a good idea. Frankly my option is if even we decided not to follow PSR-3 strictly, we shouldn't add some custom values.

File: library/CM/Log/Handler/Database.php [hunk 19]
```diff
@@ -0,0 +1,15 @@
+<?php
+
+class CM_Log_Handler_Database extends CM_Log_Handler_Abstract {
+
+    private $_connection;
+
+    public function __construct($connection, $level = CM_Log_Logger::DEBUG, $bubble = false) {
```
fvovan: For now I would leave exactly 1 place where we handle defaults. CM_Log_Handler_Abstract::__construct().

library/CM/Log/Logger.php
```diff
@@ -0,0 +1,195 @@
...
+            }
+        }
+    }
+
+    /**
+     * @param string     $message
+     * @param int        $level
+     * @param array|null $options
+     */
+    public function addMessage($message, $level = null, array $options = null) {
```
fvovan: Do we really need this method together with all helpers?

library/CM/Log/Record.php
```diff
@@ -0,0 +1,45 @@
...
+    /** @var CM_Log_Context */
+    private $_context;
+
+    /**
+     * @param int            $level
+     * @param string         $message
+     * @param CM_Log_Context $context
+     */
+    public function __construct($level, $message, CM_Log_Context $context) {
+        $this->_level = $level;
```
fvovan: Please use type casting

library/CM/Log/Logger.php
zazabe: I think it's not a bad idea to keep the concept of bubbling for handlers, then we manage the specific logic of handlers/fallbacks at the logger level.

fvovan: Why add additional complexity and spread logic? I strongly argue against TMTOWTDI approach.

fvovan:
Review is more or less done (except tests, probably we should focus on conceptual things first, and they are dependent)

Most important notes are

1. Need to discuss Logger initialisation process and as a consequence Context retrieving and further passing in case we want to have it custom (probably for JS related exceptions).

2. It would be good to eliminate duplicate interfaces for HandlerAbstract. (And in general  to remove useless code if we really are not going to use it now)

3. And decide how to deal with defaults (a perhaps change NOTSET to some consistent value).

4. We should figure out how bubbling will designed. As for me since we decided to use separated arrays for 'general' and 'fallback' handlers there is no need for 'bubble' attribute for them.

zazabe:
1. I achieve this by defining a globalContext passed to the Logger constructor, then a logContext can be optionally passed to each log method, and the globalContext will be merged to it.
2. not sure to get your idea, which useless code? methods about log levels?
3. I'm not so happy with the NOTSET level set at 999, but we need a default level for all non `CM_Exception`, and it must be the highest level if we want that our handlers catch them.
4. Yes, we can "hardcode" this logic instead of using the bubbling concept... also with the current implementation, it will use fallbacks only if the last handler fail...

zazabe:
we can maybe replace the bubbling logic by this:
```
    public function addRecord(CM_Log_Record $record) {
        $handlerHasFailed = false;
        foreach ($this->_handlerList as $handler) {
            if (!$handler->handleRecord($record)) {
                $handlerHasFailed = true;
            }
        }
        if (empty($this->_handlerList) || $handlerHasFailed) {
            foreach ($this->_fallbackList as $handler) {
                if($handler->handleRecord($record)){
                    break;
                }
            }
        }
    }
```

In this case, fallbacks will be used if any handler fail to process the record, and it will make the all handler logic more clear.

zazabe:
About default level, I'll suggest these changes:
- set all non `CM_Exception` at the `CRITICAL` level...
- default handler level at `WARNING` instead of `DEBUG`
- remove the `NOTSET` level and make the `addMessage` level argument required

fvovan:
1. Sounds good for me but probably @njam has some thoughts and suggestions about it.
2. Yes, all these levels transformation stuff until we really need it. And also ```addMessage()``` should be removed in favour of helpers or at least extracted from interface (actually helpers could use ```addRecord()``` without it)
3. It would be better to introduce mapping [ ExceptionType => Severity ]. NONSET sounds very very general (actually means nothing). We will not be able to understand should we be worried or not.
4. Agree

zazabe:
@fvovan please review
I think we can start by implementing some handlers if you're ok with the current logger.
Maybe we can start with the most basic, StreamHandler (of FileHandler?), then work on the DatabaseHandler to re-implement the exceptionHandler with it.
@njam what do you think about that ?

njam:
> Maybe we can start with the most basic, StreamHandler (of FileHandler?), then work on the DatabaseHandler to re-implement the exceptionHandler with it.

Very good.

I would suggest to create a new branch and PR for that? This way it can be reviewed independently.

library/CM/Log/Logger.php
```diff
@@ -0,0 +1,217 @@
...
+    );
+
+    /** @var CM_Log_Handler_HandlerInterface[] */
+    private $_handlerList;
+
+    /** @var CM_Log_Handler_HandlerInterface[] */
+    private $_fallbackList;
+
+    /** @var CM_Log_Context */
+    private $_globalContext;
```
fvovan: It is very hard for me to follow this convention too, but 
>Chose systematic names reading from general to specific from left to right, rather than grammatically correct names. E.g. streamChannelMessageType or ratingAverage).

library/CM/Log/Logger.php
```diff
@@ -0,0 +1,217 @@
...
+        if (null === $handlerList) {
+            $handlerList = [];
+        }
+        if (null === $fallbackList) {
+            $fallbackList = [];
+        }
+        if (null === $globalContext) {
+            $globalContext = new CM_Log_Context();
+        }
+
```
fvovan: I'm not sure if it is good idea to have an empty default context,
later if we want to check it would be more difficult than just comparison to ```null```

library/CM/Log/Logger.php
```diff
@@ -0,0 +1,217 @@
...
+        $this->addMessage($message, self::CRITICAL, $context);
+    }
+
+    /**
+     * @param CM_Log_Context|null $context
+     * @return CM_Log_Context
+     */
+    protected function _mergeWithGlobalContext(CM_Log_Context $context = null) {
+        if (null === $context) {
+            $context = new CM_Log_Context();
```
fvovan: why not return $globalContext here

library/CM/Log/Logger.php
```diff
@@ -0,0 +1,217 @@
...
+    }
+
+    /**
+     * @param CM_Log_Record $record
+     */
+    public function addRecord(CM_Log_Record $record) {
+        $handlerHasFailed = false;
+        foreach ($this->_handlerList as $handler) {
+            if (!$handler->handleRecord($record)) {
+                $handlerHasFailed = true;
```
fvovan: This way we will dispatch to fallback even if certain record has level lower than required, should we?

library/CM/Log/Logger.php
```diff
@@ -0,0 +1,217 @@
...
+     * @param Exception           $exception
+     * @param CM_Log_Context|null $context
+     */
+    public function addException(Exception $exception, CM_Log_Context $context = null) {
+        $context = $this->_mergeWithGlobalContext($context);
+        $this->addRecord(new CM_Log_Record_Exception($exception, $context));
+    }
+
+    /**
+     * @param array $handlerList
```
fvovan: Please mention that it is array of Handlers here

library/CM/Log/Logger.php
```diff
@@ -0,0 +1,217 @@
...
+     * @param CM_Log_Handler_HandlerInterface $handler
+     */
+    public function addFallback(CM_Log_Handler_HandlerInterface $handler) {
+        $this->_fallbackList[] = $handler;
+    }
+
+    /**
+     * @return CM_Log_Handler_HandlerInterface[]
+     */
+    public function getHandlers() {
```
fvovan: Possibly obsolete method

library/CM/Log/Logger.php
```diff
@@ -0,0 +1,217 @@
...
+        return array_merge($this->_handlerList, $this->_fallbackList);
+    }
+
+    /**
+     * @param                     $message
```
fvovan: please add type to helper's docu

library/CM/Log/Logger.php
zazabe: right, I'll add a test and change the code, thanks!

library/CM/Log/Logger.php
zazabe: you mean a clone of the globalContext ? ok, it will optimize this a bit...

library/CM/Log/Record.php
```diff
@@ -0,0 +1,52 @@
...
+    /** @var CM_Log_Context */
+    private $_context;
+
+    /**
+     * @param                $level
+     * @param                $message
+     * @param CM_Log_Context $context
+     * @throws CM_Exception_Invalid
+     */
+    public function __construct($level, $message, CM_Log_Context $context) {
```
fvovan: Theoretically $context could be null here (if globalContext was null and no specific one was passed) . we should think how to deal with it.

library/CM/Log/Logger.php
fvovan: it is immutable now. Probably it will be safe just to return as is.

tests/library/CM/Log/ContextTest.php
```diff
@@ -0,0 +1,19 @@
...
+
+    public function testMergeContext() {
+        $user = new CM_Model_User();
+        $computerInfo = new CM_Log_Context_ComputerInfo();
+
+        $mainContext = new CM_Log_Context(null, null, $computerInfo, ['foo' => 10, 'bar' => 10]);
+        $targetContext = new CM_Log_Context($user, null, null, ['foo' => 20, 'foobar' => 42]);
+        $mergedContext = $mainContext->merge($targetContext);
+
+        $this->assertSame($user, $mergedContext->getUser());
```
fvovan: It is better to use assertEquals for reference comparison

library/CM/Log/Logger.php
zazabe: I think I will always except to have a global context, even if it will store nothing... so maybe the change could be to make it as a required constructor argument...

tests/library/CM/Log/ContextTest.php
```diff
@@ -0,0 +1,19 @@
+<?php
+
+class CM_Log_ContextTest extends CMTest_TestCase {
+
+    public function testMergeContext() {
+        $user = new CM_Model_User();
```
fvovan: I think it is better to use ```CM_Model_User::createStatic()``` here. Please check how it's done in other tests

tests/library/CM/Log/ContextTest.php
```diff
@@ -0,0 +1,19 @@
+<?php
+
+class CM_Log_ContextTest extends CMTest_TestCase {
+
```
fvovan: we should test constructors  too

tests/library/CM/Log/Handler/AbstractTest.php
```diff
@@ -0,0 +1,19 @@
+<?php
+
+class CM_Log_Handler_AbstractTest extends CMTest_TestCase {
+
```
fvovan: we should test contsructor here too, even if class is absrtact

library/CM/Log/Record.php
zazabe: contextGlobal can't be null anymore... and even before, i don't see where a record is instantiated without a context

library/CM/Log/Logger.php
```diff
@@ -0,0 +1,209 @@
...
+    );
+
+    /** @var CM_Log_Handler_HandlerInterface[] */
+    private $_handlerList;
+
+    /** @var CM_Log_Handler_HandlerInterface[] */
+    private $_fallbackList;
+
+    /** @var CM_Log_Context */
+    private $_contextGlobal;
```
njam: Not sure if it's needed here!
I see an advantage of "genericSpecific" mainly if there are multiple elements with the *same* "generic" (`contextLocal`, `contextGlobal`). In this case I think natural language (`globalContext`) is ok.

zazabe:
- I removed fallback, this concept is not required anymore, the exceptionHandler already has a fallback method, by writing exception raised by the logger into a file 
- we manage each handler exception by raising a aggregated exception when all have been processed, by this way we will not miss any of them and we can catch them in the exceptionHandler
- I also added a `name` argument to the abstract handler constructor, so we can easily identify each instance (like in the aggregated exception message)

library/CM/Log/HandlingException.php
```diff
@@ -0,0 +1,23 @@
...
+
+    /** @var  Exception[] */
+    private $_exceptionList;
+
+    /**
+     * @param string|null $message
+     * @param Exception[] $exceptionList
+     */
+    public function __construct($message = null, array $exceptionList) {
+        $this->_exceptionList = $exceptionList;
```
fvovan: please use type casting here

library/CM/Log/Logger.php
```diff
@@ -0,0 +1,183 @@
...
+        }
+        return $mergedContext;
+    }
+
+    /**
+     * Gets all supported logging levels.
+     *
+     * @return array Assoc array with human-readable level names => level codes.
+     */
+    public static function getLevels() {
```
fvovan: Is it used anywhere?

library/CM/Log/HandlingException.php
zazabe: casting is done by the parent class, is it not enough?

library/CM/Log/Logger.php
zazabe: no, but i like them and i want to keep them, simply because level code are not human friendly and we may have some case where we will not be able to use constant to access these values... 
also... it's so common on any logging lib that i will feel bad to not have it...

library/CM/Log/HandlingException.php
fvovan: Probably. It is better to ask @njam :)
as for me I try to add it everywhere to avoid situation for example if we need to add some statements with $message we can miss checking there

library/CM/Log/Logger.php
fvovan: I mean method call. I found only 1 usage at tests, but possibly missed something

zazabe: you didn't miss anything, I just supposed that it can be useful, as explained in my previous comment, but we can remove it also if you really think it's a problem

fvovan: we can add it when it will be useful.
Increasing of exposed methods number is evil, if they are static it is double evil

njam:
- Logger::addRecord: Using the handler's name as array key will result in overriding exception if two have the same name. I would just push to the array, with numerical index, wdyt?
- Logger::_mergeWithGlobalContext: Wouldn't it be easier to set the "local" context to `new CM_Log_Context()` in case it's omitted, and then always merge with the global one?
- CM_Log_Handler_Abstract::__construct: Rename `$level` to `$levelMin`? Then it will be clear what it is about without needing to read docu.
- CM_Log_Context_ComputerInfo: Let's not access this from global functions, but rather inject it! Better for testability, and functionality.
 - E.g. see `CM_Util::getFqdn()` to reliably get the FQDN
- CM_Log_Context::merge: Docu sounds likes it modifies the current instance, but it's creating a new instance

library/CM/Log/HandlingException.php
njam: I would suggest to check/parse any input on public APIs, like this constructor of a class. While it's currently only used in some places where these checks are already in place, that can change quickly. Checking on the "edges" helps making it more robust.
Unfortunately for arrays of some type this is not super straight forward with PHP. So I have mixed feelings.
Maybe just do? :>
```php
array_walk($exceptionList, function(Exception $e) {});
```
Alternatively
```php
array_walk($exceptionList, function($e) {
 if (!$e instanceof Exception) {
  throw new CM_Exception(...);
 }
});
```

zazabe:
@njam 
- Logger::_mergeWithGlobalContext: it's just a tiny optimisation, we don't really need to call the `merge` method if there's no local context... but i can revert it if you think it will be more readable
OK for the other points

njam:
> it's just a tiny optimisation

Did you benchmark this? :D I would use the more readable version here, as I guess speed difference will be completely neglectable?

zazabe:
@fvovan @njam ok, should be fine now, please review

library/CM/Log/Handler/HandlerInterface.php
```diff
@@ -0,0 +1,21 @@
+<?php
+
+interface CM_Log_Handler_HandlerInterface {
+
```
fvovan: Interface could be "smaller"
If we don't use ```getLevelMin()``` and ```isHandling ()``` outside of ```CM_Log_Handler_Abstract```'s hierarchy

njam:
- Logger::addRecord: Let's use `count()` instead of `sizeof()` (convention, they're [aliases](http://php.net/manual/en/function.sizeof.php), I didn't even know this existed)
- HandlerInterface: Agree, only `handleRecord()` and maaaybe `isHandling()` (probably not) should be needed

zazabe:
@njam it should be ok to merge it now

njam:
lgtm!

@fvovan agree to merge?

fvovan:
i do !

PR #1924, opened 2015-09-29
Replace serialization of form with getValue on individual fields in CM_Form_Abstract.getData

vogdb:


vogdb:
@njam please attention at Location::validate. I use `userInput` in `fromArray`.

njam:
Started to review:

CM_Form_Abstract:
- getValue: Should not use form anymore (infinite loop?). Maybe rather `getInput().val()`?
- getInputValue, getSelectValue, getArrayValue: On this abstract level maybe combine them (into `getValue()`?)

@tomaszdurka could you take over?

tomaszdurka:
The changes looks good in general.

Let's not add those helper methods: `getInputValue`, `getSelectValue` and `getArrayValue` on abstract FF. 
- We should define only common things in abstract not those which are used in some of them.
- Let's inline them instead. It's gonne be more clear for the subclasses too.
- You only use `getSelectValue` in one place anyway.
- You have special `checkbox` conditions in `getArrayValue` I guess that's especially for one of the usages (Set).

The best would be to make `getValue` and abstract, not implemented method.

vogdb:
@njam @tomaszdurka you still didn't answer about this
> please attention at Location::validate. I use `userInput` in `fromArray`.

Is it ok?

vogdb:
@tomaszdurka please review.

tomaszdurka:
All looks good!

vogdb:
@NicolasSchmutz it's your time to shine! Please test this issue and check wether Forms work fine. Try to write a blogpost, upload a photo, write a message and etc.

vogdb:
And of course try to sign up and convert to premium.

NicolasSchmutz:
@vogdb lvgtm!

tomaszdurka:
I will double-check set/getValue implementations of all the form-field classes:
- [ ] CM_FormField_Boolean
- [ ] CM_FormField_Captcha
- [ ] CM_FormField_Color
- [ ] CM_FormField_Date
- [ ] CM_FormField_File
- [ ] CM_FormField_FileImage
- [ ] CM_FormField_GeoPoint
- [ ] CM_FormField_Hidden
- [ ] CM_FormField_Integer
- [ ] CM_FormField_Set
- [ ] CM_FormField_Suggest
- [ ] CM_FormField_SuggestOne
- [ ] CM_FormField_Text
- [ ] CM_FormField_Textarea
- [ ] CM_FormField_TreeSelect
- [ ] CM_FormField_Url
- [ ] CM_FormField_Set_Select
- [ ] CM_FormField_Set_Select_Radio

tomaszdurka:
I've introduced `CM_FormField_Abstract.getEnabled`.
`Form.getData` now considers this method and only returns enabled values (as in commit https://github.com/cargomedia/cm/commit/4a52bcd9851a38c39a15456a337fb0db47810813).

tomaszdurka:
@vogdb @njam wanna take a look?

njam:
Looking at the code:
- CM_Form_Abstract::getData: I understand we ignore fields that have value "" (empty string). Is that expected?
- CM_Form_Abstract::getData: If `actionName` is not passed then we will return an empty array. I think we can remove that case completely now (previously was used by `Field::getValue`)?
- CM_FormField_Abstract::getEnabled: Could use jQuery [`is()`](http://api.jquery.com/is/) ?
- CM_FormField_Abstract::getInput: How about throwing an error if we can't find an input element?
- CM/FormField/Boolean.js: Should use `prop()` instead of `attr()` (`.prop('checked', state);`)
- CM/FormField/Captcha.js: Reuse `getInput()` in other methods


BTW should we try adding all form fields to the example form? Would be useful for testing?

tomaszdurka:
@njam thanks for the input. I've actually made some changes in midtime. 
- ?
- Agree. I removed it already.
- Yes
- We can. I override this method in some places. We would need to throw everywhere then. Maybe we should rather throw in usages?
- Yes! Actually was testing checkboxes in `CM_FormField_Set` and was wandering why `attr` doesn't work.
- OK.

njam:
> CM_FormField_Abstract::getInput: How about throwing an error if we can't find an input element?

I meant to throw only in the abstract function, so that if a child class doesn't implement this properly we know about it. wdyt?

tomaszdurka:
Agree.

vogdb:
@tomaszdurka Looks good except one thing. Why do we have `this.getInput().trigger('change');` in `CM_FormField_Set_Select::setValue()`?

tomaszdurka:
@vogdb seems that when you set value of an select with jquery `val()` no change event on select is triggered. We need that event for `fancyselect` plugin (otherwise label won't update - check https://github.com/cargomedia/cm/blob/master/client-vendor/after-body/jquery.fancySelect/jquery.fancySelect.js)

tomaszdurka:
Hmm this will trigger also form field's change event. Which is probably fine, but I noticed that in some cases we don't want to trigger form field change when setting value, e.g. in `CM_FormField_Text` we do so...:
```js
  setValue: function(value) {
    this._skipTriggerChange = true;
    this.getInput().val(value);
    this._skipTriggerChange = false;
  },
```
Not sure what should be general behaviour - trigger or not to trigger...

@njam you have any memories about this?

njam:
Generally iirc setting a form field's input programmatically should *not* trigger any `change` events.

So I would rather change `fancySelect`. Maybe `watch` for changes on the value? Or expose a function to update the fancySelect?

vogdb:
> @vogdb seems that when you set value of an select with jquery `val()` no change event on select is triggered.

I've just tried on CM_Form_Example and it works without additional `this.getInput().trigger('change');`

vogdb:
@njam 
> - CM/FormField/Captcha.js: Reuse `getInput()` in other methods

Is it ok if the reuse is possible only as this:
```js
  getInput: function() {
    return this.$("input[name*=id], input[name*=value]");
  },

  /**
   * @param {{id: *, value: *}} captcha
   */
  setValue: function(captcha) {
    var $input = this.getInput();
    $input.eq(1).val(captcha.id);
    $input.eq(2).val(captcha.value);
  }
```

njam:
@vogdb sorry I didn't look closely, that's obviously not a good idea, let's leave it as is!

tomaszdurka:
@vogdb try running `cm.findView('CM_Form_Abstract')._fields.setSelect2.setValue(2);` without and with this trigger.

tomaszdurka:
@njam 
Regarding listening on the select/fancyselect value change:
- There is no easy way to listen on val() method. We could override this method, or override hooks for select for this method especially for fancyselect. Unfortunately I did not find a way to change `val()` behaviour for specific elements (e.g. fancyselect ones).
- I've also didn't like expsoing fancyselect method as it would make FF dependant on / know about fancyselect.
- Decided to use special event 'value-change' which could be used for cases like these - where value changes, but we don't want to trigger dom element change event.

tomaszdurka:
As discussed with Reto let's use `fancyselect:update` event instead of `value-change` one.

njam:
As discussed in person, let's keep it fancySelect specific (easier to understand). We'll use an event like `fancySelect:update` on the DOM element.

vogdb:
Lets revert the latest Captcha changes. The rest is lgtm.

vogdb:
lgtm

PR #1928, opened 2015-09-30
Fix issue when the redis client is used within the subscribe closure

zazabe:


zazabe:
@njam please review

njam:
- _defineConnection: I would suggest that this function returns an array, so it doesn't have "side effects", and the control flow is simpler
 - Call it `parseConfig`
 - Rename instance variable to `_config?

zazabe:
@njam please review

njam:
lgtm

PR #1929, opened 2015-09-30
php-style parsing of arrays in url-parameters in frontend

alexispeter:
urlParams like `/messages?conversation[id]=1&conversation[type]=2&foo=bar` should be parsed as 
```js
{
  conversation: {id: 1, type; 2},
  foo: bar  
}
```
so state-changes can be handled naturally when working with 2-dimensional parameters.

### DIFF
--- a/library/CM/App.js
+++ b/library/CM/App.js
@@ -1266,7 +1266,7 @@ var CM_App = CM_Class_Abstract.extend({
      */
     parseQueryParams: function(queryParams) {
       var params = queryString.parse(queryParams);
-      var arrayParamRegex = /^([\d\w]+)\[([\d\w]+)]$/;
+      var arrayParamRegex = /^(\w+)\[(\w+)]$/;
       _.each(params, function(value, key) {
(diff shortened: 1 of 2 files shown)


alexispeter:
please review @vogdb 

fyi @tomaszdurka @njam

alexispeter:
Doesn't work on arrays with more than one dimension.

vogdb:
- About regex. `\d` should be included in `\w` but please verify.
- `result` var should be declared.
- missing semicolon after `params[paramName] = {}`

alexispeter:
please review @vogdb

vogdb:
Sorry. About regex again. You escape the left `[` but not the right `]`. Why?

alexispeter:
Seems `]` is only considered a special character when there is an opening `[` somewhere in front of it.
phpstorm says `redundant character escape` if I put it there. And it works :>

vogdb:
:D ok then. lgtm. does not hurt))

PR #1931, opened 2015-10-01
Logger - StreamHandler support

zazabe:


zazabe:
@njam @fvovan The first handler is here! please review

note: this ticket is related to #1926

njam:
How about using `CM_OutputStream_Abstract` - it has `write()`

zazabe:
@njam @fvovan I don't feel so comfortable with the current formatting implementation:
- no way to have optional part
- each part have to be explicitly defined in the `formatRecord` method

I thought we can use `Smarty` for this, and assign the LogRecord directly as a template variable, much more flexible, but maybe too heavy just for formatting log messages... 
Another idea is to add a lighter template engine, like Mustache...

what do you think about that? 

Note: In most Logger I've seen, the record formatting is externalised and not managed by the handler directly, but it will be maybe over-complicated in our use case.

njam:
I would suggest to keep this as simple as possible, for two reasons:
- Reduce the probability that the logging itself will fail
- We have very limited use case for this handler so far - only as a fallback handler

The current templating looks pretty good to me. I would additionally display all the context information (maybe as `- key: value` lines below the message), including the exception (reuse `CM_ExceptionHandling_Formatter_Abstract`?). wdyt?

fvovan:
For me it is overengineering 
why not use plain sprintf ?

zazabe:
@fvovan i think `printf` only support replacement based on the argument order, not based on a key name for example like the primitive `preg_replace_callback` I'm using currently...
It's more a suggestion, we maybe don't really need more part in our logs.

zazabe:
The log formatting is taking the context/record exception into account now, like that:
```
[2015-10-02T15:39:59+00:00 - INFO] Hello World!
[2015-10-02T15:39:59+00:00 - ERROR] Super cool!
 - exception: 
   - message: Super cool!
   - type: Exception
   - stacktrace: 
     00. {main} /home/vagrant/cm/scripts/test.php:0
```

- I'm still not so satisfied by hardcoding the formatting in this large `_formatRecord` method, but as discussed, it will be mainly used as a fallback method... so it should be acceptable
- I removed the `ComputerInfo` log line for the `StreamHandler`, looks strange to have always the same line in the log file, but I supposed that the StreamHandler will be configured with a local file in output, which is maybe not always the case...

njam:
I would keep `ComputerInfo` in the output, as you mention it might not be a local FS?

zazabe:
@njam yes, I think the best is to have it on one line, like:
```
[2015-10-02T15:39:59+00:00 - foo.com - php 5.4 - INFO] Hello World!
```
So we will not have 2 lines log like:
```
[2015-10-02T15:39:59+00:00 - INFO] Hello World!
- computerInfo: fqdn: foo.com, phpVersion: 5.4
[2015-10-02T15:40:59+00:00 - INFO] Hello Basel!
- computerInfo: fqdn: foo.com, phpVersion: 5.4
[2015-10-02T15:41:59+00:00 - INFO] Hello Claraplatz!
- computerInfo: fqdn: foo.com, phpVersion: 5.4
```
The `fqdn` and the `phpVersion` are already available as part of the message, i just have to change the default log message formatting string.

njam:
sgtm!

zazabe:
@njam please review

njam:
- CM_Log_Formatter_Html: Should escape text against HTML characters (see `CM_Util::htmlspecialchars()`)
- CM_Log_Factory::createLogger: Docu for `$handlerList` could say `string[]` instead of `array` - then it'd  be clear that those are names?
- CM_Log_Factory::createLogger: Use `Functional\map()` to get from handler names to handler instances?
- CM_Log_Handler_Stream::_writeRecord: I think we should only call 1 `write()`  call here, to avoid multiple processes' output being mixed
- CM_Log_Handler_Stream: Apply formatting

zazabe:
@njam all changes done, please review

njam:
- CM_Log_Factory: Retrieving the serviceManager from the global singleton is an anti-pattern. It's used in many places because the serviceManager isn't fully integrated into CM yet. Let's try to avoid it for new code, especially for code that's so closely coupled *to* the serviceManager. Maybe the following would be good?:
 - Make the factory implement `CM_Service_ManagerAwareInterface` (using `CM_Service_ManagerAwareTrait`).
 - Change `CM_Service_Manager::_instantiateService()` to call `setServiceManager()` *before* invoking the factory method, so that the serviceManager is available
- CM_Log_Formatter_Html: Let's avoid inline-styles.. (might not be expected in some context) We can always style it in the page context or so?

Rest looks good

zazabe:
> CM_Log_Formatter_Html: Let's avoid inline-styles.. (might not be expected in some context) We can always style it in the page context or so?
For exception caught at an early stage during the bootstrapping, we will not have the page context i think, just writing on the output...

zazabe:
@njam please review

njam:
I think it's still better to *not* have inline styles, as it can mess up layouts later on. In case there's no wrapping element I guess the default browser styles should be fine.

lgtm!

PR #1932, opened 2015-10-02
Make CLI runnable serviceManager aware

njam:


njam:
Agree, it's a small step, but will help avoid singleton access in applications

PR #1908, opened 2015-10-03
Layout.loadPage(): $pagePlaceholder might not be set

vogdb:
In the error handling of `CM_Layout_Abstract`'s `loadPage()` we modify `_$pagePlaceholder`:
```js
        .catch(function(error) {
          window.clearTimeout(layout._timeoutLoading);
          if (!(error instanceof Promise.CancellationError)) {
            layout._$pagePlaceholder.addClass('error').html('<pre>' + error.message + '</pre>');
            layout._onPageError();
            throw error;
          }
        });
```

The problem is that `_$pagePlaceholder` might have been set to `null` in the `then()` block above, so this will fail.

@vogdb could you look into this when you have some spare time? I have the feeling this code in general could use some love, it's quite difficult to read (e.g. logic spread between `loadPage` and `_loadPageThrottled`).

### DIFF
--- a/library/CM/Layout/Abstract.js
+++ b/library/CM/Layout/Abstract.js
@@ -21,9 +21,12 @@ var CM_Layout_Abstract = CM_View_Abstract.extend({
 
     var layout = this;
     this._loadPageThrottled = promiseThrottler(function(path) {
+      layout._createPagePlaceholder();
+      layout._chargeSpinnerTimeout();
+
       return layout.ajaxModal('loadPage', {path: path})
         .then(function(response) {
-          window.clearTimeout(layout._timeoutLoading);
+          layout._clearSpinnerTimeout();
           if (response.redirectExternal) {
             cm.router.route(response.redirectExternal);
             return;
@@ -34,21 +37,16 @@ var CM_Layout_Abstract = CM_View_Abstract.extend({
             window.location.replace(response.url);
             return;
           }
-          layout._$pagePlaceholder.replaceWith(view.$el);
-          layout._$pagePlaceholder = null;
-          var fragment = response.url.substr(cm.getUrl().length);
-          if (path === fragment + window.location.hash) {
-            fragment = path;
-          }
-          window.history.replaceState(null, null, fragment);
+          layout._removePagePlaceholder(view.$el);
+          layout._updateHistory(path, response.url);
           layout._onPageSetup(view, response.title, response.url, response.menuEntryHashList, response.jsTracking);
           view._ready();
           return view;
         })
         .catch(function(error) {
-          window.clearTimeout(layout._timeoutLoading);
+          layout._clearSpinnerTimeout();
           if (!(error instanceof Promise.CancellationError)) {
-            layout._$pagePlaceholder.addClass('error').html('<pre>' + error.message + '</pre>');
+            layout._errorPagePlaceholder(error);
             layout._onPageError();
             throw error;
           }
@@ -80,17 +78,6 @@ var CM_Layout_Abstract = CM_View_Abstract.extend({
    */
   loadPage: function(path) {
     cm.event.trigger('navigate', path);
-
-    if (!this._$pagePlaceholder) {
-      this._$pagePlaceholder = $('<div class="router-placeholder" />');
-      this.getPage().replaceWithHtml(this._$pagePlaceholder);
-      this._onPageTeardown();
-    } else {
-      this._$pagePlaceholder.removeClass('error').html('');
-    }
-    this._timeoutLoading = this.setTimeout(function() {
-      this._$pagePlaceholder.html('<div class="spinner spinner-expanded" />');
-    }, 750);
     return this._loadPageThrottled(path);
   },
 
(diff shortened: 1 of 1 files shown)


vogdb:
@njam please review.
@zazabe please test.

zazabe:
@vogdb  the SearchContent form still produce an exception:
```
> Uncaught TypeError: Cannot read property 'addClass' of null
CM_View_Abstract.extend._errorPagePlaceholder	@	library.js:3862
(anonymous function)	@	library.js:3758
tryCatcher	@	after-body.js:15858
16.module.exports.Promise._settlePromiseFromHandler	@	after-body.js:15016
16.module.exports.Promise._settlePromiseAt	@	after-body.js:15092
16.module.exports.Promise._settlePromises	@	after-body.js:15208
1.Async._drainQueue	@	after-body.js:13004
1.Async._drainQueues	@	after-body.js:13014
1.exports.drainQueues	@	after-body.js:12896
```

But it's just the result of another exception caught earlier: `
TypeError: Cannot read property 'replaceWith' of null at CM_View_Abstract.extend._removePagePlaceholder`

The call to `_removePagePlaceholder()` is done in this piece of code: 
```
return layout.ajaxModal('loadPage', {path: path})
        .then(function(response) {
          layout._clearSpinnerTimeout();
          if (response.redirectExternal) {
            cm.router.route(response.redirectExternal);
            return;
          }
          var view = layout._injectView(response);
          var reload = (layout.getClass() != response.layoutClass);
          if (reload) {
            window.location.replace(response.url);
            return;
          }
          layout._removePagePlaceholder(view.$el);
          layout._updateHistory(path, response.url);
          layout._onPageSetup(view, response.title, response.url, response.menuEntryHashList, response.jsTracking);
          view._ready();
          return view;
        })
        .catch(function(error) {
          layout._clearSpinnerTimeout();
          if (!(error instanceof Promise.CancellationError)) {
            layout._errorPagePlaceholder(error);
            layout._onPageError();
            throw error;
          }
        });
```

To reproduce it, you can write in the search form multiple times.
I suppose the exception happen when multiple ajax requests are received at the same time with a placeholder already removed.

njam:
I'm waiting with review then

vogdb:
I couldn't reproduce the bug but ok. Lets try this.

vogdb:
@zazabe please test

zazabe:
@vogdb still have the issue

```
Uncaught TypeError: Cannot read property 'replaceWith' of null
CM_View_Abstract.extend._removePagePlaceholder	@	library.js:3854
(anonymous function)	@	library.js:3749
tryCatcher	@	after-body.js:15858
16.module.exports.Promise._settlePromiseFromHandler	@	after-body.js:15016
16.module.exports.Promise._settlePromiseAt	@	after-body.js:15092
16.module.exports.Promise._settlePromises	@	after-body.js:15208
1.Async._drainQueue	@	after-body.js:13004
1.Async._drainQueues	@	after-body.js:13014
1.exports.drainQueues	@	after-body.js:12896
```

It happens here, has described in my previous comment: 
```
  _removePagePlaceholder: function(el) {
    this._$pagePlaceholder.replaceWith(el);
    this._$pagePlaceholder = null;
  },
```

njam:
This whole approach with `_$pagePlaceholder` being access throughout this class is a bit problematic. We never know for sure if it's currently available or not (or it's very difficult to understand why).
@zazabe maybe you have some ideas how to restructure this to make it simpler?

zazabe:
@njam I need to investigate more about how this pagePlaceholder is used, but maybe simple having a `getPagePlaceholder()` that ensure the element exists will help.

vogdb:
We also can split the execution of `loadPage` success 'then' on two 'then'. The first is executed when pagePlaceholder exists. Another is executed after pagePlaceholder was replaced by the page. Unfortunately it does not simplify the code at all but should exclude the situation of `null`:
```js
      layout._createPagePlaceholder();
      layout._chargeSpinnerTimeout();

      return layout.ajaxModal('loadPage', {path: path})
        .then(function(response) {
          layout._clearSpinnerTimeout();
          if (response.redirectExternal) {
            cm.router.route(response.redirectExternal);
            return;
          }
          var view = layout._injectView(response);
          var reload = (layout.getClass() != response.layoutClass);
          if (reload) {
            window.location.replace(response.url);
            return;
          }
          layout._updateHistory(path, response.url);
          layout._removePagePlaceholder(view.$el);
          return {view: view, response: response};
        })
        .catch(function(error) {
          layout._clearSpinnerTimeout();
          if (!(error instanceof Promise.CancellationError)) {
            layout._errorPagePlaceholder(error);
            layout._onPageError();
          }
          throw error;
        }).then(function(data) {
          var view = data.view;
          var response = data.response;
          layout._onPageSetup(view, response.title, response.url, response.menuEntryHashList, response.jsTracking);
          view._ready();
          return view;
        }).catch(function(error){
          if (!(error instanceof Promise.CancellationError)) {
            layout._createPagePlaceholder();
            layout._errorPagePlaceholder(error);
          }
          throw error;
        });
    }, {cancelLeading: true});
```

`getPagePlaceholder` seems much better even it's more of a patch rather than a complete fix =(

vogdb:
@zazabe please try this one.

zazabe:
@vogdb seems fixed, lgtm!

Note: I notice some strange behaviour with the `SearchContent` component, new ticket created: cargomedia/sk#2167

vogdb:
@njam please review then.

njam:
Nice refactoring!

- initialize: How about *not* creating the placeholder here at all, as it will be created lazily later on anyway?
- _updateHistory: Agree to extract this to a function, but out of context the different arguments to the function are quite difficult to understand. How about something like?:
 - `path` -> `requestPath` - the path that we requested originally by the client
 - `url` -> `responseUrl` - the URL that the server responded with
 - `fragment` -> `responseFragment`

vogdb:
@njam please review

njam:
lgtm

@zazabe wanna do another acceptance test? Or you already tested the full functionality?

zazabe:
@njam the only test i did is about the `SearchContent` issue and I can't reproduce it manually anymore...

PR #1933, opened 2015-10-06
Breaking: Enable MaxMind through the service manager

fauvel:
Reason:
```
CM_Exception: City `London` (1) has no MaxMind code in /home/vagrant/sk/vendor/cargomedia/cm/library/CMService/MaxMind.php on line 880
```

In the clockwork this code is registered like this:
```php
        $this->_registerClockworkCallbacks('8 days', array(
            'CMService_MaxMind::upgrade' => function () {
                try {
                    $maxMind = new CMService_MaxMind();
                    $maxMind->upgrade();
```

Probably for dev machines this shouldn't even run?
How about registering "MaxMind" as a service in the serviceManager, and in the clockwork first checking if it has been registered?
For a blueprint see https://github.com/njam/sk/commit/88e22b7cdd23a59fa4354398756a186af133a3c6

### DIFF
--- a/library/CM/Location/Cli.php
+++ b/library/CM/Location/Cli.php
@@ -8,7 +8,7 @@ class CM_Location_Cli extends CM_Cli_Runnable_Abstract {
      */
     public function outdated($verbose = null) {
         /** @var CMService_MaxMind $maxMind */
-        $maxMind = CM_Service_Manager::getInstance()->get('maxmind', 'CMService_MaxMind');
+        $maxMind = $this->getServiceManager()->get('maxmind', 'CMService_MaxMind');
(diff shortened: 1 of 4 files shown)


fauvel:
@njam please review

fauvel:
@njam There are basically two changesets here:
- I'm ignoring cities without MaxMind code for the sake of upgrading the geolocation database, so that we won't have any more errors in dev;
- I'm relying on the configuration of the service manager to decide whether to upgrade periodically the MaxMind database or not, so that the upgrade won't run in dev anyway. The config has to be added in production, though (PR already done).

njam:
lgtm

fauvel:
Tagged 1.152.0

PR #1934, opened 2015-10-06
Add support of Handler bubbling

zazabe:


zazabe:
I think it's good to implement the handler bubbling again, it will be useful for the basic Logger used by `ExceptionHandling`, we can declare a minimal stderr/output handler and then add a file handler with `bubble = false` when the filesystem service is available, after that, the stderr/output handler will be used only if the file handler has failed.

FYI, the basic logger will be used by in `ExceptionHandling`, when the normal logger fail, the code will look like this:
```
    public function handleException(Exception $exception) {
        try {
            $this->getServiceManager()->getLogger()->addException($exception);
        } catch (CM_Log_HandlingException $loggerException) {
            $this->_loggerBasic->error('Origin exception:');
            $this->_loggerBasic->addException($exception);
            $this->_loggerBasic->error('Handlers exception:');
            $this->_loggerBasic->error($loggerException->getMessage());
            foreach ($loggerException->getExceptionList() as $handlerException) {
                $this->_loggerBasic->addException($handlerException);
            }
        } catch (Exception $loggerException) {
            $this->_loggerBasic->error('Origin exception');
            $this->_loggerBasic->addException($exception);
            $this->_loggerBasic->error('Logger exception');
            $this->_loggerBasic->addException($loggerException);
        }
    }
```

@fvovan @njam  what do you think about that? could you review this if we are agree to have the bubbling again?

fvovan:
What's the advantage over discussed earlier design?

zazabe:
@fvovan the idea is to have 2 backup handlers, but with the second as a fallback...

actually, we can easily reproduce the same behaviour than what we currently have in prod:
- try to log in db
- if failed: try to log in stdout (cli)/output (http)
- if failed: log in a file

Note: the description of the process is wrong in my previous comment if we want to reproduce the same behaviour, the BasicLogger with have the std/output handler without bubbling, then the file handler as a fallback.

fvovan:
Not sure if it is good idea to have several fallbacks
We'll have to check them one by one to find needed record.

njam:
Current logic is to try these ways of logging, one after the other and maximum one, right?
1. Write to DB (`logException()`)
2. Write to File  (`logException()`)
3. Write to STDOUT (`handleException()`)

I think we can simplify it to:
1. Write to DB
2. Write to File *and* STDOUT

This is still good, and much simpler to model. We can just have a backup logger with both handlers (STDOUT and file).
wdyt?

fvovan:
Actually for emergency purposes we could leave only stream (put it to STDERR and copy it to file if needed via &2 > file.log )

zazabe:
@njam ah yes, `_printException` (stdout/output) is executed only if logException has failed... 
But anyhow, our model is very flexible now, we can register handlers to the backupLogger has we want, with the fallback or not.

@fvovan yes but it's better to use a specific handler, not sure how to do the redirect for the cgi entrypoint (nginx can certainly log errors...) and by having a specific handler, we can have different format for each handler.

I think the idea of the old backup implementation is to "silently" log exception into a file, and if it fail, send a 500 error response.
Using the backupLogger means already that something bad happen, so removing the fallback mechanism and sending a 500 error response all the time seems better.

So a `BackupLogger` with an `OutputHandler` and a `FileHandler`, with the bubbling enabled for both ?

njam:
> So a `BackupLogger` with an `OutputHandler` and a `FileHandler`, with the bubbling enabled for both ?

sgtm. But then we wouldn't need the "bubbling" at all? Because it's just a logger with those two handlers?

zazabe:
@njam yes, finally, the bubbling is not required anymore, but it provides more flexibility and I think it's worth to keep it.

njam:
- CM_Log_Handler_Abstract::construct: We usually try making the default of booleans `false` (convention to make it easier to understand default values), here `$bubble` has default `true`. Should we rename it? I'm still not 100% sure it's good to add this functionality if we don't need it. It creates additional complexity also about the *order* of handlers (`addHandler`'s `$prepend`). What would be a good use case?
- Logger::addRecord: How about moving the `if (isHandling())`-check up from `handleRecord` to `addRecord`, this would allow to have all this logic together. Or is there a reason I'm missing why it's in `handleRecord`?
- Logger::addRecord: Shouldn't this check the return value of `handleRecord` instead of relying on `isHandling()` and exception throwing? Seems we have now 3 mechanism for the same concept?

zazabe:
> CM_Log_Handler_Abstract::construct: We usually try making the default of booleans false (convention to make it easier to understand default values), here $bubble has default true. Should we rename it? I'm still not 100% sure it's good to add this functionality if we don't need it. It creates additional complexity also about the order of handlers (addHandler's $prepend). What would be a good use case?

About the bubbling, it gives us some flexibility in the configuration, but also some complexity... at the beginning, I was not agree about adding this feature, it can produce some misunderstanding because the bubbling rule is based on multiple factors (handled record, failed handler, bubbling value,...).   
Later, when i thought we need to reproduce the same behaviour for the backupLogger than what we have in prod, i thought a fallback system based on the bubbling is a good solution.  
Then, we decided to simplify the backupLogger, so the fallback was not useful anymore,....
Finally, I'm still not convinced by the bubbling system but I already added/removed it 3 times, so maybe this feature still worth something... or not.

> Logger::addRecord: How about moving the if (isHandling())-check up from handleRecord to addRecord, this would allow to have all this logic together. Or is there a reason I'm missing why it's in handleRecord?

From my point of view, the logger just "propagate" a record to some handlers, it's the responsibility of the handler to know if a record has to be processed or not.   
If we move the `isHandling` call outside, we will not be able to use the `handleRecord` method directly without checking first if the record is handled, which is not really obvious.

> Logger::addRecord: Shouldn't this check the return value of handleRecord instead of relying on isHandling() and exception throwing? Seems we have now 3 mechanism for the same concept?
 
Not sure what you mean by `3 mechanism for the same concept` but maybe it's because name used are quite confusing, maybe `Logger::addRecord` should be renamed into `Logger::propagateRecord`.
Also, the complexity of this method is mainly coming from the bubbling logic, which is confusing and may be dropped probably...

My suggestion:
- drop the bubbling feature
- keep `isHandling` into `Handler:handleRecord`
- rename `Logger::addRecord` into `Logger::propagateRecord` to clarify the propose of this method

njam:
sgtm

NicolasSchmutz:
Reassigning to @fvovan 
Maybe closeable?

cc @tom

fvovan:
decided not to use bubbling 
https://github.com/cargomedia/cm/pull/2081

PR #1935, opened 2015-10-06
Breaking: Logger integration into the CM Framework

zazabe:


zazabe:
This is a draft integration of the Logger into the CM framework, i still have to fix some test using the old `CM_ExceptionHandling_Handler_Cli` but the idea is here...

@fvovan @njam could you have a look and check if I'm on the right way? ;)

njam:
- CM_Bootloader: This upgrade dance of the basic logger is quite complicated, because the control flow is not super straight forward. How about instantiating the basic logger *every time it is accessed*? Then we could just check if the filesystem is already registered and depending on that return either "STDOUT logger" or "file logger". Since this logger should only be used in edge cases the performance impact should be neglectable.
- CM_ExceptionHandling_Handler: Maybe call it "backup" instead of "basic"?
- CM_ExceptionHandling_Handler: We lost the code to send the exception to Newrelic

zazabe:
@njam please review

- CM_Bootloader: agreed, I did this because I'm still thinking about having this log facility available at an early stage, not only of the exception handler... but we don't really need this actually...
- from `basic` to `backup`: done, this name make more sense if the logger is only used by the exception handler
- Newrelic: I know, I though we can use a specific log handler for it

zazabe:
@njam please review
Note: all SK tests passed too ;)

fvovan:
may I commit to this task or would it be better to create new PR based on this branch?

njam:
Either way is good I think, what you prefer.

fvovan:
closed in favour of https://github.com/cargomedia/cm/pull/2081

PR #1937, opened 2015-10-07
Logger - NewrelicHandler support

zazabe:


zazabe:
@njam please review
- not sure about adding the `NewrelicHandler` in the `default.php` cm config...

zazabe:
also waiting for the review of #1935...

njam:
Generally looks good, one thing:
- CM_Log_Record_Exception: We should be able to construct this record even if we don't have an `Exception` instance (for when we log errors from the frontend (see `CM_Http_Response_JsError`)). Should we make the "original exception" optional?

zazabe:
@njam ok but why not creating a specific Exception from the js error on the server side then simply call `ExceptionHandler::handleException`?

Also, It could be interesting to have the full stacktrace if the Error object is passed to the `window.onerror` event handler ([seems not supported by all browsers](https://danlimerick.wordpress.com/2014/01/18/how-to-catch-javascript-errors-with-window-onerror-even-on-chrome-and-firefox/)), but it may introduce some complexity... ([cross-browser stacktrace parsing](http://www.stacktracejs.com/#!/docs/error-stack-parsing), replace the script injection by an ajax call to do the request,...).

zazabe:
@njam I introduce the `originalExeption` because `CMService_Newrelic::setNoticeError` require an `Exception`, but maybe we don't need to convert exception into `CM_ExceptionHandling_SerializableException` in `CM_Log_Record_Exception`, Handlers should know how to convert exceptions in the format they need.

So what about:
- keep only the original Exception in `CM_Log_Record_Exception`
- create a `CM_Exception_Js` built based on parameters received from the client and use the ExceptionHander to record it.

njam:
afaik there's no way to create a PHP `Exception` with a specific stack trace, it will always use the current stack. That's why we wanted to use `CM_ExceptionHandling_SerializableException`, because we can control all of its attributes completely.

So how about allowing to pass *either* `Exception` or `CM_ExceptionHandling_SerializableException`, and making the *originalException* field optional?

zazabe:
@njam why not making our `CM_ExceptionHandling_SerializableException` compatible with the `Throwable` interface, then we will be able to hint `CM_Log_Record_Exception` / `CMService_Newrelic` arguments based on that...

njam:
> Throwable is the base interface for any object that can be thrown via a throw statement in **PHP 7**, including Error and Exception.
> PHP classes cannot implement the Throwable interface directly, and must instead extend Exception.
http://php.net/manual/en/class.throwable.php

:|

fauvel:
Closed in favor of #2894

PR #1938, opened 2015-10-08
Introduce getData (without action) on CM_Form_Abstract

tomaszdurka:


tomaszdurka:
Paired with @vogdb 

@njam fyi

PR #1943, opened 2015-10-10
Follow-up of Form.getData.

vogdb:


vogdb:
@njam please review.

njam:
functionality lgtm

Disabled "integer" field looks strange:
screen shot 2015-10-10 at 13 04 00

vogdb:
Do we need to drop value to `0` ?

njam:
No I think we should keep the value, just meant that the left part of the bar should be visible too.
We can merge this though, and open a followup.

PR #1945, opened 2015-10-10
Fix File::getValue()

njam:

The server expects always an array (runs `array_filter` on the user input). Plus we're considering disabled/enabled state on submit now, in the form.


vogdb:
lgtm

PR #1948, opened 2015-10-10
Fix FormField getValue

njam:


njam:

..merging to fix the bugs

PR #1958, opened 2015-10-16
Introduce media streams repository

tomaszdurka:
Introduce stream repository to gather actions to create/remove streams (publish, subscribe) and streamChannels.

### DIFF
--- a/library/CM/Streaming/MediaStreamRepository.php
+++ b/library/CM/Streaming/MediaStreamRepository.php
@@ -1,0 +1,85 @@
+<?php
+
+class CM_Streaming_MediaStreamRepository {
+
+    /** @var int */
+    protected $_adapterType;
+
+    /**
+     * @param int $adapterType
+     */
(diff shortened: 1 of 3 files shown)


tomaszdurka:
@fvovan please review

library/CM/Streaming/MediaStreamRepository.php
```diff
@@ -0,0 +1,85 @@
...
+            'user'          => $user,
+            'start'         => (int) $start,
+            'key'           => (string) $clientKey,
+        ]);
+    }
+
+    /**
+     * @param CM_Model_Stream_Abstract $stream
+     */
+    public function removeStream(CM_Model_Stream_Abstract $stream) {
```
tomaszdurka: Yes I know it's kinda strange to have this specific logic about removing streamChannel in case no other streams are there, but that's how we use it most of the time.

fvovan:
lgtm

PR #1960, opened 2015-10-19
Remove Adapter concept from Video/MediaStream, replace CM_Video/Media_Service with CM_Wowza

tomaszdurka:


tomaszdurka:
@fvovan please review this one

library/CM/Wowza/HttpClient.php
```diff
@@ -0,0 +1,39 @@
+<?php
+
+class CM_Wowza_HttpClient {
```
tomaszdurka: Should this be named `CM_Wowza_HttpApi`? :o

fvovan: I would rather prefer ```CM_Wowza_HttpApiClient```

library/CM/MediaStreams/StreamRepository.php
```diff
@@ -13,6 +13,13 @@ public function __construct($adapterType) {
     }
 
     /**
+     * @return CM_Paging_StreamChannel_AdapterType
+     */
+    public function getStreamChannels() {
+        return new CM_Paging_StreamChannel_AdapterType($this->_adapterType);
```
fvovan: Probably we need to rename pagings as well. Not sure if it should be done in this PR.

library/CM/Wowza/Server.php
```diff
@@ -0,0 +1,74 @@
...
+    protected $_id;
+
+    /** @var string */
+    protected $_privateIp;
+
+    /** @var string */
+    protected $_publicIp;
+
+    /** @var string */
+    protected $_publicHost;
```
fvovan: But why we need  $_publicHost attribute?
Do we need to use domain name instead of IP somewhere?

library/CM/Wowza/Service.php
```diff
@@ -0,0 +1,95 @@
+<?php
+
+class CM_Wowza_Service extends CM_MediaStreams_Service {
```
fvovan: Wouldn't it be better to put it inside MediaStream "namespace" like  ```CM_MediaStreams_Service_Wowza``` or similar?

library/CM/Wowza/Cli.php
```diff
@@ -1,6 +1,6 @@
 <?php
 
-class CM_MediaStream_Cli extends CM_Cli_Runnable_Abstract {
+class CM_Wowza_Cli extends CM_Cli_Runnable_Abstract {
```
fvovan: Wouldn't it be better to put it into MediaStream ns ?

fvovan:
I like new concept.
We have separated exposed API methods, separate business logic entities in Server and Configuration classes, abstraction for HTTP communication and Service using all of these to manipulate the data.
The only thing is naming: should we always use MediaStreams or MediaStream and put all related classes into chosen namespace? Wdyt?

library/CM/Wowza/Server.php
tomaszdurka: As we've seen this is used in external repositories (e.g. Video component for rendering Flash client).

tomaszdurka:
Pair-reviewed with @fvovan

tomaszdurka:
image

PR #1962, opened 2015-10-20
Reduce error messages amount for CM_Elasticsearch_Client

fvovan:
Currently it add every error message for bulk API requests, and as each bulk request consists of many elements it pollutes our logs.
So need to change it to something like "bulk request contains 34 errors" and display few of them


fvovan:
@njam please review

njam:
lgtm!

PR #1970, opened 2015-10-22
don't clear cache on time-reset in tests

alexispeter:
In order to have more control over tests related to cache-clearing.


njam:
lgtm

alexispeter:
TH::timeForward() actually clears caches as well. 
@njam do you remember if there is a specific reason for this?

njam:
I don't remember. Agree it seems strange.
If everything works without this, I'd say it's good!

alexispeter:
Will investigate if any tests in our projects are relying on timeForward() clearing cache.

njam:
@alexispeter status?

alexispeter:
On hold until I have some spare time. I imagine this is going to break a lot of tests.

alexispeter:
please review @njam

njam:
lgtm

PR #1972, opened 2015-10-23
Add support of browserify for after/before vendor js assets

zazabe:


zazabe:
note: require `browserify`, see cargomedia/puppet-packages#954

zazabe:
@njam please review

njam:
- CM_Cache_Persistent: Uh oh, `getInstance()` is wrong!
 - Can you check manually that the caching for browserify works correctly (goes to file)?

Let's add the browserify npm as a dependency to `cm::application` before merging this:
https://github.com/cargomedia/puppet-packages/blob/master/modules/cm/manifests/application.pp

zazabe:
@njam 
- cache persistent: ok
- `cm:application`, already done: https://github.com/zazabe/puppet-packages/blob/browserify/modules/cm/manifests/application.pp
- only the last source map is used when multiple libraries are define in `after-body-source`, i'll check for a solution...

njam:
Ah so we should have 1 `browserify` call only? We can pass multiple entry points?

zazabe:
update: 
- fix `CM_Cache_Persistent`   
  - allow to have `$this->_lifetime = null` in `CM_Cache_Abstract`, for file storage
  - fix the singleton...
- browsersify all at once   
  the root path will be `after-body-source`, so js modules can be referenced relatively to this folder or relatively to where they are:  
```js
// after-body-source/foo/bar/toto.js
// after-body-source/foo/bar/file.js

// load toto from file.js
require('foo/bar/toto') === require('./toto')
```

zazabe:
mmmh, travis don't like browserify, cmd not found...
```
$ npm install -g browserify@11.2.0
---/\/---/home/travis/.nvm/v0.10.36/bin/browserify -> /home/travis/.nvm/v0.10.36/lib/node_modules/browserify/bin/cmd.js
```

zazabe:
@njam please review
Note:
For some reasons, passing the `NODE_PATH` env variable to the `exec` method produce some troubles, the command is not recognized, I guess it overrides other env var, the problem can be easily reproduced with a simple php script.   
I put the `NODE_PATH` before the cmd and it seems ok now.

zazabe:
FYI, the build failed is here: https://travis-ci.org/cargomedia/cm/jobs/87044153

njam:
- CM/Asset/Javascript/Abstract.php: Formatting
- CM/Asset/Javascript/VendorAfterBody.php: I remember we had some troubles once with some browsers if the javascript contained `;;`. If `_browserify()` returns empty string that will happen here. Maybe use `CM_Frontend_JavascriptContainer`?

zazabe:
@njam `CM_Frontend_JavascriptContainer` seems to return an empty string too if there's no operation... I think the best is to simply append the generated string, without any additional character, wdyt?

njam:
I thought we could use `CM_Frontend_JavascriptContainer` for *all* javascript "operations" in `CM_Asset_Javascript_VendorAfterBody`, and then concatenate them at then and and assign to `$_content`?

zazabe:
@njam please review

zazabe:
hum... the source maps is not so useful at the end, for debugging at least, because the browser is lost with breakpoints in mapped files.
I think it will work much nicely if we do a request specifically for `*-body-source`...

zazabe:
@njam please check 8751652
I separated the build of `*-body` and `*-body-source` js assets, now breakpoints are working fine but the downside is that we have 2 more requests, for `before-body-source.js` and `after-body-source.js`...
It's useful only in debug mode, but i think it's worth the advantage to have breakpoints ;)

Also, we can imagine to merge them together in the future, when all our js files will support CommonJS...

njam:
- Let's have 1 request only in production mode
- smarty_function_resourceJs: Add multiple `<script>` in dev mode

zazabe:
@njam please review

_Note: manually tested with `after-body-source` in multiple modules, in debug and production mode._

njam:
- CM/Asset/Javascript/Abstract.php::_browserify: Avoid keeping all JS in memory (use md5)
- CM_Asset_Javascript_Vendor: I would prefer if we would keep all logic about dev/prod *inside* this class, and not expose `mergeJs` and `browserifyJs`. This way we don't need to repeat the logic elsewhere. wdyt?

zazabe:
@njam please review

njam:
…just one more thing:
- CM/Asset/Javascript/Vendor/Abstract.php::_mergeJs: The method name doesn't reflect so clearly what's happening. Maybe `_appendDistJs()`? And `_appendSourceJs()`?

Rest looks good!

zazabe:
@njam please review
also I changed the `_process` method to something more explicit.

njam:
lgtm!

PR #1973, opened 2015-10-23
Small Request and Site improvements

njam:
@tomaszdurka could you review?

### DIFF
--- a/library/CM/Http/Request/Abstract.php
+++ b/library/CM/Http/Request/Abstract.php
@@ -335,14 +335,15 @@ abstract class CM_Http_Request_Abstract {
      * @throws CM_Exception_AuthRequired
      */
     public function getViewer($needed = false) {
-        if ($this->_viewer === false) {
-            $this->_viewer = $this->getSession()->getUser();
-        }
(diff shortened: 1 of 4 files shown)


njam:
@fauvel could you review?

tomaszdurka:
Looks good. Rerun failing test

fauvel:
Make sense, what are the use case you have in mind?

tomaszdurka:
What is the name of pull-request of adding User to Request? (this commit: https://github.com/njam/CM/commit/e2a60392c2fb410c0dba49ed24ad2b6c55a04104)
Can't find it.

I was actually a bit against that. Session is something better to be stored on a request and User can be part of session instead. Wdyt?

njam:
@tomaszdurka agree, Viewer and Session are conflicting concepts on the Request. I tried to remove Viewer in this PR, but it couldn't be done easily so I reverted it.
For now everything stays the same: Viewer has precedence.

njam:
@fauvel uses cases for `setSession()` is the cross-site redirect, will show you!

PR #1975, opened 2015-10-30
Add SetupScript to write class types to table

njam:
Table like `tmp_classType` which wouldn't be used by application.
Can be used for data warehouse.

Manual migration needed:
```sql
CREATE TABLE `cm_tmp_classType` (
  `type` int(10) unsigned NOT NULL,
  `className` varchar(255) NOT NULL,
  PRIMARY KEY (`type`)
) ENGINE=MyISAM DEFAULT CHARSET=utf8;
```

### DIFF
--- a/library/CM/Class/TypeDumper.php
+++ b/library/CM/Class/TypeDumper.php
@@ -1,0 +1,46 @@
+<?php
+
+class CM_Class_TypeDumper extends CM_Provision_Script_Abstract implements CM_Provision_Script_UnloadableInterface {
+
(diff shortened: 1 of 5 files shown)


njam:
@tomaszdurka @fvovan please review

See description for explanation.

fvovan:
looks good for me but probably naming is not the best, I would name it `aux_`

njam:
We already use `cm_tmp_` for a few other "generated" tables, so I suggest to stick with it?

fvovan:
ah, okay then.

PR #1979, opened 2015-10-30
CM_StreamChannel_Media invalidate key-cache on create

alexispeter:
Needs to invalidate cache on Create like parent class does. Probably was forgotten during refactoring.

please review @tomaszdurka


alexispeter:
please review @tomaszdurka

tomaszdurka:
Hmmm ok, but why it needs to be invalidated on creation? :o

alexispeter:
Because findByKeyAndAdapter() caches misses as well as hits, which can lead to problems with streamchannels that have some sort of repeating key.

tomaszdurka:
Looks good.

PR #1981, opened 2015-11-01
Consider CDN URL in site matching

njam:
@alexispeter @tomaszdurka please review

This is not necessarily needed atm because we pass the site-id along with CDN requests.
Still I think it makes sense to make the site matching consider it, in case we need it in the future…


alexispeter:
lgtm

PR #1977, opened 2015-11-01
Breaking: Add splash screen for Chrome

njam:
https://developers.google.com/web/updates/2015/10/splashscreen?hl=en

IE Tiles: https://msdn.microsoft.com/en-us/library/dn455106(v=vs.85).aspx


njam:
@christopheschwyzer I've added the necessary changes to `manifest.json`. Chrome will color the full screen in the "background_color" and place the app "icon" on top:


The icon should be rendered a bit bigger, but we lack the necessary files.
- We have: 36px, 48px, 72px, 96px, 144px, 192px
- [Article recommends](https://developers.google.com/web/updates/2015/10/splashscreen?hl=en): 128px, 196px, 256px, 384px, 512px

Do you know if there's any reason to keep such small images? If not I would say let's go with 128px - 512px?

christopheschwyzer:
Agree, small ones are probably not needed anymore.

njam:
I've updated the required sizes in the layout's HTML according to *best practices* I could find on the Internet. Can you take a look?

The problem is that we now need images that aren't generated by realfavicongenerator.net (e.g. the higher resolution android icons). Maaaybe we should look into scaling 1 image down on the fly…

christopheschwyzer:
> Maaaybe we should look into scaling 1 image down on the fly

Yes, let's do that. Same for the apple splash screens. Let's discuss.

christopheschwyzer:
@njam please review

njam:
lgtm, I trust you tested this

PR #1965, opened 2015-11-03
Add cm-janus service

tomaszdurka:
- Can be dummy
- Separate service instances for video/audio?

Used for:
- Expose registered [`cm-janus`](https://github.com/cargomedia/cm-janus/issues/4) instances
- Disabling functionality in dev VMs (not register service?)

### DIFF
--- a/library/CM/Janus/Configuration.php
+++ b/library/CM/Janus/Configuration.php
@@ -32,7 +32,7 @@ class CM_Janus_Configuration {
      * @param string $token
      * @return CM_Janus_Server|null
      */
-    public function findServerByToken($token) {
+    public function findServerByKey($token) {
         foreach ($this->_servers as $server) {
             if ($server->getKey() === $token) {
                 return $server;
--- a/library/CM/Janus/Configuration.php
+++ b/library/CM/Janus/Configuration.php
@@ -34,7 +34,7 @@ class CM_Janus_Configuration {
      */
     public function findServerByToken($token) {
         foreach ($this->_servers as $server) {
-            if ($server->getToken() === $token) {
+            if ($server->getKey() === $token) {
                 return $server;
             }
         }
--- a/library/CM/Janus/Configuration.php
+++ b/library/CM/Janus/Configuration.php
@@ -29,12 +29,12 @@ class CM_Janus_Configuration {
     }
 
     /**
-     * @param string $token
+     * @param string $key
      * @return CM_Janus_Server|null
      */
-    public function findServerByKey($token) {
+    public function findServerByKey($key) {
         foreach ($this->_servers as $server) {
-            if ($server->getKey() === $token) {
+            if ($server->getKey() === $key) {
                 return $server;
             }
         }
--- a/library/CM/Janus/Configuration.php
+++ b/library/CM/Janus/Configuration.php
@@ -1,0 +1,57 @@
+<?php
+
+class CM_Janus_Configuration {
+
+    /** @var CM_Janus_Server[] */
+    protected $_servers;
+
+    /**
+     * @param array|null $servers
+     */
+    public function __construct(array $servers = null) {
+        foreach ((array) $servers as $server) {
+            $this->addServer($server);
+        }
+    }
+
+    /**
+     * @param CM_Janus_Server $server
+     */
+    public function addServer(CM_Janus_Server $server) {
+        $this->_servers[] = $server;
+    }
+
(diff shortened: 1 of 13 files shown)


tomaszdurka:
Example config:

```
$config->services['janus'] = [
    'class'  => 'CM_Janus_Factory',
    'method' => [
        'name'      => 'createService',
        'arguments' => [
            'servers' => [
                [
                    'token'            => 'foo-bar',
                    'httpAddress'      => 'http://cm-janus.dev:80',
                    'webSocketAddress' => 'ws://cm-janus.dev:8188',
                ],
            ],
        ],
    ],
];
```

tomaszdurka:
@njam please review

tomaszdurka:
Added tests.
@njam or @fvovan please review.

library/CM/Janus/Factory.php
```diff
@@ -0,0 +1,25 @@
...
+                $serverId,
+                $serverConfig['token'],
+                $serverConfig['httpAddress'],
+                $serverConfig['webSocketAddress']
+            ));
+        }
+
+        $httpClient = new GuzzleHttp\Client();
+        $httpApiClient = new CM_Janus_HttpApiClient($httpClient);
+        $wowza = new CM_Janus_Service($configuration, $httpApiClient);
```
fvovan: Probably  `wowza` is not the best choice :p

library/CM/Janus/Configuration.php
```diff
@@ -0,0 +1,57 @@
...
+        return $this->_servers;
+    }
+
+    /**
+     * @param string $token
+     * @return CM_Janus_Server|null
+     */
+    public function findServerByToken($token) {
+        foreach ($this->_servers as $server) {
+            if ($server->getToken() == $token) {
```
fvovan: I would use strict comparison here

library/CM/Janus/HttpApiClient.php
```diff
@@ -0,0 +1,60 @@
...
+        $this->_httpClient = $httpClient;
+    }
+
+    /**
+     * @param CM_Janus_Server $server
+     * @param string $clientKey
+     * @return string
+     * @throws CM_Exception_Invalid
+     */
+    public function stopStream(CM_Janus_Server $server, $clientKey) {
```
fvovan: I would name it `$streamId`

fvovan:
I've got a little bit confused about authentication keys.
For now we need to have it for 3 kinds of connections: 
    - CM to cm-janus http server
    - cm-janus to CM RPC 
    - cm-janus to Janus via WS
Will we use the same value for all of these tokens?

library/CM/Janus/HttpApiClient.php
tomaszdurka: Hmm, this is naming taken from already existing streamChannels and stream (check `CM_Model_Stream_Abstract` and `CM_Model_StreamChannel_Abstract`)

tomaszdurka:
So we use secret/token/key for connecting:
- cm-janus -> cm
- cm -> cm-janus

The key is set per server and is the same on both sides - which I think is ok. Wdyt?

Let's not consider `cm-janus` <-> `janus` here at all.

library/CM/Janus/Configuration.php
tomaszdurka: Oh definetely. My mistake.

library/CM/Janus/Factory.php
tomaszdurka: :)

library/CM/Janus/HttpApiClient.php
fvovan: On cm-janus part it's called stream.id . afaiu related entities 'id'  and 'key' are interchangeable (most of times within this project). Possible it would be better to adjust cm-janus part to use `key` too.

fvovan:
Aha, so this `token` is not related to Janus authentication token https://janus.conf.meetecho.com/docs/auth.html#token 
Will adjust it to pass within HTTP headers for cm-janus auth according  https://github.com/fvovan/cm-janus/commit/45fcd3262f88f5b837f3dbf62384b0b6d7f5ad0d

tomaszdurka:
@fvovan please review

fvovan:
lgtm

tomaszdurka:
@fvovan there is still need to add `synchronize` method on the service which would use status from cm-janus.

PR #1983, opened 2015-11-09
Implement CM Janus synchronization

fvovan:
Implement synchronize method on CM_Janus_Service which uses cm-janus status to add/remove channels and their streams.
This method should be called within maintenance cli when service configured.

### DIFF
--- a/library/CM/Janus/Service.php
+++ b/library/CM/Janus/Service.php
@@ -60,8 +60,6 @@ class CM_Janus_Service extends CM_MediaStreams_Service {
                 if (!array_key_exists($streamChannel->getKey(), $channelSubscriberList)) {
                     $streamRepository->removeStream($streamPublish);
                 }
-            } else {
-                //TODO Is it possible to have channel w/o publishers?
             }
 
             /** @var CM_Model_Stream_Subscribe $streamSubscribe */
--- a/library/CM/Janus/Service.php
+++ b/library/CM/Janus/Service.php
@@ -37,9 +37,7 @@ class CM_Janus_Service extends CM_MediaStreams_Service {
             }
             $this->_httpApiClient->stopStream($stream->getServer(), $clientKey);
         }
-
         $startStampLimit = time() - 3;
-
         $streamKeyMap = [];
         foreach ($streamList as $stream) {
             $streamKeyMap[$stream->getStreamKey()] = true; //need just to define key
--- a/library/CM/Janus/Service.php
+++ b/library/CM/Janus/Service.php
@@ -9,8 +9,8 @@ class CM_Janus_Service extends CM_MediaStreams_Service {
     protected $_httpApiClient;
 
     /**
-     * @param CM_Janus_Configuration $configuration
-     * @param CM_Janus_HttpApiClient $httpClient
+     * @param CM_Janus_Configuration                $configuration
+     * @param CM_Janus_HttpApiClient                $httpClient
      * @param CM_MediaStreams_StreamRepository|null $streamRepository
      */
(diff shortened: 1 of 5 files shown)


library/CM/Janus/Service.php
```diff
@@ -20,7 +20,64 @@ public function __construct(CM_Janus_Configuration $configuration, CM_Janus_Http
...
+            $streamPublish = $streamChannel->getStreamPublishs()->getItem(0);
+            if ($streamPublish) {
+                if ($streamPublish->getStart() > $startStampLimit) {
+                    continue;
+                }
+                if (!array_key_exists($streamChannel->getKey(), $channelSubscriberList)) {
+                    $streamRepository->removeStream($streamPublish);
+                }
+            } else {
+                //TODO Is it possible to have channel w/o publishers?
```
fvovan: @tomaszdurka wdyt?

fvovan:
@tomaszdurka  please review,
generally it works but i'm thinking about edge cases
also we need either change output format in `/status` API method at cm-janus
or convert corresponding param names in CM's JanusHttpClient

library/CM/Janus/Service.php
```diff
@@ -20,7 +20,64 @@ public function __construct(CM_Janus_Configuration $configuration, CM_Janus_Http
     }
 
     public function synchronize() {
-        throw new CM_Exception_NotImplemented();
+        $streamRepository = $this->getStreamRepository();
+
+        $startStampLimit = time() - 3;
+        $streamList = [];
+        $channelSubscriberList = [];
+        foreach ($this->_configuration->getServers() as $server) {
```
tomaszdurka: Oh I see you've copied synchronization from wowza. I think you can rather use `CM_Janus_Service-> _fetchStatus()`

fvovan: Actually, not entirely. :) Yes, I saw this method, but as you can see there is some processing inside the loop.
I can fetch status using existing method but anyway will have to iterate through it once more.

fvovan:
it's adjusted to use `_fetchStatus`

tomaszdurka:
@fvovan can you double-check what we wrote, plus run tests with coverage.

tomaszdurka:
Change `CM_Janus_HttpApiClient` to expect `{id: stream.id, channelName: stream.channelName, isPublish: stream.isPublish};` response.

fvovan:
@tomaszdurka please review
I fixed some flaws and added several test cases.
Line coverage was increased from 84% to  96%  (basically only getConfigure() left untested, but it is very trivial)

library/CM/Janus/Service.php
```diff
@@ -20,7 +20,55 @@ public function __construct(CM_Janus_Configuration $configuration, CM_Janus_Http
...
+                }
+                if ($streamChannel->getStreamSubscribes()->findKey($clientKey)) {
+                    continue;
+                }
+            }
+            $this->_httpApiClient->stopStream($stream->getServer(), $clientKey);
+        }
+
+        $startStampLimit = time() - 3;
+
```
tomaszdurka: Remove empty line?

tomaszdurka: Oh this is on purpose I can see. For each block. Maybe rather not? Probably newline separator for separating stopping client and removing them from database is enough.

tomaszdurka:
Looks good except maybe those newlines. I've restarted failing build.

tomaszdurka:
@fvovan oh, you forgot to add it to maintenance cli.

Please add here https://github.com/cargomedia/cm/issues/1986.
Thanks!

PR #1984, opened 2015-11-10
Introduce ArrayConvertible::toArrayIdOnly()

njam:


njam:
Paired with @fauvel , @tomaszdurka fyi

PR #1986, opened 2015-11-10
Run Janus synchronization periodically

fvovan:
Run Janus synchronization periodically when service present.
Add to CM_Maintenance_Cli (next to wowza).


fvovan:
@tomaszdurka  please review

tomaszdurka:
Looks good.

PR #1955, opened 2015-11-11
Fix CM_Elasticsearch_Client error handling

fvovan:
Currently we handle 503 ElasticSearch errors which are not errors by nature but only a consequence of delayed shard initialization (and this error occurs only during tests execution)
Author of elasticsearch-php library advised to insert awaiting block to tests
```
 $client->cluster()->health([
  'wait_for_status' => 'yellow'
]);
 ```
and then we will be able to remove mentioned error handling from CM_Elasticsearch_Client


fvovan:
@njam please review.

njam:
lgtm!

PR #1988, opened 2015-11-11
Add promiseOnce/removePromiseOnce util method

zazabe:


zazabe:
@vogdb please review

zazabe:
@vogdb you can check usages in zazabe/sk@13031f5

vogdb:
lgtm
How do you think? Should we throw an Error in case if there is an existing `storage[uid]` already at the stage of `promiseOnce` call?

zazabe:
@vogdb mmmh, i don't get your idea because the expected behaviour in this case is to return the existing promise, isn't it?

vogdb:
I meant rather the situation
```js
promiseOnce(function hello(){console.log('hello')}, 'the_same_key')
```
```js
promiseOnce(function bye(){console.log('bye')}, 'the_same_key')
```

zazabe:
@vogdb ok, done

vogdb:
Perfetto!

PR #1991, opened 2015-11-17
Add color param to contentPlaceholder

zazabe:


zazabe:

christopheschwyzer:
Could we just use `css`?

zazabe:
I tried, but the image generated is just a plain white img without transparency... but yes, we can also always generate it with `alpha: 100%`...

zazabe:
@christopheschwyzer 

we can do: 
```
.contentPlaceholder img {
  visibility: hidden;
}
```

and then use any background on `.contentPlaceholder`, but I thought that having an hardcoded color for the placeholder is a little bit strange, isn't it?

njam:
100% transparency and then using CSS sounds very good!

zazabe:

christopheschwyzer:
lgtm

PR #1992, opened 2015-11-18
New feature detections to Modernizr.

vogdb:


zazabe:
WebGL detection is not working well with the Modernizr method, please check this implementation with the three.js method: d97201f

zazabe:
@vogdb if you agree, please cherry-pick my changes

vogdb:
@zazabe you have generated `modernizr.js` from the latest master. I'm not sure about its stability.

zazabe:
@vogdb yes, but there's no instruction about which version to use, i added a `README.md`. Thinking also about doing a PR on Modernizr for WebGL, the current official test doesn't work in all cases.

vogdb:
lgtm here but lets wait a bit for Modernizr/Modernizr#1784

PR #1993, opened 2015-11-18
set user's currency as environment-currency

alexispeter:
So environment-currency can be used throughout views. 
Simplifies things for views that need a currency but not necessarily a viewer.


alexispeter:
please review @dlondero 

paired with @njam

dlondero:
@alexispeter lgtm!

alexispeter:
tagged https://github.com/cargomedia/cm/releases/tag/1.160.5

PR #1997, opened 2015-12-02
Extract base64-encryption from streamchannel

alexispeter:
So as to have it available for other applications.


alexispeter:
please review @fvovan

fvovan:
I would add non-empty key checking.

alexispeter:
Could you elaborate?

fvovan:
sure, 15 mins. please add me to repo's collaborators.

alexispeter:
If you're talking about empty string for encryption key, they work perfectly well, so I don't see a reason to explicitly prohibit empty keys.

library/CM/Util/UrlEncryptor.php
```diff
@@ -0,0 +1,23 @@
+<?php
+
+class CM_Util_UrlEncryptor {
```
fauvel: I wouldn't call this `CM_Util_UrlEncryptor` – what about `CM_Util_Encryption`?

fvovan:
>If you're talking about using empty strings for encryption key, they work perfectly well, so I don't see a reason to explicitly prohibit that.

yes I checked that it works with empty string and even with nulls and that could be insecure. But may be I'm just paranoid :)

alexispeter:
Actually mcrypt is padding keys that are too short with '\0' if they are too small. So an empty key would probably actually be something like `\0\0\0\0\0\0\0...`

library/CM/Util/UrlEncryptor.php
fvovan: The point is it returns and gets url compatible strings (processed with Base64)
Probably it is worth to separate Base64 from encryption and add methods like `decryptUrl` and `encryptUrl`

fauvel:
I thought the *data* would be padded to match a multiple of the data bock size required by the encryption algorithm?

fauvel:
Oh, we should test it with php 5.6...
> PHP 5.6.0:	Invalid key and iv sizes are no longer accepted. mcrypt_encrypt() will now throw a warning and return FALSE if the inputs are invalid. Previously keys and IVs were padded with '\0' bytes to the next valid size.

alexispeter:
Then let's just enforce the minimum keysize.

fauvel:
For `MCRYPT_RIJNDAEL_128`, the key length should be exactly 32 bytes according to [this post](http://stackoverflow.com/questions/4128497/picking-encryption-cipher-for-mcrypt#answer-19737721).

alexispeter:
I think it's 128 bits = 16 bytes.

> * AES-128 refers to 128 bit key size
* AES-256 refers to 256 bit key size

fauvel:
MCRYPT_RIJNDAEL_128 is AES-256.

alexispeter:
Are you sure? According to [wikipedia](https://en.wikipedia.org/wiki/Advanced_Encryption_Standard#Description_of_the_cipher) there are 128, 196 and 256 bit variants of AES and there are 128, 192 and 256 bit rijndael variants according to [php documentation](http://php.net/manual/en/mcrypt.ciphers.php)

fauvel:
Tested in php56:
```
Warning: mcrypt_encrypt(): Key of size 29 not supported by this algorithm. Only keys of sizes 16, 24 or 32 supported
```

fauvel:
Let's throw if `strlen($encryptionKey)` differs from 16, 24 and 32?

fvovan:
>Let's throw if strlen($encryptionKey) differs from 16, 24 and 32?

agree

fauvel:
Ok, so it looks like `MCRYPT_RIJNDAEL_128` and `MCRYPT_RIJNDAEL_256` both implement AES-128, 196 and 256 (key sizes of 16, 24 and 32 bytes) – they only differ by the block size (zero-padded to a multiple of 16 or 32 bytes, respectively.)

alexispeter:
There is a function `mcrypt_module_get_supported_key_sizes($algorithm)`. Let's use that. It returns [16, 24, 32] for MCRYPT_RIJNDAEL_128.

fauvel:
sgtm!

fauvel:
@alexispeter see also https://github.com/cargomedia/cm/pull/1998

library/CM/Util/UrlEncryptor.php
fauvel: It is ok to keep base64 encoding/decoding here imo. But having "Url" in class or method names is misleading – it sounds like we're encrypting full urls, and that we cannot encrypt anything else. I guess you have authentication tokens for direct login urls in mind, but we already have a couple of other use cases in place, so let's keep naming generic!

library/CM/Util/Encryption.php
```diff
@@ -0,0 +1,54 @@
...
+
+    /**
+     * @param string $secretKey
+     * @throws CM_Exception_Invalid
+     */
+    private function _validateKey($secretKey) {
+        $keySize = strlen($secretKey);
+        $validKeySizes = mcrypt_module_get_supported_key_sizes(MCRYPT_RIJNDAEL_128);
+        if (!in_array($keySize, $validKeySizes, true)) {
+            throw new CM_Exception_Invalid('Using wrong key for encryption/decryption');
```
fauvel: What about "Invalid key size"?

fauvel:
Btw I've merged #1998, so this will be tested on Travis with php56, too.

alexispeter:
@fvovan You know, I would really appreciate it if you didn't simply push code onto a branch I'm currently working on. Or at least warn me that you're taking over so I don't waste time working on something that is being worked on by someone else in parallel.

library/CM/Util/Encryption.php
```diff
@@ -0,0 +1,54 @@
+<?php
+
+class CM_Util_Encryption {
+
+    /**
+     * @param string $data
+     * @param string $secretKey
+     * @return string
+     */
+    public function encryptUrl($data, $secretKey) {
```
fauvel: Could you please rename this to `encryptBase64`/`decryptBase64` for clarity?

fvovan:
sorry @alexispeter 
 I just thought  that
> Could you elaborate?

was an invitation

alexispeter:
It was an invitation to explaining to me what you meant, because I wasn't too sure what exactly you meant.

library/CM/Util/Encryption.php
alexispeter: Should we rename the class to CM_Util_Base64Encryption instead? Or do you intend to add non-base65 encryption/decryption to the class?

fauvel: Personally I would name the methods:
- `CM_Util_Encryption::decrypt`
- `CM_Util_Encryption::decryptBase64`
- `CM_Util_Encryption::encrypt`
- `CM_Util_Encryption::encryptBase64`

and I wouldn't use the term "url" at all. Wdyt?

fauvel: Probably we're going to encrypt json, too (e.g. `userId` and `expireStamp` for authentication tokens) – what about wrapping this in:
- `CM_Util_Encryption::decrypt` (in: binary string, out: string)
- `CM_Util_Encryption::decryptBase64` (in: base64 string, out: string)
- `CM_Util_Encryption::decryptJson` (in: binary string, out: mixed)
- `CM_Util_Encryption::decryptJsonBase64` (in: base64 string, out: mixed)
- `CM_Util_Encryption::encrypt` (in: string, out: binary string)
- `CM_Util_Encryption::encryptBase64` (in: string, out: base64 string)
- `CM_Util_Encryption::encryptJson` (in: mixed, out: binary string)
- `CM_Util_Encryption::encryptJsonBase64` (in: mixed, out: base64 string)

alexispeter: I would still just inline the base64 encoding, or are there instances where it could actually hurt to do it? Additional processing is negligible.
If we absolutely have to implement 3 ways of encoding, I would prefer doing it in separate classes with a common interface.

fauvel: Agree on inlining base64. What about having a convenience interface for json, though?
- `CM_Util_Encryption::decrypt` (base64 string => string)
- `CM_Util_Encryption::decryptJson` (base64 string => mixed)
- `CM_Util_Encryption::encrypt` (string => base64 string)
- `CM_Util_Encryption::encryptJson` (mixed => base64 string)

fvovan:
sorry again @alexispeter , a misunderstanding occurred
shall I revert my changes or continue with this PR?

alexispeter:
Your call, I don't care.

fvovan:
okay then I would finish it.

fauvel:
@fvovan agree on the interface specified in https://github.com/cargomedia/cm/pull/1997#discussion_r46550866?

library/CM/Util/Encryption.php
alexispeter: Again, imo this should be a separate class.

fauvel: Something like that?
- `CM_Util_Encryption::decrypt` (base64 string => string)
- `CM_Util_Encryption::encrypt` (string => base64 string)
- `CM_Util_Encryption_Json::decrypt` (base64 string => mixed)
- `CM_Util_Encryption_Json::encrypt` (mixed => base64 string)

alexispeter: Yes something like that.

fvovan:
So do you want to inline `base64()` at `encrypt() / dectypt()` methods, or in client code (like `CM_Model_StreamChannel_Abstract`)?
@fauvel do you mean not 'Interface' literally but just two additional methods?

library/CM/Util/Encryption.php
alexispeter: A json implementation is actually needed for the very reason we decided to extract this functionality in the first place. But we can leave it out of this PR, I need something to do anyway.

fvovan:
@alexispeter or @fauvel please  review.

fvovan:
it seems adding one more php version, made Travis slower.. Inconsistent errors appear in every build

alexispeter:
looks good to me

PR #1998, opened 2015-12-03
Test on php56

fauvel:


fauvel:
@njam I'm testing CM in php56 because of these [mcrypt changes](http://php.net/manual/de/function.mcrypt-encrypt.php#refsect1-function.mcrypt-encrypt-changelog).

How about upgrading runkit to [1.0.4](https://github.com/zenovich/runkit/releases/tag/1.0.4) in [puppet](https://github.com/cargomedia/puppet-packages/blob/master/modules/php5/manifests/extension/runkit.pp#L3)? Do you remember the reason why we're using specifically [this commit](https://github.com/zenovich/runkit/commit/5e179e978af79444d3c877d5681ea91d15134a01) (which doesn't support php56) instead of an official release?

njam:
Agree, lgtm

See https://github.com/cargomedia/puppet-packages/pull/1022

PR #1996, opened 2015-12-03
Remove streamPublish instead of subscribe in Janus endpoints

fvovan:
https://github.com/cargomedia/cm/blob/master/library/CM/Janus/RpcEndpoints.php#L92


fvovan:
@tomaszdurka please review

tomaszdurka:
Looks good.

PR #2000, opened 2015-12-04
Test on php7

fauvel:


fauvel:
@njam feels like [donating to Dmitry](https://github.com/zenovich/runkit/issues/87#issuecomment-146682396)?

njam:
Let's upgrade Debian first! :)

fauvel:
I've seen php7-rc2 has been packaged for deb8, but probably we'll have to wait for deb9 for full php7 support? ETA 1-2y...

fauvel:
Not nice for open source projects built on top of cm :/

njam:
Maybe instead we should drop runkit usage? :)
The only thing it's needed for I think is:
```
runkit_function_copy('time', 'time_original');
```

fauvel:
Yep, maybe we could use a `CM_Util_Time` class for that, with a getter and a setter to "freeze" time in `CM_Bootloader_Testing`?

njam:
There's timecop: https://github.com/hnw/php-timecop
Unfortunately also only 5.5…

fauvel:
It looks like the simplest *reliable* way to achieve time freezing in unit tests would be to intercept the system calls made by the phpunit process – see for instance the [faketime](https://github.com/wolfcw/libfaketime) utility. Otherwise you must either write a php extension to override the time functions (and there are a lot of them), or write your own userspace wrapper class/function to mock time functions – and adjust all usages in your code (which you obviously cannot do easily for 3rd-party dependencies.)

njam:
*libfaketime* sounds like a good solution! As it would also work for `new DateTime()`…
Let's look into this 2016 though, as it's not a priority currently!

fauvel:
agree!

njam:
@dlondero maybe also for you

njam:
Closing in favor of https://github.com/cargomedia/cm/issues/2478

PR #2001, opened 2015-12-04
move json encoding/decoding from CM_Params to CM_Util

alexispeter:


alexispeter:
please review @njam

njam:
lgtm

PR #2003, opened 2015-12-04
make encryption url-safe

alexispeter:
`+/=` are not url-safe, see http://stackoverflow.com/questions/1374753/passing-base64-encoded-strings-in-url/5835352#5835352

### DIFF
--- a/library/CM/Util/Encryption.php
+++ b/library/CM/Util/Encryption.php
@@ -9,7 +9,8 @@ class CM_Util_Encryption {
      */
     public function encrypt($data, $secretKey) {
         $this->_validateKey($secretKey);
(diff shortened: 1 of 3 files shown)


alexispeter:
please review @njam

njam:
lgtm, please revert test.php

@fvovan wanna have a look as well?

alexispeter:
Hmm I think I should put a comment into the code.

fvovan:
But why it worked before (when is was not extracted to separate class)?

alexispeter:
I suppose the previous applications (nodejs message streaming) didn't care about it.

alexispeter:
tagged https://github.com/cargomedia/cm/releases/tag/1.160.12

PR #2005, opened 2015-12-07
Specify apcu pecl version.

fvovan:
Recently v.5 of  requires PHP v.7


fvovan:
@njam  please review

njam:
Let's use 4.0.4 - which is the same as we use internally: https://github.com/cargomedia/puppet-packages/blob/master/modules/php5/manifests/extension/apcu.pp

njam:
As discussed we'll go to 4.0.10 - lgtm!

PR #2007, opened 2015-12-09
Adjust `CM_Log_Handler_MongoDb` to use MongoDB write concerns options

fvovan:


fvovan:
@fauvel please review.
I'm not sure if it's worth to make a validation for this parameter. probably it will be an overkill.

fauvel:
It's good to have a default, but you've forgotten to use `$insertOptions` if it's not `null` :P
The rest lgtm!

fvovan:
oh bad me! thanks!

library/CM/Log/Handler/MongoDb.php
```diff
@@ -11,17 +11,22 @@ class CM_Log_Handler_MongoDb extends CM_Log_Handler_Abstract {
...
      * @param int|null $level
      * @throws CM_Exception_Invalid
      */
-    public function __construct($collection, $recordTtl = null, $level = null) {
+    public function __construct($collection, $recordTtl = null, array $insertOptions = null, $level = null) {
         parent::__construct($level);
         $this->_collection = (string) $collection;
         $this->_mongoDb = CM_Service_Manager::getInstance()->getMongoDb();
         $this->_recordTtl = null !== $recordTtl ? (int) $recordTtl : 3600 * 24 * 60;
+        $this->_insertOptions = null !== $insertOptions ? $insertOptions : ['w' => 0];
```
fauvel: Maybe you should still use your write concerns option as a default if further options are being passed?
```php
$insertOptions = (array) $insertOptions;
if (!isset($insertOptions['w'])) {
    $insertOptions['w'] = 0;
}
```

fvovan: hmm. not sure about it. Probably it's better to be more explicit. I mean to have default values only as a fallback and pass to mongoDB precisely what users (developers) mentioned in method calling. These params are quite tricky, some of them depend and implies from each other. Could be confusing.

fauvel:
lgtm then!

fvovan:
oh. I completely forgot about storing HTTP method at logs. Probably you will suggest some other fields need from `CM_Model_User` or from `CM_Http_Request_Abstract`
Currently `headers` , `uri` and `server` attributes are stored, and I'm going to add `method` to them, may be include  `cookies` too?

fvovan:
depends on https://github.com/cargomedia/cm/pull/2009

fvovan:
@fauvel, please review. 
I've just added request `method` and `query` to stored fields.

fauvel:
lgtm

PR #2008, opened 2015-12-09
Adjust `CM_Log_Logger` interface

fvovan:
make `CM_Log_Logger::addRecord` non public

### DIFF
--- a/library/CM/Log/Logger.php
+++ b/library/CM/Log/Logger.php
@@ -40,25 +40,6 @@ class CM_Log_Logger {
     }
 
     /**
-     * @param CM_Log_Record $record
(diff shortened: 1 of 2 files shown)


fvovan:
@fauvel  please review

fauvel:
Lgtm. Regarding the logging of logging errors: Is your code safe against infinite error handling loops?

fvovan:
Actually it's supposed to be, as we made a special class for  exceptions occurred during handling and it  could be checked in `CM_ExceptionHandling_Handler_Abstract`. But migration is not complete now. I mean that old system is used for exceptions handling currently.

PR #2010, opened 2015-12-10
Add user-agent parsing library

fvovan:
For several reason we need a user-agent parsing library (again).
The main goals are:
- Detect device category (desktop, tablet, mobile)
- Detect device vendor and model

Options:
- https://github.com/serbanghita/Mobile-Detect
- https://github.com/ua-parser/uap-php
- …possibly others - let's investigate!

Considerations:
- Wrap it somehow?
- Should work together with `CM_Http_Request_Abstract::getHeader('user-agent')`
- Should be cached, if not done in the library already?


fauvel:
@njam quick review?

fvovan:
unfortunately they rely on PHP HTTP headers format ( I mean prefixed with "HTTP_")

njam:
How long does it take to instantiate the class and check for `isMobile()` (see `CM_Util::benchmark()` for example).

Let's remove some application specific info from the test! Otherwise looks good!

fvovan:
>How long does it take to instantiate the class and check for isMobile() (see CM_Util::benchmark() for example).

about 3ms

fvovan:
@njam please review.
decided together with Sebastien to add caching.

njam:
- Use `__METHOD__`  instead of `__CLASS__, 'isMobile'`

rest lgtm

PR #2002, opened 2015-12-10
Add a method to retrieve the default Janus Server

fvovan:
It's not very convient to know which Janus server should be used, I need to get janus config to connect the CamShow.

I'm currently doing something like this:
```
$serviceManager = CM_Service_Manager::getInstance();
$janus = $serviceManager->getJanus('janus');
$janusServer = $janus->getConfiguration()->getServer(0));        
```

- why i have to pass the `serviceName` to `getJanus()`?
- how can i know which server i should use? this is not about load balancing? should it be transparent at my level?

### DIFF
--- a/library/CM/Janus/Configuration.php
+++ b/library/CM/Janus/Configuration.php
@@ -56,4 +56,16 @@ class CM_Janus_Configuration {
         }
         throw new CM_Exception_Invalid('Cannot find server with id `' . $id . '`');
     }
+
+    /**
+     * @return CM_Janus_Server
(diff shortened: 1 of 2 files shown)


fvovan:
@tomaszdurka please review

tomaszdurka:
Looks good.

@zazabe we've added `getServerRandom` on `Configuration` class. The same place where `getServer($id)` is.

fvovan:
it that's okay, please feel free to merge it

PR #2013, opened 2015-12-11
Don't catch InvalidParam in RPC endpoint

njam:
We experienced some "blindness" because of this catching. I suggest to *not* catch InvalidParam, so that it gets logged and is more visible.

@alexispeter @zazabe please review


alexispeter:
Couldn't we just change the config so that it catches AND logs the exception?
```php
'CM_Exception_InvalidParam' => ['log' => 'CM_Paging_Log_NotFound']
```

njam:
Good point!
I added it for all errors.
Wdyt?

alexispeter:
lgtm

PR #2016, opened 2015-12-11
CM_View_Abstract ajax_callbacks should be static

alexispeter:


alexispeter:
Actually the shouldn't

njam:
Not anymore!

alexispeter:
Force of habit :D

PR #2018, opened 2015-12-14
StreamChannel: Make stream-object available in canSubscribe/canPublish

alexispeter:
See discussion in cargomedia/sk#2432


alexispeter:
please review @tomaszdurka

tomaszdurka:
Hmm, but shouldn't canPublish/canSubscribe be run without stream?

alexispeter:
We want to be able to use isSubscribe/isPublish inside canSubscribe/canPublish. For that we need the stream.

tomaszdurka:
Bit confused here. Didn't you @njam comment on this one once we talked?

njam:
I'm also not sure about this… let's discuss in https://github.com/cargomedia/sk/issues/2456

alexispeter:
Discontinued.

PR #2020, opened 2015-12-14
add 'needed' param to smarty.load

fvovan:


fvovan:
@alexispeter please review

alexispeter:
* library/CM/SmartyPlugins/function.load.php
 + test the returnValue of getLayoutPath() for null, instead of `$file` which might mean something else.
* tests/library/CM/SmartyPlugins/function.load.php fileName needs to end with Test, else the test-runner will ignore it.

fvovan:
oh. bad me :( . need to stop pushing after 18-00.
thanks @alexispeter.
could you look check  it please?

alexispeter:
lgtm

PR #2027, opened 2015-12-17
Extend button_link with href

christopheschwyzer:


christopheschwyzer:
@njam @tomaszdurka what do you think?
Theoretically you could pass now `page` and `href` - don't think it's a problem.

njam:
Hmhmhm shouldn't we use an  `<a>` element with an `href` attribute?
Maybe we could based on if there is an `href` render `a` or `button`, and all the rest stays the same?
Let me know if you want to pair.

christopheschwyzer:
We would need some additional css - but yes - sounds reasonable.

christopheschwyzer:
@njam please have a look

njam:
lgtm

PR #2028, opened 2015-12-17
Update phpunit

dlondero:


dlondero:
@tomaszdurka please review

`@runInSeparateProcess` annotation has some issues when tests are run in PHPStorm: looks like `PHPUNIT_COMPOSER_INSTALL` is not defined when phpstorm runs phpunit... See [here](https://devnet.jetbrains.com/message/5565216?tstart=0) and [there](https://www.drupal.org/node/2597814).

tomaszdurka:
So this constant is used by intellij?

dlondero:
@tomaszdurka looks like it is. If you run tests manually from the console they are run w/o any issue.

tomaszdurka:
Ok, if there is no other way. Btw this is connected with the version you upgrade to? Or it's present atm?

dlondero:
Connected with 4.8 which is the last old stable (support ends on August 5, 2016). 4.3 support ended ages ago.

PR #2030, opened 2015-12-21
Janus' checkStreams not working

fvovan:
Reproduce:
- Create a video stream
- Set "allowedUntil" to the past (e.g. no more money)
- -> The stream is not stopped.

It looks to me like `checkStreams()` is not called from maintenance.

@tomaszdurka @fvovan can you check? And maybe test the whole thing to make sure the whole mechanism works ?


fvovan:
@tomaszdurka please review

tomaszdurka:
Looks good.

PR #2029, opened 2015-12-21
Create streamchannel when subscribing in Janus

fvovan:
Do not throw, but rather crete stream-channel:
https://github.com/cargomedia/cm/blob/master/library/CM/Janus/RpcEndpoints.php#L61

Should we create stream-channel here only if it doesn't exist then?
https://github.com/cargomedia/cm/blob/master/library/CM/Janus/RpcEndpoints.php#L30

### DIFF
--- a/library/CM/Janus/RpcEndpoints.php
+++ b/library/CM/Janus/RpcEndpoints.php
@@ -25,11 +25,13 @@ class CM_Janus_RpcEndpoints {
         $paramsChannel = CM_Params::factory(CM_Params::jsonDecode($channelData), true);
         $streamChannelType = $paramsChannel->getInt('streamChannelType');
 
-        $server = $janus->getConfiguration()->findServerByKey($serverKey);
         $streamRepository = $janus->getStreamRepository();
         $streamChannel = $streamRepository->findStreamChannelByKey($streamChannelKey);
-        if (!$streamChannel) {
+        if (null !== $streamChannel) {
+            $server = $janus->getConfiguration()->findServerByKey($serverKey);
             $streamChannel = $streamRepository->createStreamChannel($streamChannelKey, $streamChannelType, $server->getId(), 0);
+        } elseif ($streamChannel->getType() !== $streamChannelType) {
+            throw new CM_Exception_Invalid('Passed stream channel type does not match existing');
         }
         try {
             $streamRepository->createStreamPublish($streamChannel, $user, $streamKey, $start);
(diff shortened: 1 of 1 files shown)


fvovan:
@tomaszdurka  please review.
I added the checking we discussed about.
Not sure about adding absent stream channel on publish. I would rather throw there.

tomaszdurka:
Please make those methods almost identical. Publish should check type too.

fvovan:
it's done.
btw please note the additional parameter for `rpc_subscribe()`

tomaszdurka:
Looks good. This is probably breaking. @zazabe @kris-lab please keep in mind subscribe requires channel_data now too.

zazabe:
@tomaszdurka 
by subscribe, you mean `"janus": "attach"` request or specific request to connect to a channel, like:
- audio: `"request": "join"` and `"request": "changeroom"`
- video: `"request": "watch"` and `"request": "switch"` 
?

fvovan:
afaik 
`"janus": "webrtcup"`  and  `"request": "switch"`

zazabe:
`"janus": "webrtcup"` is send by janus, not by the client, but I suppose @kris-lab can set it on the janus side...

@tomaszdurka @fvovan What about sending the `channel_data` for any client plugin request? seems not clear to me which one will require it or not...

fvovan:
I checked it now. 
it's not `"request": "switch"` but rather a response to it. so it seems janus side as well

kris-lab:
@zazabe well I can change thing related to the plugins. I cannot change anything in the core of Janus. Not sure what you have in minds? E.g. `webrtcup` is very low level Janus operation where we have no access from plugins. Nothing can be added to this `webrtcup`, `hangup` etc event message.

zazabe:
@kris-lab I really don't know, i'm just saying that on the client side, we can't had the `channel_data` to request like `"janus": "webrtcup"` because they are coming from janus...
@tomaszdurka @fvovan : could you clarify exactly on which request you will need the `channel_data`?

fvovan:
it's actually the same value we use during `publish` e.g.  only `streamChannelType` (JSON encoded)
the problem is that we `publish` streams to CM side on browser's requests but `subscribe` to them on janus' requests
 
@tomaszdurka @kris-lab could we have a call?
afaiu this issue is connected with https://github.com/cargomedia/cm-janus/pull/108 we want to create only audiostreams on the fly, probably `streamChannelType` corresponding to videostream could be used (stored somehow)

tomaszdurka:
What about storing channel_data on Stream in cm-janus? I don't like this solution though.

tomaszdurka:
As discussed we will store `channelData` in cm-janus Stream. Done in https://github.com/cargomedia/cm-janus/pull/108.

tomaszdurka:
@fvovan please release/tag new version of cm. Update in depending project.

fvovan:
1.161.0 added

PR #2036, opened 2016-01-04
Don' use static singleton for smarty

njam:
@tomaszdurka please review


tomaszdurka:
This was not for some reason? I thought that maybe we needed to have some global smarty state (loading plugins?). Still I think it will be ok.

Btw there is an issue in #1652 which was to make render a service.

njam:
It was for performance I think. But I checked, and it doesn't have an impact, not even on tests, where we have multiple render instances.

PR #2032, opened 2016-01-06
Add ICE config for Janus

fvovan:
Basically we need an array of ICE servers:
```
iceServers: [{url: "turn:yourturnserver.com:3478", username: "janususer", credential: "januspwd"}]
```

Initially username/credential will probably be hardcoded application-wide. In the future we might want to generate them on the server side (or client side with an ajax call).

I think they could be configured *per* janus server. E.g. in case we have geographically distributed janus instances, then it makes sense to use a *close* ICE server.

cc @tomaszdurka @zazabe @kris-lab

### DIFF
--- a/library/CM/Janus/Factory.php
+++ b/library/CM/Janus/Factory.php
@@ -13,7 +13,8 @@ class CM_Janus_Factory {
                 $serverId,
                 $serverConfig['key'],
                 $serverConfig['httpAddress'],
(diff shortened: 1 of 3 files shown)


fvovan:
@njam please review,
is it okay that I changed key name  or is it already mentioned somewhere?

njam:
lgtm, could you tag and update?

PR #2039, opened 2016-01-08
Enable audioStream-archiving and identification of archives from janus

alexispeter:
##### audioStream archiving
Archiving should behave according to two modes
- streamChannel has a publisher (old behaviour)
 + archiving should occur on unpublish
 + createStamp is streamPublish startStamp
 + sets userId according to streamPublish (can be null)
 + hash is derived from streamPublish key
- streamChannel has no publisher (new)
 + archiving should occur on delete
 + createStamp is streamChannel createStamp (needs to be added to the model)
 + userId is null
 + hash is derived from streamChannel key

##### archive-identification
Introduce unique `mediaId` that is generated by janus and passed to the php-application
- add optional `mediaId` property to streamchannel and archive
- pass `mediaId` to janus publish/subscribe rpc-endpoints
- identify archive by `mediaId` in CM_MediaStreams_Cli archiving function

cm-janus changes cargomedia/cm-janus#122


alexispeter:
please review @tomaszdurka

library/CM/Janus/RpcEndpoints.php
```diff
@@ -24,12 +24,13 @@ public static function rpc_publish($serverKey, $streamChannelKey, $streamKey, $s
 
         $paramsChannel = CM_Params::factory(CM_Params::jsonDecode($channelData), true);
         $streamChannelType = $paramsChannel->getInt('streamChannelType');
+        $mediaId = $paramsChannel->getString('mediaId');
```
tomaszdurka: All the `channelData` are data generated by cm server-side and passed into client-side which passes it back to server-side.
Here we need to use something generated by cm-janus, so we need to pass it as additional parameter - not part of `channelData`.

So: Use new method parameter instead of channelData param.

library/CM/Janus/RpcEndpoints.php
```diff
@@ -24,12 +24,13 @@ public static function rpc_publish($serverKey, $streamChannelKey, $streamKey, $s
...
 
         $streamRepository = $janus->getStreamRepository();
         $streamChannel = $streamRepository->findStreamChannelByKey($streamChannelKey);
         if (null === $streamChannel) {
             $server = $janus->getConfiguration()->findServerByKey($serverKey);
-            $streamChannel = $streamRepository->createStreamChannel($streamChannelKey, $streamChannelType, $server->getId(), 0);
+            $streamChannel = $streamRepository->createStreamChannel($streamChannelKey, $streamChannelType, $server->getId(), 0, $mediaId);
         } elseif ($streamChannel->getType() !== $streamChannelType) {
             throw new CM_Exception_Invalid('Passed stream channel type does not match existing');
```
tomaszdurka: I guess we should/could also add check about mediaId here. Or maybe `findServerByKey` could have additional parameter?

library/CM/MediaStreams/Cli.php
```diff
@@ -18,11 +18,16 @@ public function importVideoThumbnail($streamChannelId, CM_File $thumbnailSource)
...
 
     /**
-     * @param int     $streamChannelId
+     * @param string  $mediaId
      * @param CM_File $archiveSource
+     * @throws CM_Exception_Invalid
      */
-    public function importArchive($streamChannelId, CM_File $archiveSource) {
-        $streamChannelArchive = new CM_Model_StreamChannelArchive_Media($streamChannelId);
+    public function importArchive($mediaId, CM_File $archiveSource) {
```
tomaszdurka: Maybe call it `streamChannelMediaId`?

library/CM/Model/StreamChannelArchive/Media.php
```diff
@@ -180,4 +195,16 @@ public static function deleteOlder($age, $streamChannelType) {
             $streamChannelArchive->delete();
         }
     }
+
+    /**
+     * @param string $mediaId
+     * @return CM_Model_StreamChannelArchive_Media|null
+     */
+    public static function findByMediaId($mediaId) {
```
tomaszdurka: Let's put this into `CM_MediaStreams_StreamRepository`.

tomaszdurka:
Review done.

alexispeter:
please review @tomaszdurka

library/CM/Janus/RpcEndpoints.php
tomaszdurka: @njam this checking goes to strange, paranoic direction, but I think it's ok. I just thought that we should also do that check for `serverId`. What should we do in case we have incoming sub/pub for streamchannel which exists, but it comes from different server? I say throw.

Should we rather not worry about this for this very moment? I understand that this server will send already validated data to clients and they will connect from proper clients.

library/CM/Janus/RpcEndpoints.php
```diff
@@ -9,12 +9,13 @@ class CM_Janus_RpcEndpoints {
...
      * @param string $sessionData
      * @param string $channelData
+     * @param string $mediaId
      * @throws CM_Exception_AuthFailed
      * @throws CM_Exception_AuthRequired
      * @throws CM_Exception_Invalid
      * @throws CM_Exception_NotAllowed
      */
-    public static function rpc_publish($serverKey, $streamChannelKey, $streamKey, $start, $sessionData, $channelData) {
+    public static function rpc_publish($serverKey, $streamChannelKey, $streamKey, $start, $sessionData, $channelData, $mediaId) {
```
tomaszdurka: Can we change signature to:
```
public static function rpc_publish ($serverKey, $sessionData, $channelKey, $channelMediaId, $channelData, $streamKey, $streamStart);
```
Same for subscribe?

library/CM/Janus/RpcEndpoints.php
njam: Generally agree, throwing early in such cases can be useful, to detect problems well *early*.

alexispeter:
please review @tomaszdurka

tomaszdurka:
Looks good

PR #2043, opened 2016-01-08
make CM_Model_StreamChannelArchive_Media::file optional

fvovan:


fvovan:
for https://github.com/cargomedia/sk/issues/2582

fvovan:
@njam please review

njam:
- CM/Model/StreamChannelArchive/Media.php: Usually we tend to make getters fail if the value is unavailable. I think we can cover the use cases with `hasFile()`?
 - Adjust CM/Model/StreamChannelArchive/MediaTest.php accordingly

fvovan:
is it okay to use only `_get()` results check inside getter?
seems a little bit odd to use `hasFile()` there and then enclose all calls to `hasFile` again

njam:
agree, lgtm

PR #2044, opened 2016-01-08
Introduce DelayedQueue for jobs

njam:
Paired with @alexispeter - wanna double check?

@zazabe fyi

### DIFF
--- a/library/CM/Jobdistribution/DelayedQueue.php
+++ b/library/CM/Jobdistribution/DelayedQueue.php
@@ -1,0 +1,52 @@
+<?php
+
+class CM_Jobdistribution_DelayedQueue implements CM_Service_ManagerAwareInterface {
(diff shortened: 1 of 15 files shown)


alexispeter:
revert library/CM/Params.php

rest lgtm

njam:
I thought it can be useful, don't you think?

alexispeter:
But it doesn't work, you would need to pass a getter that uses the streamChannel-factory.

Jut saw another thing: CM_Model_User::offlineDelayed() can be removed.

njam:
Oh true.

The `getStreamChannel()` can be used without a factory, in case of encoding/decoding with ArrayConvertible? I added a test. wdyt?

alexispeter:
Ok, lgtm

PR #2045, opened 2016-01-08
fix CM_MediaStreams_Cli::importArchive()

fvovan:
There are two problems.
- we save full file path in database.
- dot between filename and extension is lost


fvovan:
now it should save values like `20-24365fcd1ff4b2031cc464ecde2ed547-original.mp4`
@njam please review

njam:
@tomaszdurka @alexispeter could you review?

alexispeter:
lgtm

PR #2048, opened 2016-01-12
Add Params Getter for streamchannelDefinition

njam:
@tomaszdurka please review

### DIFF
--- a/library/CM/StreamChannel/Definition.php
+++ b/library/CM/StreamChannel/Definition.php
@@ -1,6 +1,6 @@
 <?php
 
-class CM_StreamChannel_Definition {
+class CM_StreamChannel_Definition implements CM_ArrayConvertible {
 
     /** @var string */
     private $_key;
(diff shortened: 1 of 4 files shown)


library/CM/StreamChannel/Definition.php
```diff
@@ -74,4 +74,19 @@ public function getStreamChannel() {
...
+    public function toArray() {
+        return [
+            'key'         => $this->getKey(),
+            'type'        => $this->getType(),
+            'adapterType' => $this->getAdapterType(),
+        ];
+    }
+
+    public function toArrayIdOnly() {
+        return $this->toArray();
```
tomaszdurka: Is this one valid?

tomaszdurka:
Rest looks good.

library/CM/StreamChannel/Definition.php
njam: `toArray()` should always be a superset of `toArrayIdOnly()`. They can also be the same. Does this answer the question?

PR #2052, opened 2016-01-14
Fix argument casting for Janus rpc endpoint removeStream

fvovan:
https://github.com/cargomedia/cm/blob/master/library/CM/Janus/RpcEndpoints.php#L118


fvovan:
@tomaszdurka please review

tomaszdurka:
lgtm

PR #2054, opened 2016-01-15
StreamRepository::createStreamChannel(): handle concurrent calls

fvovan:
`StreamRepository::createStreamChannel()` is concurrently from multiple RPC calls in `CM_Janus_RpcEndpoints`.
With audio subscribe requests this is quite likely to happen.

What could we do to prevent errors, and ignore/merge subsequent requests and make them atomic?
Those tables are InnoDB…

@tomaszdurka @fvovan @alexispeter thoughts?

### DIFF
--- a/library/CM/Db/Query/Insert.php
+++ b/library/CM/Db/Query/Insert.php
@@ -78,6 +78,6 @@ class CM_Db_Query_Insert extends CM_Db_Query_Abstract {
      * @return bool
      */
     private function _isValidLiteral($literal) {
-        return (1 === preg_match('/LAST_INSERT_ID\((?:id)?\)/', $literal));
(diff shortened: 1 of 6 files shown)


fvovan:
Not sure if this approach is good, but I need to be able to include non escaped values somehow.
Second insert in library/CM/Model/StreamChannel/Media.php could have another dummy `ON UPDATE KEY` field but needs to be inserted as 'literal' again.
`INSERT IGNORE` can't be used there (for 2nd query) as it doesn't care of NULL constraints and inserts records with 
> invalid values are adjusted to the closest values

Weird.
@njam  @alexispeter please review

njam:
weird weird, but lgtm.

Can you also adjust it for the abstract streamchannel?

fvovan:
@njam please review.
I also added a throwing

njam:
lgtm!

can you merge and release?

PR #2059, opened 2016-01-19
Remove publish dependency from media archive

tomaszdurka:
Hash content depends on publish existence.
```
    /**
     * @return string
     */
    public function getHash() {
        if ($this->hasStreamPublish()) {
            return md5($this->getStreamPublish()->getKey());
        }
        return md5($this->getKey());
    }
``


tomaszdurka:
@alexispeter please review

alexispeter:
toArray() should probably not include the user anymore.

alexispeter:
Hash issue fixed in #2062

njam:
@alexispeter could you take over here?

Also related: https://github.com/cargomedia/cm/issues/2061

PR #2058, opened 2016-01-19
Disallow creating streamChannel with mediaId used by an existing archive

fvovan:
`createStreamChannel()` should check that there's no archive with the same mediaId.

Mysql:
- `WHERE NOT EXISTS` - http://dev.mysql.com/doc/refman/5.7/en/exists-and-not-exists-subqueries.html

PHP:
- Check twice: before, and after INSERT

### DIFF
--- a/library/CM/MediaStreams/StreamRepository.php
+++ b/library/CM/MediaStreams/StreamRepository.php
@@ -34,7 +34,6 @@ class CM_MediaStreams_StreamRepository {
      * @param int|null    $thumbnailCount
      * @param string|null $mediaId
      * @return CM_Model_StreamChannel_Abstract
-     * @throws CM_Exception_Invalid
      */
     public function createStreamChannel($streamName, $streamChannelType, $serverId, $thumbnailCount = null, $mediaId = null) {
         if (null !== $thumbnailCount) {
@@ -42,13 +41,7 @@ class CM_MediaStreams_StreamRepository {
         }
         if (null !== $mediaId) {
             $mediaId = (string) $mediaId;
-            $existsMediaArchive = CM_Db_Db::exec('SELECT EXISTS( SELECT 1 FROM `cm_streamChannelArchive_media` WHERE mediaId = ? )', [$mediaId])->fetchColumn();
-
-            if (1 === $existsMediaArchive) {
-                throw new CM_Exception_Invalid('Archive with given mediaId already exists');
-            }
         }
-
         return CM_Model_StreamChannel_Abstract::createType($streamChannelType, [
             'key'            => $streamName,
             'adapterType'    => $this->_adapterType,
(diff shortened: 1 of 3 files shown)


fvovan:
@alexispeter what do you exactly mean by `be valid if created with given values` ? Should it be just discussed checking if `cm_streamChannelArchive_media` with given `mediaId` not exists or should it be the checking similar to `SK_Cam_VideoStreamChannel::isValid()` method, verifying that corresponding session is active i.e extract sessionId from passed channel key  and trying to instantiate it?

alexispeter:
The point is that we check the validity of the streamChannel when publishing/subscribing, but not when creating it.
We think this problem is caused by a new subscriber that connects right as the streamchannel is deleted on the server, but before janus realises that. As a result, the new subscriber "recreates" the streamchannel with the same key and same mediaId, realises that the streamchannel is invalid when trying to create the subscriber and subsequently deletes the whole channel again. 
This could be prevented, if we checked "isValid" before the streamchannel is actually created.

I'm not saying this should be done now, just as a thought for the future.

fvovan:
ah. makes sense. thanks
and for mediaId checking will concurrency unsafe https://github.com/fvovan/CM/commit/a9b14c154748ac1fea20db258f3946e3e368d63f
be sufficient? 
Canonically this issue could be resolved with triggers, I guess we don't use them.

alexispeter:
I thought the idea was to do something like "insert if not exists"?

fvovan:
@njam , @alexispeter could you have a look please.
This ugly SQL statement works, there is just another problem

here `affected_rows` returns 0 if no records were inserted due `cm_streamChannelArchive_media::mediaId` restrictions and if `ON DUPLICATE KEY` was called and 
http://dev.mysql.com/doc/refman/5.5/en/mysql-affected-rows.html
> if an existing row is set to its current values

so to prevent it I need to modify any field (here I tried to increase `createStamp` by 1 it should be changed)

Another option is specifying `CLIENT_FOUND_ROWS ` flag for mysql connection but I'm not sure if it can affect something else.

`LAST_INSERT_ID()` is useless here because it returns inappropriate value if  `cm_streamChannelArchive_media::mediaId` "constraint" hitted.

So I propose just to create "counter of duplicated queried" column.
wdyt?

njam:
oh!

Isn't this dangerous in case we lose connection inbetween?
```sql
$id = CM_Db_Db::exec("SELECT LAST_INSERT_ID()")
```

Maybe we should consider alternatives?
- Combine streamChannel and archive tables
- Do the check in PHP (in `createStreamChannel()` check for existing archives *after* creating the channel, then delete it again)

fvovan:
I would prefer PHP checking then.

>Isn't this dangerous in case we lose connection inbetween?

it's possible to keep `$client`

njam:
Closing in favor of https://github.com/cargomedia/cm/issues/2064 and https://github.com/cargomedia/cm/issues/2056

PR #2063, opened 2016-01-19
Sort archives by ascending createStamp when fetching by key

alexispeter:


alexispeter:
Archives are currently displayed horizontally from youngest to oldest, which might be confusing.

please review @njam

njam:
lgtm

PR #2064, opened 2016-01-20
StreamChannel creation: Check for existing archives with same mediaId

fvovan:
Use `findByMediaId()` and throw.

Similar to https://github.com/fvovan/CM/commit/a9b14c154748ac1fea20db258f3946e3e368d63f

cc @tomaszdurka @alexispeter


fvovan:
@njam please review

njam:
Could you include the mediaId in the exception's metaInfo - can be useful for debugging.

Otherwise lgtm.

PR #2056, opened 2016-01-20
Handle concurrent streamchannel *archive* creations?

fvovan:
Could we something similar like in https://github.com/cargomedia/cm/pull/2054, but for streamchannel *archives*

ON DUPLICATE KEY UPDATE


fvovan:
@njam please review.

I investigated #2057 related possible problem while was resolving this issue.
Is it okay if we fail when sequently creating two `CM_Model_StreamChannel_Media` with different keys but same `mediaId` field. I guess it should never happen, right?

njam:
lgtm!

@tomaszdurka can you answer Vladimir's question?

njam:
@tomaszdurka can you answer Vladimir's question?
 Show all checks

tomaszdurka:
Should never happen.

PR #2067, opened 2016-01-21
Implement copy from "local" to "s3" filesystem without reading content into memory

fauvel:
As discovered with @tomaszdurka 
Currently `CM_File::copyToFile(CM_File $file)` will read the complete file contents into memory to write them to the new location, if the two filesystems of the operation are different.

If we copy a big file from *local* to *s3* filesystem this can easily lead to memory exhaustion of PHP.

Should we introduce a special `copyFromLocal()` which could be optionally implemented by some file systems (defined with an interface, similar like we do it for `CM_File_Filesystem_Adapter_AppendInterface`).

@fauvel could you look into this?

### DIFF
--- a/library/CM/File.php
+++ b/library/CM/File.php
@@ -184,9 +184,18 @@ class CM_File extends CM_Class_Abstract implements CM_Comparable {
      * @param CM_File $file
      */
     public function copyToFile(CM_File $file) {
-        $sameFilesystemAdapter = $this->_filesystem->equals($file->_filesystem);
-        if ($sameFilesystemAdapter) {
+        $filesystemSource = $this->_filesystem;
+        $filesystemDestination = $file->_filesystem;
+        $adapterSource = $filesystemSource->getAdapter();
+        $adapterDestination = $filesystemDestination->getAdapter();
+        if ($filesystemSource->equals($filesystemDestination)) {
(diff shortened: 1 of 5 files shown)


fauvel:
@njam could you please review and do some acceptance testing with our S3 account (sandbox?)

fauvel:
@njam please review

njam:
- Shouldn't the call to `registerStreamWrapper()` go into the constructor of `CM_File_Filesystem_Adapter_AwsS3`? Otherwise one could instantiate an adapter that's not functional. Can the function be called repeatedly?

Can you do some testing on cargomedia-playground?

fauvel:
I'm not sure we should call `registerStreamWrapper` several times – it unregisters and registers the stream wrapper again (the rationale behind this is that you potentially want to use different credentials and a different S3 client, I guess), so it would probably close stream connexions. @tomaszdurka wdyt?

njam:
Can we check if it's registered? Would be good if the adapter also works independent of the factory. Otherwise if the overhead is not too big I would just call it again?

fauvel:
@njam please review

njam:
lgtm

fauvel:
Tagged 1.166.2

PR #2069, opened 2016-01-21
Prevent browser from caching page loads, even in history

njam:
Usually our pages load via an AJAX POST request, so they're never cached.
But if the next page has a different layout than the current one, the whole page will be refreshed.
In this case we want to also avoid that the browser caches the page load.

@zazabe @fauvel please review


zazabe:
acceptance test passed, lgtm!

PR #2065, opened 2016-01-21
Store media channel thumbnails in separate table

alexispeter:
Remove `thumbnailCount` in favor of a table containing thumbnails. Each thumbnail references streamChannel or streamChannelArchive (same ID).

`CM_MediaStreams_Cli::importVideoThumbnail()` should expect the `createStamp`.

Table name: `cm_streamChannel_mediaThumbnail`?

Also let's drop `thumbnailCount` and `data`.

cc @tomaszdurka @fvovan

PR #2070, opened 2016-01-21
Optional $createStamp for video thumbnails

njam:
In preparation for https://github.com/cargomedia/cm/pull/2065

@tomaszdurka please review


njam:
@tomaszdurka please review

@alexispeter fyi

tomaszdurka:
Create stamp will be optional now, so you will pass it as `--createStamp` parameter, right?

njam:
True, changed it!

tomaszdurka:
lgtm

alexispeter:
I'm a bit confused by the comments and the commit-history in here, was it manipulated? So should the createStamp be optional or not? Comment suggest so, but the code says otherwise.

PR #2072, opened 2016-01-22
Throw errors of createAudioPlayer

vogdb:


njam:
Are you sure we get an `err` as an argument? Is this code really running synchronously?
Did you test this with an unsupported file?

vogdb:
> Did you test this with an unsupported file?

No :(. Fixing now

> Are you sure we get an `err` as an argument?

no. I relied on a common practice. I was wrong. But here someone had decided to be an artist and create its own error mechanism :(

njam:
Maybe this would be a good opportunity to also upgrade to the newest version 2.19.0?
https://github.com/johndyer/mediaelement/blob/master/changelog.md

vogdb:
It's quite a trick to extract the reason of error. Created an issue on the mediaelement repo johndyer/mediaelement#1678.

vogdb:
@njam please look

vogdb:
@njam please wait with the review

njam:
So that error will abort `createAudioPlayer()`?
In that case looks good.

The only concern I have is in SK we will now fail to set up some components if we can't load the audio. But that might be okay?

vogdb:
> So that error will abort `createAudioPlayer()`?

Not anymore. Just log it now. I think it will be good to change `notification.mp3` to `ogg`? It's the only mp3 sound that I see in our repo.

njam:
lgtm, using OGG sounds good, but we should check that it works everywhere (what about IE?). Maybe we need to provide 2 files, can it be done with MediaElements?

vogdb:
> can it be done with MediaElements?

It can be done with Modernizr which already have tests for those audio types

njam:
Or use html's `<source>`: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/source

vogdb:
@njam I think we should do those checks outside of `createAudioPlayer`. It's not part of its logic. We should check for playability of extension and call `createAudioPlayer` with the corresponding extenstion. Wdyt?

njam:
Agree, it should not care. If we use `<source>` I guess MediaElement will handle it and pick the right source?

vogdb:
@njam please review

njam:
Maybe we can always use `<source>`, and convert string to an array, to simplify the code further?
Did you test it, does it work good?

vogdb:
> Did you test it, does it work good?

Works fine. Please review

njam:
- I would still allow to pass `String|String[]`, but cast it internally to array? This way the change is not breaking, and the API is more flexible?
- You removed the overwriting of `play` and `pause` with fake functions on error. Did you do that on purpose? We should decide if we want the playback to be optional (based on browser support), or mandatory.
 - If optional: Only log error, make play/pause a no-op
 - If mandatory: Fail if there's not playback support.

I'd say let's keep it optional, to be safe?

vogdb:
@njam please review

njam:
lgtm

Issue #2071, opened 2016-01-22
Reduce "cache warmup" time


njam:
Let's analyze what times how much time, then optimize.

cc @tomaszdurka

njam:
I have 2 languages and 18 sites in my app.
I've instrumented the code like this: 

Collecting the files and classes takes: 16 seconds
Total: 182 seconds

Top offenders:
```
CM_Asset_Javascript_Library: 0.98693490028381s
CM_Asset_Css_Vendor: 0.98997187614441s
CM_Asset_Css_Vendor: 0.99139094352722s
CM_Asset_Css_Vendor: 0.99153304100037s
CM_Asset_Css_Vendor: 0.99715995788574s
CM_Asset_Css_Vendor: 1.0091688632965s
CM_Asset_Css_Vendor: 1.0141839981079s
CM_Asset_Css_Vendor: 1.0179049968719s
CM_Asset_Css_Vendor: 1.0399208068848s
CM_Asset_Css_Vendor: 1.0594129562378s
CM_Asset_Css_Vendor: 1.0697989463806s
CM_Asset_Css_Vendor: 1.0915310382843s
CM_Asset_Css_Vendor: 1.0948178768158s
CM_Asset_Css_Vendor: 1.1067910194397s
CM_Asset_Css_Vendor: 1.1197698116302s
CM_Asset_Css_Vendor: 1.1203911304474s
CM_Asset_Css_Vendor: 1.1223139762878s
CM_Asset_Css_Vendor: 1.1291592121124s
CM_Asset_Css_Vendor: 1.133385181427s
CM_Asset_Css_Vendor: 1.1360440254211s
CM_Asset_Css_Vendor: 1.1479368209839s
CM_Asset_Css_Vendor: 1.1710250377655s
CM_Asset_Css_Vendor: 1.2857460975647s
CM_Asset_Css_Vendor: 1.3026688098907s
CM_Asset_Javascript_Library: 1.3823390007019s
CM_Asset_Css_Vendor: 1.4917569160461s
CM_Asset_Css_Vendor: 1.9127612113953s
CM_Asset_Css_Library: 2.7346150875092s
CM_Asset_Css_Library: 2.7667880058289s
CM_Asset_Css_Library: 2.8118410110474s
CM_Asset_Css_Library: 2.8172919750214s
CM_Asset_Css_Library: 2.8305299282074s
CM_Asset_Css_Library: 2.8315510749817s
[...]

PR #2073, opened 2016-01-22
Remove unneeded variables from CSS cache key

njam:
siteId and languageId are *not* used to compile the CSS, so they can't affect it.

### DIFF
--- a/library/CM/App.php
+++ b/library/CM/App.php
@@ -38,18 +38,16 @@ class CM_App implements CM_Service_ManagerAwareInterface {
 
         /** @var CM_Asset_Javascript_Abstract[] $assetList */
         $assetList = array();
-        $languageList = new CM_Paging_Language_Enabled();
         foreach (CM_Site_Abstract::getAll() as $site) {
+            $render = new CM_Frontend_Render(new CM_Frontend_Environment($site));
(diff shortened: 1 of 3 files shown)


njam:
Closes #2071

@dlondero @alexispeter please review

dlondero:
@njam lgtm

njam:
@tomaszdurka how is Laax? up for a review?

tomaszdurka:
It's great, thanks!

I have looked into this code too when saw the 10 minutes warmup. Also thought about removing this loop. I think the idea before was *right* as we probably should consider all the possibilities of render which is used in less callbacks. On the other hand we can analyse if language has influence on them (`getUrlResource`, `getUrlStatic`) and seem that not. Possibly in the future we could change implementation of `getUrl*` to use language (url translation, etc.) and then it would fail. Still we should not care now.

So - looks good!

njam:
There's `md5($content)` in the cache key, so we will definitely re-compile if the language has any impact!

tomaszdurka:
Well if we won't include it into render then we won't :) Definitely good change to simplify the key with md5 only. I agree to language removal too!

njam:
True! But we would re-compile later, let's hope it doesn't happen.

PR #2074, opened 2016-01-26
Introduce streamchannel DisallowInterface

alexispeter:
See cargomedia/sk#2769

> Introduce `DisallowInterface`:
- canSubscribe, canPublish, isValid
- Move isValid check for `on*()` into the specific `on*()` functions
- Move `isValid()` call out of stream creation into RPC functions?
- In RPC endpoints delete streamchannel on "InvalidException" if we just created it (i.e. it's empty):
```php
            if (!$streamChannel->hasStreams()) {
                $streamChannel->delete();
            }
```


alexispeter:
please review @njam

alexispeter:
I decided to keep the isValid() check inside the model because it's used in accord with canSubscribe(), which used to be outside and was moved inside at some point.

njam:
@fvovan could you review this?

fvovan:
Weren't we going to make new interface implemented starting with `CM_Model_StreamChannel_Media` ?
Please fix the docu for Interface's methods `$user` param

alexispeter:
Whether or not a streamChannel can disallow connections depends primarily on it's adapter-type, so we can't implement it on streamChannel_media, since the adapter is mutable. 
We always implement a new child to streamChannel_media anyway, so I think it's better to implement it on the implementation instead of the quasi-abstract media-streamChannel.

fvovan:
just was a little bit confused with 
`/** @var CM_Model_StreamChannel_Media|CM_StreamChannel_DisallowInterface $streamChannel */`
well, if `StreamChannel_Media`'s childs can have various adapters, it's okay.
btw do we use `streamChannel_media` concrete instances? 

One more note 
Shouldn't the signature for `CM_Model_Stream_Abstract::setAllowedUntil` be changed to allow nulls as well?

alexispeter:
I don't think this makes much sense. `allowedUntil` is only used for disallowed streamChannels, and setting it to null for such a streamChannel would set it up for termination by the next checkStreams-call.

fvovan:
ok, lgtm then

njam:
Small notes:
- RpcEndpoints.php: rpc_publish: I think we should only delete if the channel is empty. (because we should never enforce state from CM's side, except for state that we just created in the rpc call)
- DisallowInterface.php: Add "public" visibility keyword for functions

alexispeter:
please review @njam

njam:
lgtm

PR #2076, opened 2016-01-28
Breaking: Upgrade to Bluebird 3.x

vogdb:
See http://bluebirdjs.com/docs/new-in-bluebird-3.html
Especially: http://bluebirdjs.com/docs/api/cancellation.html

@vogdb when you find some time, let's upgrade?
cc @zazabe

### DIFF
--- a/client-vendor/after-body/bluebird/bluebird.js
+++ b/client-vendor/after-body/bluebird/bluebird.js
@@ -24,8 +24,8 @@
  */
 /**
  * bluebird build version 3.1.5
- * Features enabled: core, race, props, settle, timers
- * Features disabled: call_get, generators, map, nodeify, promisify, reduce, some, using, filter, any, each
+ * Features enabled: core, race, props, timers
+ * Features disabled: call_get, generators, map, nodeify, promisify, reduce, settle, some, using, filter, any, each
 */
 !function(e){if("object"==typeof exports&&"undefined"!=typeof module)module.exports=e();else if("function"==typeof define&&define.amd)define([],e);else{var f;"undefined"!=typeof window?f=window:"undefined"!=typeof global?f=global:"undefined"!=typeof self&&(f=self),f.Promise=e()}}(function(){var define,module,exports;return (function e(t,n,r){function s(o,u){if(!n[o]){if(!t[o]){var a=typeof _dereq_=="function"&&_dereq_;if(!u&&a)return a(o,!0);if(i)return i(o,!0);var f=new Error("Cannot find module '"+o+"'");throw f.code="MODULE_NOT_FOUND",f}var l=n[o]={exports:{}};t[o][0].call(l.exports,function(e){var n=t[o][1][e];return s(n?n:e)},l,l.exports,e,t,n,r)}return n[o].exports}var i=typeof _dereq_=="function"&&_dereq_;for(var o=0;o<r.length;o++)s(r[o]);return s})({1:[function(_dereq_,module,exports){
 "use strict";
@@ -182,7 +182,7 @@ Async.prototype._reset = function () {
 module.exports = Async;
 module.exports.firstLineError = firstLineError;
 
-},{"./queue":18,"./schedule":20,"./util":25}],2:[function(_dereq_,module,exports){
+},{"./queue":18,"./schedule":20,"./util":24}],2:[function(_dereq_,module,exports){
 "use strict";
 module.exports = function(Promise, INTERNAL, tryConvertToPromise, debug) {
 var calledBind = false;
@@ -391,7 +391,7 @@ Promise.prototype._resultCancelled = function() {
 
 };
 
-},{"./util":25}],5:[function(_dereq_,module,exports){
+},{"./util":24}],5:[function(_dereq_,module,exports){
 "use strict";
 module.exports = function(NEXT_FILTER) {
 var util = _dereq_("./util");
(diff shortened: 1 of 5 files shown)


vogdb:
@zazabe please check VR part

zazabe:
ok, testing...

zazabe:
wooow, plenty of warning!

vogdb:
Yep but they were before :)

njam:
Does it still reject with `Promise.CancellationError`?

vogdb:
Yes. You can try both variants:
```js
    var promise = new Promise(function(resolve, reject) {
      reject(Promise.CancellationError());
    }).catch(function(error) {
      console.log('#####', error instanceof Promise.CancellationError);
    });
```
```js
    var promise = new Promise(function(resolve, reject) {
      throw Promise.CancellationError();
    }).catch(function(error) {
      console.log('#####', error instanceof Promise.CancellationError);
    });
```

vogdb:
or better those variants for 100%
```js
    var promise = new Promise(function(resolve, reject) {
      reject(Promise.CancellationError());
    })
      .then(function() {
        console.log('$$$$$$$$$$$$');
      })
      .catch(function(error) {
        console.log('#####', error instanceof Promise.CancellationError);
      });
```
```js
    var promise = new Promise(function(resolve, reject) {
      throw Promise.CancellationError();
    })
      .then(function() {
        console.log('$$$$$$$$$$$$');
      })
      .catch(function(error) {
        console.log('#####', error instanceof Promise.CancellationError);
      });
```

zazabe:
yes but from what i've tested, `cancel()` is sync now and has no argument, cancelling with this method will not call `catch` anymore.

vogdb:
@zazabe How about that? Should be more consistent now.

njam:
So on `cancel()` the promise will *not* execute `catch()` behaviour, but it *will* execute `finally()` code?

If that's the case, then we should make sure we're using `finally()` wherever we expect a promise to be fulfilled sooner or later. For example `jquery.clickDecorators/decorators/30-spinner.js` (which uses `finally()`).

vogdb:
Also one interesting thing:

> Note that it is an error to consume an already cancelled promise, doing such a thing will give you a promise that is rejected with new CancellationError("late cancellation observer") as the rejection reason.

So, I would keep the handling of CancellationError in error handlers. @njam ?

zazabe:
@vogdb cargomedia/sk#2728 is working well with your branch now, VR Components are okay but we still have warning during the feature detection step, mostly because many promises are rejected without an `Error` argument.

njam:
So *are* cancelled promises going through `unhandledrejection`? If so, then I think we should keep the handling of `CancellationError`.

vogdb:
> So *are* cancelled promises going through `unhandledrejection`? If so, then I think we should keep the handling of `CancellationError`.

Yes. The thing is that now they are generated less than before cause direct `.cancel` call does produce `CancellationError`.

njam:
I see!

```js
Promise.config({cancellation: true});
var p = new Promise(function() {})
 .then(function(){ console.log('then'); })
 .catch(function(){ console.log('catch'); })
 .finally(function(){ console.log('finally'); })
p.cancel();

finally
```

njam:
- promiseThrottler: Why remove `cancelTrailing`?
- CM/Layout/Abstract.js: _loadPageThrottled: I think here we need to introduce `finally()` (since the behaviour changed we need to make sure that code that needs to be executed no matter what, also on cancellation, should go into a `finally`, right?)
- CM/App.js: ajax: In `fail()` we potentially `reject` the promise twice - should that second part go into an `else`? [this is only vaguely related]

Let's also get Michel's opinion here.

vogdb:
> - promiseThrottler: Why remove `cancelTrailing`?

The used approach was appropriate while we could cancel promise with CancellationError. Now cancel does not reject. Thus we can't use rejection to cancel. So, I removed it cause couldn't think of solution that can cancel a trailing promise. Plus this option wasn't used.

vogdb:
@njam @zazabe please look

zazabe:
lgtm!

I just have one concern about the bluebird `cancel` implementation, which is sync now: a cancelled promise is directly settled and there's no way to know if the resolver code is still running or not...

It probably make sense... but it would be good also if it was possible to wait the end of the cancelled promise before doing something else...

With the `promiseThrottler` we have to be careful and check in `finally` if the promise has been cancelled, then clean up everything to ensure that the current call will not be impacted by a previous one...


```
var promise = new Promise(function(r) {setTimeout(function(){console.log('foo');r();}, 1000)}); 
promise.finally(function() {console.log('finally', promise.isFulfilled(), promise.isPending())}).cancel()
> finally false false
undefined
> foo
```

vogdb:
Very good catch!

vogdb:
@zazabe @njam please review. Should be fine now

zazabe:
lgtm

@vogdb maybe not related to this ticket, but I thought about some simplification: https://github.com/zazabe/cm/commit/0953f9ff2c4b69d2e7beb6accfe9af4ed2d54523

My acceptance test:
```javascript
var foo = function(text, timeout) { return new Promise(function(r) {setTimeout(function() {console.log(text); r();}, timeout)})}; 
var step = function(text) { return function() { return foo('step ' + text, 1000);}; };

var throttle = promiseThrottler(function(text) { return step(text + ' 1')().then(step(text + ' 2')).then(step(text + ' 3')); }, {cancelLeading: true});

throttle('a'); throttle('b'); throttle('c'); 
> step c 1
> step c 2
> step c 3

var throttle = promiseThrottler(function(text) { return step(text + ' 1')().then(step(text + ' 2')).then(step(text + ' 3')); }, {cancelLeading: false});

throttle('a'); throttle('b'); throttle('c'); 
> step a 1
> step a 2
> step a 3
> step b 1
> step b 2
> step b 3
> step c 1
> step c 2
> step c 3
```

zazabe:
@vogdb I forget the `queue: false` options... so the code will not be much simpler anyhow...

I think we can merge it like it is ;)

PR #2078, opened 2016-01-29
button_link: Use "<a>" tag for "page" param

njam:


njam:

christopheschwyzer:
What about the `onclick` case - I think we don't use it currently, so lgtm.

njam:
We don't use it - and I think we shouldn't (rather bind behaviour from JS files).

PR #2081, opened 2016-02-02
Logger integration into CM

fvovan:


fvovan:
I have a question about how we are going to store exceptions. 
Currently they go to `cm_log` table, if i'm not mistaken now they will be collected into Mongo, right?
It could be done as `mysql` handler as well, but then we end up to create too specific handlers because of schemed storage. Probably it could be reasonable idea that handlers for general logs and exceptions logs will be different (so loggers will be different) as well, but if we decide to migrate all these things to mongodb it could be simplier.

njam:
I think everything should go to 1 logger, with the same set of handlers (mongodb).
We just need to change the admin UIs to render logs from there, but we need to do it anyway.
Or am I missing something? Would there be an upside to store it in mysql?

fvovan:
I don't blame mysql :) it's just the matter of concept. I thought about separation but it's just an idea and If we want to keep mysql  we can use scheme for CM_Log_Record_Exception scheme as more specific.

njam:
Then let's stick to mongodb only? Will be simpler and should get the job done (and maybe more powerful)?

fvovan:
okay, sounds good.
Frankly I'm a bit lost with how we were going to perform "backup logging". We discussed the conception several times few months ago and it would be good to get a summarised opinion.
Currently (I mean this dev branch) we have 2 of so called "log pipelines"
1. MongoDB for INFO level and stderr for ERROR messages. So error messages will be duplicated to stderr and I think that's as should be.
2. Second pipeline for "Fallback logger" and here it has another "stdError or stdOut" stream handler and logs/error.log handler. They are completely duplicated with the exception that error.log handler will doesn't work on early stages (before CM_Service_Filesystems initialisation)

So things are duplicated. If we get a handling exception somewhere in fists log pipeline, we still get original message processed in following handlers of 1st pipeline, and put it to all handlers of second fallback pipeline.

Not sure if we wanted it that way. On the one hand fallback stderr handler is the only more or less failsafe here. On the other when logs/error.log is available we don't need to mess stderr with additional bunch of lines.
 Besides fallback logger currently used as html error printer and it's obviously a misusage.

Another issue is we need to move this code which outputs exceptions to HTML (formatted exception plus 500 code).
I am thinking about separate handler which will handle only when non CLI context is applied. Probably this handler should be first (we need to send headers ASAP) and mandatory.

So in the end there will be still 2 pipelines
1 General: 
html (level warn) -> mongo (level info) -> stderr (level warn) 
or 
mongo -> html (stop if was used) -> stderr 
2 Backup file (stop if was used)  -> stderr :

wdyt?

fvovan:
One more question is how to combine js and "general" exceptions.
As i understand from https://developer.mozilla.org/ru/docs/Web/API/GlobalEventHandlers/onerror only 
message, line and file can be accessed at `window.onerror()` handler? Even stacktrace is not reachable there (because there is no `Error` object)
I think it would be good  to pass `window.location` along with additionally. 
But in the end .. will it be an exception record for logger?

I propose to define new CM_Exception_Javascript class and pass these 3  attributes (message, line, file ) as metainfo.
And then it will be possible to distinguish js errors from others. We want to keep separation (by severity or by exception class) in the admin UI, don't we?

@njam wdyt?

njam:
> One more question is how to combine js and "general" exceptions.

Yes, currently we don't record stack traces of javascript errors. But it is possible for some browsers. We shouldn't do it right now, but let's keep it in mind.
Therefore I think we should have the ability to store a stacktrace for a JS error in the log. That's why I thought it's good to use `CM_ExceptionHandling_SerializableException` - there we can store the stacktrace.

Regarding *html* and *cli* output. The solution you proposed should work. I would put "mongo" first, but probably doesn't matter.
Alternatively we could also consider catching errors directly in `CM_Http_Handler` and `CM_Cli_CommandManager`.
wdyt? @tomaszdurka any input?

fvovan:
>Therefore I think we should have the ability to store a stacktrace for a JS error in the log.

is it just `Error.stack()` that should be passed as url param in https://github.com/cargomedia/cm/blob/master/client-vendor/after-body/00-jserror/jserror.js#L58 ?
using `CM_ExceptionHandling_SerializableException` is good idea but currently it requires an `Exception` instance.  Ok actually it's not a problem to change it.

>Alternatively we could also consider catching errors directly in CM_Http_Handler and CM_Cli_CommandManager.

Hmm. Do you mean to use `try` there and then print error message and  rethrow to invoke global exception handler?

What should changed in admin Log menu? As for me it should be preserved.

njam:
> Hmm. Do you mean to use try there and then print error message and rethrow to invoke global exception handler?

Exactly, instead of installing a global error handler. But maybe that's for later?

> What should changed in admin Log menu? As for me it should be preserved.

Agree

tomaszdurka:
Imo (sorry if it covers or not covers what you wrote):
- we should probably have general exception logger which would log exceptions regardless of environment they are thrown in (mongodb, file, mysql, file, console, etc).
- we should have individual exception handlers for environments (cli, http) which would use above loggers 
- additionally we should have exception handlers (http, cli) on top level which would properly exit exceptions (print, send headers, send status code, whatever needed)
- this means html wouldn't be a logger adapter, but something completely separate
- should we have backup as fallback or rather one of the logger "adapter"s?

njam:
I agree, let's rather add the handling for http and cli errors into `CM_Http_Handler` and `CM_Cli_CommandManager`. The chance that an error happens *before* the handlers is pretty small, and in that case it's okay to have not-so-userfriendly feedback from some default handlers.

fvovan:
probably  it  was a bad idea to combine logger and exception handler, 
they definitely should be separated.

>additionally we should have exception handlers (http, cli) on top level which would properly exit exceptions (print, send headers, send status code, whatever needed)

Not sure if I understand it. What do you mean by "top level"? If we have individual exception handlers then we can gracefully exit and print there.

>should we have backup as fallback or rather one of the logger "adapter"s?

The idea was to have backup mainly to collect handlers loggers errors.

tomaszdurka:
Main thought: **Separate exception handler from logger**

#### Handling uncaught exceptions
- Revert removal of cli/http exception handlers

##### Html
- Log exception (using Logger service)
- Print html-formatted exception
- Exit

##### Cli
- Log exception (using Logger service)
- Print text-formatted exception
- Exit

#### Logging
- Log_Logger as service (public api for logging)
- Log_Handler_Interface (Adapter)
- Move bubbling outside from handlers (or remove)
- Remove `CM_Log_Formatter_Html` (use its parts in Http exception handler)
- Remove `Log_Handler_Stream_Http/Cli`
- Remove `Log_Formatter_Html/Abstract`

:panda_face:

fvovan:
It's almost ready :)
Bubbling was moved to logger level and renamed to "propagation", only 1 non backup handler (mongo) was left, file and stderr are only used as fallback.
admin UI is still in progress but I think template changes will be not bound to whole implementation.

So now I have few more questions.
1) To preserve `CM_Paging_Log` structure I had to add an extra `$type` parameter to all logger methods. So now every log message has two orthogonal but similar attributes: level and type. Is that okay? One more issue is how to deal with untyped records? Do we want to display them? If yes I can add "default paging".. It's only related to "general" records, not to containing exceptions. They already have FATAL as default.

2) As I mentioned on Monday we decided to extract bubbling logic from handlers. It caused creation of "config structured" like ```[['name' => 'mongo-handler', 'propagation' => true], ['name' => 'file-handler', 'propagation' => false]]```. Maybe it would be reasonable to change it with typed structure, possibly not for using in service  config but rather as a passed "data structure" , I'm not 100% sure.

3) Is there a way to easily print arrays for smarty? I'm trying to make admin log template page more or less record structure agnostic.

@njam , @tomaszdurka  wdyt?

njam:
> To preserve CM_Paging_Log structure

I would rather introduce a new structuring, based on level. The logs would change like this:
- fatal -> fatal
- error -> error
- warning -> warning
- not_found -> warning
- js_error -> [whatever level we chose, probably warning?]
- mail -> debug
- payment_callback -> debug

Some entries that are currently separate would be combined. But with "Sentry"'s ability to search we can easily find those entries again. For debug entries we should then have an extra field `type` set to "mail" or "payment callback".
wdyt?

> Is there a way to easily print arrays for smarty?
- `CM_Debug_VariableInspector` is pretty good :)
- And smarty has some built-in stuff: http://www.smarty.net/forums/viewtopic.php?t=11435

fvovan:
Agree with `warning` level for JS. In the future can be adjusted  to `Error` for some very critical errors though.
Not sure `type` is needed for mail or payments then. For 'tagging' we can just use message prefix there.

njam:
I think a type-tag would still be good for payment callbacks. It will make it very easy to filter in Sentry, and there's no downside of it?

fvovan:
well it depends on what we mean be `type`, we can't  just leave existing `type` only for payment logs, because log records are not separated to 'payment' and 'non-payment'. So in this case it should be left for all remaining kinds of records (including exceptions).
But if we speak about just a key for `context->extra` array, then it's okay.

tomaszdurka:
1) I agree with adding `type` (payment, mail) on context level. Let's keep debug level for these. 
Btw can we biuld single hash-like index on mongodb or we need to add index for each context field we want to use in more advanced way?

tomaszdurka:
2) Still not sure about the instance. I think propagation within Handler instance doesn't say anything and it would be no different for this new structure. It doesn't make any sense without context of Logger service, does it?

njam:
I meant to add a `type` to context's **extra**.

fvovan:
>Btw can we build single hash-like index on mongodb or we need to add index for each context field we want to use in more advanced way?

I think we can't but anyway it would search only 'precise matchings', what is not useful for contexts.

fvovan:
>Still not sure about the instance. I think propagation within Handler instance doesn't say anything and it would be no different for this new structure.

It's not within Handler instance anymore. Propagation is defined on `CM_Log_Logger` level.

tomaszdurka:
As talked with Reto briefly over hipchat yesterday I was thinking what exactly is bubbling mechanismfor and how to remove/replace it.

As I understand we to have couple of levels of loggers which fallback to another level if failed. This levels of loggers though can consist of multiple loggers like sentry+mongo.


### Proposition 1
Create "logger composition" which would consist of multiple loggers, it would also implement logger interface. This kind of logger would consist of multiple loggers instances and would have option "continue-on-success" or similar. Code example:
```
$firstLogger = new LoggerComposition(['mongo', 'sentry'], ['continue-on-success' => true]); // will try to log to all no matter what, would fail if all fail
$secondLogger = new FileLogger();
$mainLogger = new LoggerComposition([$firstLogger, $secondLogger],  ['continue-on-success' => false]); // would log to first and to second if first succeeded
```


### Proposition 2
Similar but implement fallback mechanism.
```
$firstLogger = new LoggerComposition(['mongo', 'sentry'], ['continue-on-success' => true]); // will try to log to all no matter what, would fail if all fail
$secondLogger = new FileLogger();
$firstLogger->setFallbackLogger($secondLogger);
```


### Proposition 3
Use layers concept instead of propagation in our current Logger service. 
```
$mainLogger = new Logger([
  ['mongodb', 'sentry'],
  ['file'],
]);
```

I understand you @njam was after proposition 1 or 2 (not sure now :panda_face: ).
@vogdb wdyt?

vogdb:
Please elaborate more on **Proposition 3**.

tomaszdurka:
Sorry again Aleksei! I meant to mention @fvovan :panda_face:

fvovan:
Frankly, *Proposition1* is not really clear to me , 
on first string you compose handlers, but on third you combine loggers.
But 1st string actually describes actual `CM_Log_Logger` implementation i.e. combination of handlers

My 5 cents are that backup mechanism matters only for Exception logging thus to not overcomplicate things it should be just another logger (in terms of combination of handlers). As for bubbling/propagation, mostly it matters only for backup mechanism. Intention was to write for ex. to file but if filesystem is not ready or unavailable then to print message to STDERR.

Since we reverted exception handlers with printers,  bubbling/propagation can be simply deleted, so we will write to all handlers. If error occurs during any handler writing - logger exception will be raised (after other handlers writing attempts) and if the same handler fails again during - backup will be obtained and used. And "backup" could be just file logger.

tomaszdurka:
Regarding proposition 1)
Yes I thought about having the same interface for both logger and logging handlers. Now I see they different one. I thought they should have single, simple log($record) interface. Currently we basically map Record's constructor into the logger->log method.
Imho creating separate interfaces for handler and logger might overcomplicate rather than simplify.

I think backup is not only for exceptions. Backup is very specific and honestly it should almost never happen. It would and should be triggered the same for debug messages as for exceptions.

Not sure if I understand correctly throwing not-possible-logging exception. We need to wrap that with try-catch block and handle in the logger. Understand that it would be caught and logged with backup?

@fvovan let's schedule a call and talk about. I think text might miss a lot of points.

fvovan:
>Backup is very specific and honestly it should almost never happen. 

it will happen because some of our handlers will be 3rd part remote services

>It would and should be triggered the same for debug messages as for exceptions.

It will be triggered but later (debug handler will raise an exceptions which should be logged as exception, but since our handler doesn't work  - backup will be used). Actually that's the point of using backup handler - to not get into loop.

tomaszdurka:
- Move backup concept to `Logger`
- Use either backup logger, or multiple layers of handlers
- All `Loggers` are registered as services (and `Handlers`?)
- Only use backup logger (next layer) if all exception handlers failed AND it's not `HandlingException`

library/CM/Log/Record/Exception.php
```diff
@@ -7,12 +7,17 @@ class CM_Log_Record_Exception extends CM_Log_Record {
 
     /**
      * @param Exception      $exception
+     * @param int|null       $logLevel
```
fauvel: Adjust order

fvovan:
Discussed layer system is ready. 
With the following workflow:
- Messages are written to all handlers of the layers, not dependent if there were error at some of them
- If error(s) occurred during exception handling process then it is logged to remained working handlers
- In case ALL handlers of given layer gave errors during MESSAGE (Message can be either message or some exception thrown outside of log handlers) processing this Message is sent 1 layer down. All handler exceptions are sent there as well.
- in case if ANY error happened during logging handler exception it goes 1 level down.
- when no remained layer is available - it silently returns

There few more questions
1) I set up default log level for native exceptions to CRITICAL, is it good?
2) Not sure how to do Mail and Payment logs better. For now they are just written as info messages but i guess it would be better to have separate list for them in admin
3) Not sure how this layered system will work with new relic. The problem is it doesn't notify if something goes wrong. And that means when NR handler presents in handler list no down layers will be used as recordings to NR will be succeed
4) There is one very very weird test in  `CM_Http_Response_AbstractTest`  for me.
When I run only `CM_Http_Response_AbstractTest::testRunWithCatching` it works, but failing when was run together with other test cases despite `tearDown()` is on its place. Had to reorder test cases for now but it should be a reason..

@tomaszdurka wdyt

fvovan:
made adjustments to 
> in case if ANY error happened during logging handler exception it goes 1 level down.

now it writes handler exception to all of layer's handlers and if no one was working, goes to another level.
NOTE: it does not add these "new" exceptions to existing list but just pass them down.

tomaszdurka:
1) Seems reasonable. Not sure what is the difference between critical, emerg, error, etc :)
2) As we said, there will be different `type` for payments/mails and those pagings should be based on that.
3) Yes, as discussing in hipchat
4) Please investigate why it's "kinda-flaky"

fvovan:
> 1) Seems reasonable. Not sure what is the difference between critical, emerg, error, etc :)

It's the strictest and means that will be written to all handlers.

>2) As we said, there will be different type for payments/mails and those pagings should be based on that.

Should it be done in this PR?

>3) Yes, as discussing in hipchat

Probably needs to be discussed more, but I just got an idea. If "untrusted" (remote) handler will be used along with for example mongo, and if sending part of this remote handler is designed as separate module (background job) then we can just keep track this job status, I mean if something goes wrong with async we can act with its error like we do with the rest of exceptions. And we may not consider the results of "untrusted" handlers processing as a reason not to log into next backup layer.
@tomaszdurka wdyt?

>4) Please investigate why it's "kinda-flaky"

Fixed that. The problem was related to usage of `CM_Http_Request_Abstract::getInstance()`.

fvovan:
Mail and payment pagings are done.
I divided payment pagings to 'Successes' and 'Failures'.
So for now the actual question is what to do with async handlers.

Also I'm a bit lost with how migrations get from **cm** to **sk**. Added indexes to cm part but it seems changes are not applied to `sk`. @tomaszdurka could you explain this topic please?

fauvel:
You mean `$mongo->createIndex('cm_log', ['level' => 1, 'createdAt' => 1], ['background' => true]);`? It will be applied when you upgrade the cm dependency in sk and deploy the app.

fvovan:
@fauvel I mean both migration and  new created `resources/db/mongo/collections.json` file. 
I guess they must be applied during `bin/cm app setup --reload` command but they actually aren't.

fauvel:
The `collections.json` file is only being used for unit testing, and you don't have to do anything apart from upgrading cm in order to apply it.

The migration script should be applied when you run `bin/cm db run-update 65`. Could you please verify?

fvovan:
I found its usage at `CM_MongoDb_SetupScript` and this class is loaded during sk's `app setup`The >migration script should be applied when you run bin/cm db run-update 65. Could you please verify?

Which `bin/cm db run-update` do you mean? cm's or sk's? It was okay with cm but not with sk.

fauvel:
I mean `sk/bin/cm db run-update 65`. Are you sure the new index isn't being created?

fvovan:
Ah. yes @fauvel  you are right.  I realised we can specify CM namespace for db migrations run within sk's VM.
`bin/cm db run-update 65 --namespace=CM`

fvovan:
Aaaand. It's finished. there are places for improvement but can be done later.
So there were some points besides our discussions.
Log record type and log record level are supposed to be completely orthogonal things and type is stored inside of `extra` field of context. not sure should it be moved to separate field.

Also I'm not really sure if admin UI is okay. I put a lot of `print_r` to log listing. Seemed to be the easiest way.

And the last thing is https://github.com/cargomedia/cm/pull/2081/files#diff-e4c45fda91e603ebfb7ece2abbe0d388R158
don't know how if it's better to delete it, but testing can be more complex due some assertion errors can be skipped.

@tomaszdurka please review.

should be reviewed together with https://github.com/cargomedia/sk/pull/2883/

njam:
@tomaszdurka please review.

library/CM/Http/Response/JsError.php
```diff
@@ -13,11 +14,8 @@ protected function _process() {
 
         $suppressLogging = $request->isBotCrawler() || !$request->isSupported();
         if (!$suppressLogging) {
-            $text = $message . PHP_EOL;
-            $text .= '## ' . $fileUrl . '(' . $fileLine . ')' . PHP_EOL;
-
-            $log = new CM_Paging_Log_JsError();
-            $log->add($text, array('url' => $url, 'errorCounter' => $counter));
+            $exception = new CM_Exception_Javascript($message, $url, $counter, $fileUrl, $fileLine);
+            CM_Bootloader::getInstance()->getExceptionHandler()->logException($exception, CM_Log_Logger::WARNING);
```
tomaszdurka: What about using logger here? I would remove `logException` from exception handler all together and leave `handleException` only.

library/CM/Log/Handler/MongoDb.php
```diff
@@ -80,21 +79,38 @@ protected function _formatRecord(CM_Log_Record $record) {
...
                 'uri'     => $request->getUri(),
-                'query'   => $request->getQuery(),
+                'query'   => [],
                 'server'  => $request->getServer(),
                 'headers' => $request->getHeaders(),
             ];
 
+            try {
+                $formattedContext['httpRequest']['query'] = $request->getQuery();
+            } catch (CM_Exception_Invalid $e) {
```
tomaszdurka: Can u check if there is query?

library/CM/Http/Response/JsError.php
fvovan: Basically handler's `logException()` just directly proxies to logger's `logException()`. They are almost aliases. But can be changed to logger's only calls.

library/CM/Log/Factory.php
```diff
@@ -5,24 +5,46 @@ class CM_Log_Factory implements CM_Service_ManagerAwareInterface {
...
+        $handlersLayerList = [];
+        foreach ($handlersLayerConfigList as $handlersLayerConfig) {
+            $currentLayer = [];
+            foreach ($handlersLayerConfig as $handlerName) {
+                $currentLayer[] = $this->getServiceManager()->get($handlerName);
+            }
+            $handlersLayerList[] = $currentLayer;
+        }
+        if (true === $addStderrHandler) {
+            $handlersLayerList[sizeof($handlersLayerList) - 1][] = (new CM_Log_Handler_Factory())->createStderrHandler('{levelname}: {message}');
```
tomaszdurka: Use `Functional\last`?

tomaszdurka: Or rather something like:
```php
$handlersLayerList[] = [(new CM_Log_Handler_Factory())->createStderrHandler('{levelname}: {message}')];
```

?

tomaszdurka: Oh wait did I got it right? You trying to add handler to last handler layer?
So maybe:
```php
Functional\last($handlerLayerList)[] = (new CM_Log_Handler_Factory())->createStderrHandler('{levelname}: {message}');
```

library/CM/Log/Logger.php
```diff
@@ -19,119 +19,171 @@ class CM_Log_Logger {
...
+
+        $numberOfHandlers = sizeof($handlerList);
+        for ($i = 0, $n = $numberOfHandlers; $i < $n; $i++) {
+            /** @var CM_Log_Handler_HandlerInterface $handler */
+            $handler = $handlerList[$i];
             try {
                 $handler->handleRecord($record);
+            } catch (PHPUnit_Framework_ExpectationFailedException $e) {
+                throw $e; //TODO without this line mocked method assert exceptions will be treated as handler exceptions
             } catch (Exception $e) {
```
tomaszdurka: Oh noes!

tomaszdurka:
@fvovan please take a look we can have a call about it too, thanks!

fvovan:
Thanks, will fix some things before.
Actually i didn't invent more robust solution for  `PHPUnit_Framework_ExpectationFailedException`
probably should be even more general `PHPUnit_Framework_Exception`. In case we catch all the things in global handler we could make some settings for bootLoader, but here..

library/CM/Log/Handler/MongoDb.php
fvovan: Query always exists. The problem is sometimes (in `CM_Http_Request_Post`) additional checks are performed and for certain circumstances it throws.

library/CM/Log/Factory.php
fvovan: >You trying to add handler to last handler layer?

Exactly, though I think we need to discuss the reasonability of doing this. It's some kind of last frontier.

>`Functional\last($handlerLayerList)[] = ...`

It will not work. Arrays are passed and returned by value but not by reference.

fvovan:
@tomaszdurka , `logException()` has been removed.
let's discuss tomorrow remaining two topics ( if we need a last fallback handler via  `StderrHandler` and how to figure out PHPUnit exception handling )

fvovan:
PHPunit exceptions mess has been deleted.
The downside is now failure messages are not so convenient (diff and sometimes proper stacktrace are lost). but it's a tradeoff and for complex debuging it could be temporary put back
@tomaszdurka wdyt?

njam:
@alexispeter as discussed please test and have a look at the new logger in CM!

fvovan:
@njam just to make things clear.
record `type` is actually a type of corresponding `LogPaging` and it's the part of record's context. so can't be defined for every log record

library/CM/ExceptionHandling/Handler/Abstract.php
```diff
@@ -122,17 +90,17 @@ public function setPrintSeverityMin($severity) {
...
     }
 
     /**
-     * @return int|null
+     * @param Exception $exception
+     * @param int|null  $logLevel
      */
-    private function _getPrintSeverityMin() {
-        return $this->_printSeverityMin;
+    protected function _logException(Exception $exception, $logLevel = null) {
```
tomaszdurka: Try to remove `$logLevel` from signature.

tomaszdurka:
After our discussion and my further investigation:

### Grouping exceptions into logs

I see pre-this-pull-request we had divided logs into different pagings:
- https://github.com/fvovan/CM/blob/master/library/CM/Http/Response/Abstract.php#L251
- https://github.com/fvovan/CM/blob/master/resources/config/default.php#L49

Currently we only have configuration option `logLevel` and we divide it into those levels only.
- https://github.com/fvovan/CM/blob/cm-log-2/library/CM/Http/Response/Abstract.php#L272
- https://github.com/fvovan/CM/blob/cm-log-2/resources/config/default.php#L47

I think we are missing grouping now.
Let's rather use the same config as before, but simply pass type via Context:
```php
if ($catchException && isset($errorOptions['log'])) {
    $context = new CM_Log_Context(null, null, null, ['log' => $errorOptions['log']]);
    $this->getServiceManager()->getLogger()->addException($exception, $context);
}
```

### Treating some exceptions as non-errors
Beforehand we have been logging exceptions into specific pagings. This division was one-dimensional for Errors, Warnings, NotFounds. Using new approach we have multiple dimensions - level, log-type, etc..

Questions are:
- Shall we convert those exceptions into non-error-level entries (no exception related)? This would mean we don't allow exceptions as non-error entries.
- Shall we keep them as exceptions rely on filtering them out? This mean they would appear as errors...
- Or shall we keep them as exceptions but log with different level. This means we need to `addException` with specific `level` - as you suggested. I don't like the signature so much you have now:

```php
public function addMessage($message, $level, CM_Log_Context $context = null);
public function addException(Exception $exception, CM_Log_Context $context = null, $logLevel = null);
```

I would rather streamline that to:
```php
public function addMessage($message, $level, CM_Log_Context $context = null);
public function addException(Exception $exception, $level = null, CM_Log_Context $context = null);
```

Or rather, or even:
```php
public function addMessage($message, $level, CM_Log_Context $context = null);
public function addException(Exception $exception, CM_Log_Context $context = null);
public function addRecord(LogRecord $record);
```

@fvovan @njam wdyt?

fvovan:
>I think we are missing grouping now.
Let's rather use the same config as before, but simply pass type via Context:

Grouping can be done using `type` (which is basically type of Paging) . Binding's to paging type advantage over binding to exception types that is it can be defined for plain log messages as well.

>Shall we convert those exceptions into non-error-level entries (no exception related)? This would mean we don't allow exceptions as non-error entries. 

I think it entirely depends on which details related to incident we want to store. If we convert all (or almost) exceptions to plain log entries, we will lose at least trace info. Can be not so useful for using with sentry https://docs.getsentry.com/hosted/learn/rollups/ . In some cases it uses stack traces for grouping. We can pass additional info along with plain log record as well but I'm afraid it will not be so convenient. But I think some of exceptions like `PageNotFound` can be converted to plain messages as they are quite straightforward and we're most likely not interesting at its detailed info.

>I don't like the signature so much you have now:

Me actually too, but as you know unlike `addMessage()` last parameter  for `addException` almost always  will be omitted.
But agree, would be better to swap them.

njam:
Notes from talk with Tomasz:

change:
```
    $config->CM_Http_Response_Page->exceptionsToCatch = array(
        'CM_Exception_Nonexistent'  => ['errorPage' => 'CM_Page_Error_NotFound', 'log' => 'CM_Paging_Log_NotFound'],
```
to:
```
    $config->CM_Http_Response_Page->exceptionsToCatch = array(
        'CM_Exception_Nonexistent'  => ['errorPage' => 'CM_Page_Error_NotFound', 'log' => true, 'level' => WARN],
```

fvovan:
As discussed yesterday, I also renamed `level` to `minLevel` attribute for logger handlers

alexispeter:
Here's my feedback from testing the logging-system. 

* General tabs for the different levels of logging should have the option of showing all entries of this level, regardless of type. If not, you might run the risk of adding an exception-type combination, that is not accessible via interface? Then again, we are usually careful enough, right? :D
* The graphs don't contain log-entries with specific types. E.g. payment errors, which are critical, are not shown in the critical-graph and logged mails are not shown in the info-graph.
* Query-params for payment-logs are displayed twice, once in Extras and once as part of the Request.
* The payment-error-log doesn't show the exception-message and the trace is contained in the meta-infos and rendered differently than regular errors. There is also the issue that rethrowing the exception will cause it to also appear in the regular error log. It might be a bit tricky, but could we enrich the exception with directives on how to log it and let the regular exception handler log it instead of manually logging it?
* Debug-messages are not logged because all handlers have a minimum lvl of 200. I suppose this should be overwritten in local dev-config?
* CM_Component_Loglist template tries to access user-information directly on `$context` instead of `$contex.user` which causes an exception when a user is included in the context.
* The aforementioned exception looks quite weird. [Screenshot](https://cloud.githubusercontent.com/assets/1921425/13989942/b3b1111c-f112-11e5-8493-5f3fd6df0995.png)
Looking at the mongodb-entry, it seems to put the html-code in the message.
Investigating a bit further I discovered that it logs 4 exception. 1 warning and 3 fatals. The fatals look fine but they are all identical. The html code is contained in the warning.
You should be able to reproduce by adding an entry to one of the logs with a user set in the context and then accessing this log.
* Global context does not contain the viewer.

fvovan:
Alexis, thanks for the testing

>General tabs for the different levels of logging should have the option of showing all entries of this level, regardless of type. If not, you might run the risk of adding an exception-type combination, that is not accessible via interface? Then again, we are usually careful enough, right? :D

And wdyt about this `logLevel` and `type` division conception in general?
Actually I thought about implementing of showing all the levels of specified type? With the option to show specific level, or may be to select minimum displayed level for particular type? What will be more useful?

>The graphs don't contain log-entries with specific types. E.g. payment errors, which are critical, are not shown in the critical-graph and logged mails are not shown in the info-graph.

What would be better? to include them to log levels division, or to create separate graphs for them? Or to put it to both graphs? I think these two topics should be consistent with each other and probably for stats it would be better to store all the types for particular log level and all the levels for certain type. But on listing page for types we'll create  a logLevel selector. wdyt?

>Query-params for payment-logs are displayed twice, once in Extras and once as part of the Request.

That's true. But `extras` may contain post request's body. I just didn't decide where to put this `body` extraction at the logger handers side. looks like it would be better to include this code just to all of handlers.

>The payment-error-log doesn't show the exception-message and the trace is contained in the meta-infos and rendered differently than regular errors. There is also the issue that rethrowing the exception will cause it to also appear in the regular error log. It might be a bit tricky, but could we enrich the exception with directives on how to log it and let the regular exception handler log it instead of manually logging it?

The reason is the same as previous one. I will fix it, and they will be logged as just the rest of exceptions.

>Debug-messages are not logged because all handlers have a minimum lvl of 200. I suppose this should be overwritten in local dev-config?

It depends, we can enable debug level for mongo handler, or to use completely another logger for debugging purposes (file based) and even delete `debug` tab in admin to not pollute the space on the screen. I basically have been using one for various background stuff. 

>CM_Component_Loglist template tries to access user-information directly on $context instead of $contex.user which causes an exception when a user is included in the context.
The aforementioned exception looks quite weird. Screenshot Looking at the mongodb-entry, it seems to put the html-code in the message. Investigating a bit further I discovered that it logs 4 exception. 1 warning and 3 fatals. The fatals look fine but they are all identical. The html code is contained in the warning. You should be able to reproduce by adding an entry to one of the logs with a user set in the context and then accessing this log.

Thanks. I will find out what's wrong there.

>Global context does not contain the viewer.

I'm not sure how to obtain it at the moment of logger creation. I remember somebody told me that it's not right to get it from `CM_Http_Request_Abstract` (and that's why we have separate attribute but not rely on request only). But probably it will be okay as a fallback option.

fvovan:
It's an update related to 
>The aforementioned exception looks quite weird. Screenshot Looking at the mongodb-entry, it seems to put the html-code in the message. Investigating a bit further I discovered that it logs 4 exception. 1 warning and 3 fatals. The fatals look fine but they are all identical...

The problem of duplicated errors is a result of several ajax requests to server. I'm not completely aware how it works but checked dev console. the page sends 3 requests to server. Each of them throws and the error was logged, besides JS generates  `CM_Exception_Javascript` , which goes to `warning` section (though I'm not sure if it's the right place for js errors). This behaviour is identical to `master` branch. 
Second thing is an exception entry shows message twice (and 1 of them was not properly escaped). I removed it.

alexispeter:
>> General tabs for the different levels of logging should have the option of showing all entries of this level, regardless of type. If not, you might run the risk of adding an exception-type combination, that is not accessible via interface? Then again, we are usually careful enough, right? :D

> And wdyt about this logLevel and type division conception in general?
Actually I thought about implementing of showing all the levels of specified type? With the option to show specific level, or may be to select minimum displayed level for particular type? What will be more useful?

The concept is solid. 
I agree, dividing the logs by type and providing filtering-options on the level would be more useful than the other way around. In addition I would add a tab that shows all exceptions (just exceptions, not normal "message"-entries), regardless of type. 
Filtering-options could encompass filtering a specific level, filtering everything but a specific level and greater than or smaller than a specific level. Options should be passed to the url, so we can easily refresh the page with the current filter.

>> The graphs don't contain log-entries with specific types. E.g. payment errors, which are critical, are not shown in the critical-graph and logged mails are not shown in the info-graph.

>  What would be better? to include them to log levels division, or to create separate graphs for them? Or to put it to both graphs? I think these two topics should be consistent with each other and probably for stats it would be better to store all the types for particular log level and all the levels for certain type. But on listing page for types we'll create a logLevel selector. wdyt?

I'm not sure a general approach for each type and level is appropriate. We should concentrate on which types/level combinations are important. For example, there should definitely be a graph that shows only FATAL errors and one that shows any payment-errors etc, but a graph that only shows normal errors or warnings is probably not needed. A dynamic graph that shows all types of errors and has options to show specific levels or types could be cool though.

>> Debug-messages are not logged because all handlers have a minimum lvl of 200. I suppose this should be overwritten in local dev-config?

> It depends, we can enable debug level for mongo handler, or to use completely another logger for debugging purposes (file based) and even delete debug tab in admin to not pollute the space on the screen. I basically have been using one for various background stuff.

Yes of course, I just meant that it might make sense to enable it by default for dev-environment, so that people that don't know the configuration don't wonder where their debug-messages disappeared to. Me for example, before I looked into logger-internals ;)

>> Global context does not contain the viewer.

> I'm not sure how to obtain it at the moment of logger creation. I remember somebody told me that it's not right to get it from CM_Http_Request_Abstract (and that's why we have separate attribute but not rely on request only). But probably it will be okay as a fallback option.

I'm not sure. I agree that it shouldn't be taken from the request if the request is not available, which means we would have to set up the logger before the request becomes unavailable. Either inside CM_Http_Request itself or inside CM_Http_Response.
Another possibility would be to not use the global context to access the viewer, but to add functionality to CM_Exception to set the context on the exception itself. We could add the relevant context in the catch-block inside `CM_Http_Response_Abstract::_runWithCatching()`.

fvovan:
As discussed yesterday with @tomaszdurka , we decided not to overcomplicate listing options for now, but rather to include required minimum of usability. 
So for now all types of records will be displayed on certain level page (and they will be included to stats graph as well), and  typed records listing will contain all available levels. (I changed `CM_Paging_Log` signatures to allow passing of specific list of log levels).
But. to prevent the cleanup of sensitive data (which typed logs could be), i had to keep the exclusion of typed records for cleanup and flushing. That's why for example flushing CRITICAL level will not give any results (because the only usage of this level is logging payment errors).

njam:
Can we get rid of accessing the request singleton?
Slightly related we discussed once to remove the global exception handler in favor of local ones in `CM_Http_Handler` and `CM_Cli_CommandManager`.

As discussed with Tomasz: An idea would be to make the Logger accessible on ServiceManager, and modify its Context while we move along (e.g. add request, user etc).

fvovan:
>Can we get rid of accessing the request singleton?
Slightly related we discussed once to remove the global exception handler in favor of local ones in CM_Http_Handler and CM_Cli_CommandManager.
As discussed with Tomasz: An idea would be to make the Logger accessible on ServiceManager, and modify its Context while we move along (e.g. add request, user etc).

Logger is already is accessible on ServiceManager via `getLogger()` method. What if we keep global exception handler but only move logger context setup to  `CM_Http_Handler` and `CM_Cli_CommandManager` catch  blocks and then rethrow to use global exception handler?
I don't like the idea of complete removing global exception handler, because some of exceptions can occur outside of mentioned classes. I mean outside of their `try` enclosed statements.

njam:
> I don't like the idea of complete removing global exception handler, because some of exceptions can occur outside of mentioned classes. I mean outside of their try enclosed statements.

Agree, let keep both, but global handler only as a fail-safe, that should hopefully never be reached? (level: FATAL?).

library/CM/Log/Context/App.php
```diff
@@ -58,13 +77,59 @@ public function getExtra() {
...
+            throw new CM_Exception_Invalid('Exception is not set');
+        }
+        return $this->_exception;
+    }
+
+    /**
+     * @return CM_ExceptionHandling_SerializableException
+     * @throws CM_Exception_Invalid
+     */
+    public function getSerializableException() {
```
tomaszdurka: What about creating `new CM_ExceptionHandling_SerializableException` lazily?

tomaszdurka: You also check `hasException` here, so makes sense. Would simplify things by not holding this `serializableException`.

library/CM/Log/Handler/Layered.php
```diff
@@ -84,22 +86,25 @@ protected function _addRecordToLayer(CM_Log_Record $record, $layerIdx) {
...
-        if ($record instanceof CM_Log_Record_Exception && $record->getException() instanceof CM_Log_HandlingException) {
+    protected function _logHandlersExceptions(CM_Log_Record $record, array $exceptionList) {
+        $context = $record->getContext();
+        $appContext = $context->getAppContext();
+        if ($appContext->hasException() && $appContext->getException() instanceof CM_Log_HandlingException) {
             return;
         }
         foreach ($exceptionList as $exception) {
-            $this->_addRecordToLayer(new CM_Log_Record_Exception(CM_Log_Logger::ERROR, $context, $exception), 0);
+            $appContext->setException($exception);
```
tomaszdurka: Not sure if reusing the same context is fine. Especially that in the end it will contain the last exception and while it should not when going out of `_logHandlersExceptions`. What about cloning for each?

tomaszdurka:
@fvovan format Metadata better (rename to Context?)

image

fvovan:
>what is the number instead of trace?

It's timestamp. Traces were not displayed. I've fixed it.

>wdyt about presenting this as pretty-print JSON?

the whole record or some unstructured parts printed currently with `print_r()`?

library/CM/Log/Handler/Layered.php
```diff
@@ -88,9 +88,11 @@ protected function _addRecordToLayer(CM_Log_Record $record, $layerIdx, array $ex
                 $nextLayerIdx = $layerIdx + 1;
                 if ($this->_hasLayer($nextLayerIdx)) {
                     $this->_addRecordToLayer($record, $nextLayerIdx);
+                    $this->_addExceptionListToLayer($exceptionList, $record->getContext(), $nextLayerIdx, $failingHandlers);
```
tomaszdurka: I guess we don't need (should not) pass those failing handlers to next layer. It doesn't concern it.

fvovan: I would keep it for now to finally stop fixing the tests again and again and again.
can be changed after release with other adjustments

tomaszdurka:
lgtm

njam:
Great job!
Looking forward to the fluentd adapter.

PR #2082, opened 2016-02-03
Add Janus removeStreams endpoint

tomaszdurka:


tomaszdurka:
@fvovan please review 

@alexispeter @njam wdyt about removing `final` from `CM_Model_Abstract::delete()` method? :panda_face:

library/CM/Janus/RpcEndpoints.php
```diff
@@ -178,6 +178,27 @@ public static function rpc_removeStream($serverKey, $channelKey, $streamKey) {
     }
 
     /**
+     * @param string $serverKey
+     * @return bool
+     * @throws CM_Exception_AuthFailed
+     */
+    public static function rpc_removeAllStreams($serverKey) {
```
fvovan: Why it's called `rpc_removeAllStreams ` when we actually remove stream channels?

library/CM/Model/Abstract.php
```diff
@@ -101,7 +101,7 @@ public function commit($useReplace = null) {
         $this->_autoCommit = true;
     }
 
-    final public function delete() {
```
fvovan: was it done by accident or to allow mocking?

njam:
Agree to removing `final`!

library/CM/Janus/RpcEndpoints.php
tomaszdurka: StreamChannel is concept which is not really known to cm-janus. Cm-janus only publishes, subscribes (creates stream) or removes stream.

library/CM/Model/Abstract.php
tomaszdurka: Yeah. This is due to Mocka (nor any other mocking framework) not able to extend final methods.

I have seen ways using php runkit - but it's quite dirty as it modifies original behaviour.

tomaszdurka:
@fvovan please confirm.

fvovan:
lgtm

PR #2083, opened 2016-02-03
Enable autocomplete param for text form fields

christopheschwyzer:


christopheschwyzer:

njam:
Wow, those available values: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input

lgtm

christopheschwyzer:
indeed

PR #2084, opened 2016-02-04
Add short url flag to linkUrl

christopheschwyzer:


christopheschwyzer:
Maybe it should be called differently. Couldn't think of an accurate name. _stripProtocol_ or sth.

christopheschwyzer:
Maybe just `short`

njam:
- Shot sgtm
- I would do regexp pattern `'#^https?://(www\.)?#'`
 - Use `#` for limiting, so we don't need to escape `/`
 - Use `^` to match beginning of string
 - Match `s?` instead of `.?` (dot is any character)
 - Escape `\.` after www, because `.` is any character

christopheschwyzer:
Thanks. Makes sense! @njam please verify

njam:
lgtm

PR #2087, opened 2016-02-16
Optionally override CM config with `override.php`

zazabe:


zazabe:
cancelled, better solution with cargomedia/sk#2901

PR #2088, opened 2016-02-23
Allow configuration of multiple janus services

fvovan:
As a preparation for scaling the janus infrastructure we'd like to separate the *audio* and the *video* streams to separate janus servers.

My initial thought is that we could configure two independent janus services in the service manager (`janus-audio` and `janus-video`). When connecting to a stream we would then read the configuration from the right service.

@fvovan what do you think, does it make sense?
We also need to make sure that synchronization with multiple januses works on the streamchannel level. Could you check this as well?

cc @tomaszdurka

### DIFF
--- a/library/CM/Janus/Configuration.php
+++ b/library/CM/Janus/Configuration.php
@@ -43,6 +43,20 @@ class CM_Janus_Configuration {
     }
 
     /**
+     * @param string $plugin
+     * @return CM_Janus_Server|null
+     */
+    public function findServerByPlugin($plugin) {
+        $plugin = (string) $plugin;
+        foreach ($this->_servers as $server) {
+            if (in_array($plugin, $server->getPluginList())) {
+                return $server;
+            }
+        }
+        return null;
+    }
+
+    /**
      * @param int $id
      * @return CM_Janus_Server
      * @throws CM_Exception_Invalid
--- a/library/CM/Janus/Configuration.php
+++ b/library/CM/Janus/Configuration.php
@@ -48,11 +48,11 @@ class CM_Janus_Configuration {
      */
     public function findServerByPlugin($plugin) {
         $plugin = (string) $plugin;
-        $pluginServerList = Functional\filter($this->_servers, function (CM_Janus_Server $server) use ($plugin) {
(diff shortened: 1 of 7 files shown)


fvovan:
@tomaszdurka please review.
it's `$service->getConfiguration()->findServerByPlugin('<plugin-name>');`

tomaszdurka:
Ok, so you allow only one server per plugin. This basically kills our build-in scaling for stream-channels in CM.
Let's rather allow registering same plugin for multiple servers and bring simple (random) balancer for `findServerByPlugin`. It will also simplify the code (no fancy intersection).

library/CM/Janus/Server.php
```diff
@@ -32,6 +36,9 @@ public function __construct($serverId, $key, $httpAddress, $webSocketAddress, ar
         $this->_key = (string) $key;
         $this->_httpAddress = (string) $httpAddress;
         $this->_webSocketAddress = (string) $webSocketAddress;
+        $this->_pluginList = array_map(function ($el) {
```
tomaszdurka: Use `$plugin` instead of `$el`.

fvovan:
please review. I user `mt_rand` instead of `array_rand` to define random seeds in tests.

library/CM/Janus/Configuration.php
```diff
@@ -43,6 +43,21 @@ public function findServerByKey($key) {
...
+     * @param string $plugin
+     * @return CM_Janus_Server|null
+     */
+    public function findServerByPlugin($plugin) {
+        $plugin = (string) $plugin;
+        $pluginServerList = array_values(Functional\filter($this->_servers, function (CM_Janus_Server $server) use ($plugin) {
+            return in_array($plugin, $server->getPluginList());
+        }));
+        if (!empty($pluginServerList)) {
+            return $pluginServerList[mt_rand(0, count($pluginServerList) - 1)];
```
tomaszdurka: Just a note I think we generally tend to throw/return-null early, meaning here it would fit more to return null on empty server-list and then return randomized. Rather point-of-view than anything else.

tomaszdurka:
lgtm

PR #2091, opened 2016-02-25
Throw if S3 upload fails

fauvel:


fauvel:

tomaszdurka:
If that's the only way to check then looks good.

- Can it still "fail" after copying the same amount of bytes? :panda_face: 
- What are `@` usages for closing the streams? Should we really ignore errors there?

njam:
Let's be as strict as possible, so we should definitely check the return values of `fclose()` for example (and throw if there was a failure).

fauvel:
Agree, please review!

fauvel:
@ :panda_face: let's investigate which kind of issues we actually have first...

njam:
Let's move this to a separate function! Otherwise lgtm.

fauvel:
Agree, last review please!

njam:
lgtm

PR #2096, opened 2016-03-08
Add enable/disable feature to toggleModal

zazabe:


zazabe:
@njam please review

njam:
So it will disable the fully toggling (enabling and disabling). Can you show me the usecase?

zazabe:
on cargomedia/sk#3019, to lock the panel. 

it can't be done in `ToggleModal` callback, because the state could be changed by `ModalClose` and the next "close" command will not execute the callback, so something like this will not work:

```
$el.toggleModal('open', function(state) {
  if(isEnabled) {
    $el.toggleClass('panel--active', state);
  }
});
```

njam:
Ok makes sense, lgtm.

PR #2098, opened 2016-03-11
Add user agent getter

fauvel:


fauvel:
@dlondero please review

dlondero:
@fauvel adding a test?

fauvel:
Sure, please review!

dlondero:
@fauvel lgtm

njam:
What's the reasoning to return empty string if the header is missing - shouldn't this return `null`?

fauvel:
Loosing information to hide complexity – a missing and a empty header should probably be handled the same in our application code, I think.

PR #2102, opened 2016-03-18
rebind 'killChidren' on process respawn

fvovan:


fvovan:
@njam could you take a look on that please?
I tested it for our particular case. Can it harm something else? I will try to create unit tests for it.

njam:
Maybe add a test to/similar like `testKillChildrenOnExit`?

How about just moving the `bind('exit'` from `fork()` into `_fork()`?

fvovan:
>Maybe add a test to/similar like testKillChildrenOnExit?

probably for this circumstance higher level test would be more useful. We most likely want to be sure that respawned child will be killed along with its parent.

But here we can test if `killChild()` is added as a callback.
please review

njam:
lgtm

PR #2101, opened 2016-03-18
Allow "target" param in link, button_link smarty functions?

christopheschwyzer:
- to be able to use `target='_blank' - maybe we can have an "external" flag that would also add an icon to the link?!
- only if `href` is used I guess

@njam wdyt?


christopheschwyzer:
@njam please review

njam:
lgtm

PR #2104, opened 2016-03-22
Don't make TypeDumper unloadable

njam:
@zazabe please review


njam:
@zazabe please review

njam:
@zazabe please review

zazabe:
lgtm

I can't fully test it because `http://lorempixel.com` is down (`This site is temporarily down for maintenance.`) which produce an exception during `cm/bin app setup --reload`, but I dropped the db before and the problem with the initialization is fixed.

PR #2108, opened 2016-03-24
Add less-easing

christopheschwyzer:
https://github.com/dmnsgn/less-easing
https://github.com/emagnier/sass-less-stylus.easing

Small collection of variables for css transitions. 
- Check with http://easings.net/#
- Check off-canvas menu transition


christopheschwyzer:
I've added the file to the `css` folder, but technically it's just variables. Not sure if we should import them into the `variables.less` or move the file somewhere else. 
@njam please review

njam:
We should make sure that these are loaded *before* any LESS code. This is why we have `variables.less`, it gets loaded in the very beginning.

I would add it to `variables.less` in the end? Maybe add a comment about where it comes from?

christopheschwyzer:
I guess we can do it like suggested, as this will not change anytime soon. At some point we might consider splitting this file up for readability. Still ok for now. @njam please verify

njam:
agree, lgtm

PR #2109, opened 2016-03-25
Add time since page was rendered to cm.debug.log output

njam:
@zazabe please review

### DIFF
--- a/library/CM/App.js
+++ b/library/CM/App.js
@@ -309,7 +309,9 @@ var CM_App = CM_Class_Abstract.extend({
         return;
       }
       var messages = _.toArray(arguments);
-      messages.unshift('[CM]');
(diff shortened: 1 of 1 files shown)


njam:
retest this please

njam:
@zazabe please review

njam:
@zazabe updated, please review

zazabe:
Maybe we could use `performance.now()` vs `new Date()` (http://caniuse.com/#feat=high-resolution-time) but I don't think it will make a big difference.

njam:
I would stick with `new Date()` because of Android 4.3 support. Also I think the differences will not be relevant for our use cases. wdyt?

zazabe:
yes, lgtm, it was more a thought...

PR #2110, opened 2016-03-29
Add JS user agent parser library

zazabe:
Check different solutions:
- https://github.com/ua-parser   
 wildly used in different languages, but unfortunately JS support seems not be easily available: https://github.com/ua-parser/uap-ref-impl/issues/8... 
- https://github.com/faisalman/ua-parser  
  much more simple to integrate but there's quite a lot of open issues, and not so much activity...
- https://github.com/3rd-Eden/useragent  
  JS implementation based on the `ua-parser-core` regexp database, but for node only 

Finally, this PR implement a solution based on https://github.com/3rd-Eden/useragent, build through this fork https://github.com/zazabe/useragent/tree/browserfy


zazabe:
@njam what do you think about this solution?

cc @vogdb

vogdb:
> - https://github.com/faisalman/ua-parser.

The link does not work. This https://github.com/faisalman/ua-parser-js should be correct. Also I wouldn't say that it does not have so much activity. The latest commit is on the 25 of February.

zazabe:
@vogdb yes... not sure if using `useragent` is the best option but I like the idea to base the parsing on https://github.com/ua-parser/uap-core, which is used by more languages...

njam:
Agree on using https://github.com/3rd-Eden/useragent (for having "ua-parser" data).

Could we use browserify dynamically (like for vr/)?

zazabe:
I thought we will not have to change `useragent` source code, but yes, let me add it to our `after-body-source`, so we will not have to maintain a repo on my github account :D

zazabe:
@vogdb please review

- dynamically browserify `useragent`
- the parsed user agent is exposed as a `cm.userAgent` property   
  
  ```js
  cm.userAgent
  > {
  	"family": "Chrome",
  	"major": "49",
  	"minor": "0",
  	"patch": "2623",
  	"device": {
  		"family": "Other"
  	},
  	"os": {
  		"family": "Mac OS X",
  		"major": "10",
  		"minor": "10",
  		"patch": "5"
  	}
  }
  ```

vogdb:
In `README.md` - `window.UserAgent` but `main.js` - `UserAgentParser`. The rest is lgtm

PR #2114, opened 2016-04-01
Fix button_link isHtmlLabel

christopheschwyzer:


christopheschwyzer:
…alternatively we could hide the attribute altogether. @njam please review

njam:
Let's remove the value attribute.

christopheschwyzer:
@njam please verify

njam:
lgtm

PR #2116, opened 2016-04-05
Configure bluebird Promise before loading anything else

zazabe:
When a bluebird promise has already been created, the bluebird promise configuration fails:

It seems to happen only with Firefox:
screen shot 2016-04-05 at 18 10 21

Note that promise cancellation is not usable at all after that:
screen shot 2016-04-05 at 18 12 43


zazabe:
@vogdb please review

vogdb:
Have you checked that
```js
Promise.config({
  cancellation: true
});
```
is not overwritten by the late call
```js
var promiseConfig = {};
//....
Promise.config(promiseConfig);
```
?

zazabe:
Yes https://github.com/petkaantonov/bluebird/blob/v3.1.5/src/debuggability.js#L164

vogdb:
lgtm them

PR #2118, opened 2016-04-11
Introduce CM_App->getName()

zazabe:
Should we introduce the concept we have in SK's config in to CM?
```php
$config->installationName = 'FB';
```

Pass it to the client with `cm.options`?


zazabe:
@fauvel please review

fauvel:
lgtm!

PR #2119, opened 2016-04-11
Store user agent upon authentication

fauvel:


fauvel:
@alexispeter please review

alexispeter:
lgtm, though SK should probably be adjusted so we don't store it twice.

fauvel:
Will be done in https://github.com/cargomedia/sk/pull/3104

alexispeter:
alright then

PR #2123, opened 2016-04-13
[Experimental] Modules with Browserify support

zazabe:
*done during the 13 April 2016 research day*

I've achieved my goal but I'm not satisfied, there's not enough benefit and too many changes with this current implementation...

Pros:
- Clean global scope
- `sourcemaps` supported for module JS files (in dev)    
  
screen shot 2016-04-13 at 18 13 11

Cons:
- generating the `library.js` must be optimized (could take >10s!)
- all JS files need to be CommonJS compliant
- no way to load external JS module (`require()` from `source-after-body` for example) currently

My last thought: loading external JS module would make it more useful and could worth the effort to support `browserify` (or maybe an alternative module system?)...

### DIFF
--- a/library/CM/App.js
+++ b/library/CM/App.js
@@ -1,3 +1,5 @@
+var CM_Class_Abstract = require('CM/Class/Abstract');
+
 /**
  * @class CM_App
  * @extends CM_Class_Abstract
@@ -1048,7 +1050,8 @@ var CM_App = CM_Class_Abstract.extend({
      */
     _getAdapter: function() {
       if (!this._adapter) {
-        this._adapter = new window[cm.options.stream.adapter](cm.options.stream.options);
+        var Adapter = require(cm.options.stream.adapter);
+        this._adapter = new Adapter(cm.options.stream.options);
       }
       return this._adapter;
     },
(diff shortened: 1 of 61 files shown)


zazabe:
At the end, I think supporting a module system for all our javascript files is important, but using `browserify` seems to be more a hack, on the other hand, using another system will require a deep refactoring of the CM framework, change how files are organized and how they are provided to the client...

Also, if it's just about the `sourcemaps`, `uglifyjs` could do the job...

zazabe:
update: I've tried to use https://github.com/koala-framework/sourcemaps to generate a sourcemaps in dev for `library.js`, but at the end, I don't think it's a good idea. 
Sourcemaps is about making minified code readable, we can think about always uglyfing our `library.js` code and make it debuggable in dev and prod the same way. 
The `sourcemaps` could be provided by a new `CM_Http_Response_Resource_SourceMaps` that would check if the user is allowed to access it... 

Anyhow, I think I can close this PR, a next research could be about transpiled ES6 code, which may bring much more benefits than just the sourcemaps :)

@njam @vogdb any thoughts?

vogdb:
On ES6 code? Well, the only thought is that it is an infinite task :)

PR #2122, opened 2016-04-14
Add CM_Janus_StreamChannelInterface

fvovan:
For https://github.com/cargomedia/sk/pull/3215

### DIFF
--- a/library/CM/Janus/Configuration.php
+++ b/library/CM/Janus/Configuration.php
@@ -1,6 +1,6 @@
 <?php
 
-class CM_Janus_Configuration {
+class CM_Janus_Configuration extends CM_Class_Abstract {
 
     /** @var CM_Janus_Server[] */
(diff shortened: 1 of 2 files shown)


fvovan:
@njam please reivew

njam:
Shouldn't this be `findServerByChannelDefinition`, and fail if we don't find it? Then we don't need to check in the usage?

fvovan:
>Shouldn't this be `findServerByChannelDefinition`

May be, I just wanted to keep signatures consistent.
i.e. `findServerByPlugin` may return `null` , so `findServerByChannelDefinition()` behaves the same.

njam:
True, but there's also `getServerRandom()`. In this case I think failing is more useful?

The convention is that `get` should always return something (fail otherwise), and `find` can return `null`. So let's call it `get`.

otherwise lgtm

PR #2117, opened 2016-04-21
Add fluentd logger adapter

fvovan:
Use https://github.com/fluent/fluent-logger-php ?

Schema: https://github.com/cargomedia/guides/tree/master/logging

Make App name available: https://github.com/cargomedia/cm/issues/2118

@fvovan @fauvel @alexispeter please pick up when you have some time


fvovan:
it depends on https://github.com/cargomedia/cm/pull/2081

changes can be seen here https://github.com/fvovan/CM/compare/cm-log-2...fvovan:issue-2117

finally record has the following format 
```
2016-04-21T10:45:11+00:00	tag	{"message":"CM_Exception_Invalid: message","level":"error","timestamp":"2016-04-21T10:45:11+0000","computerInfo":{"fqdn":"ex.com","phpVersion":"PHP8"},"request":{"uri":"/path/path2?param=11&param2=22","method":"GET","referrer":"http://ooo.com","user_agent":"IE10 Apple like FF"},"SKApp":{"user":3,"clientId":15,"extra1":1,"extra22":"2","ff":"3333"},"exception":{"type":"CM_Exception_Invalid","message":"aaaaaaaaaaaa","stack":[{"code":"{main}","line":0,"file":"scripts/test.php"}]}}
```

is it okay that `exception` stack is represented as an array but not as a string here?

njam:
We need to ensure that `exception.stack` has the same type on all our applications.
@vogdb I guess you're passing a string there?

vogdb:
> @vogdb I guess you're passing a string there?

yep

fvovan:
paired with @tomaszdurka

PR #2128, opened 2016-04-21
Caching of query: cm_model_languagekey

alexispeter:
The following query is part of the translation module. When starting an application up an application (or releasing a new version), this query piles up. Since it's quite heavy this becomes a problem quickly.

> SELECT k.name AS `key`, v.value, k.variables FROM `cm_model_languagekey` as `k` LEFT JOIN `cm_langua[…]

@fvovan @alexispeter could you have a look?
Can we make the caching smarter (prevent many cuncurrent lookups)? Or can we warm up this cache in `fillCaches()`?

### DIFF
--- a/library/CM/App.php
+++ b/library/CM/App.php
@@ -47,16 +47,25 @@ class CM_App implements CM_Service_ManagerAwareInterface {
             $assetList[] = new CM_Asset_Css_Vendor($render);
             $assetList[] = new CM_Asset_Css_Library($render);
         }
+        $this->fillLanguageCache();
         $languageList = new CM_Paging_Language_Enabled();
         /** @var CM_Model_Language $language */
         foreach ($languageList as $language) {
-            $language->getTranslations()->getItemsRaw();
             $assetList[] = new CM_Asset_Javascript_Translations($language);
         }
         foreach ($assetList as $asset) {
             $asset->get(!$debug);
         }
         CM_Bootloader::getInstance()->getModules();
+    }
+    
+    public function fillLanguageCache() {
+        $languageList = new CM_Paging_Language_Enabled();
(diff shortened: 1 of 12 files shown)


alexispeter:
please review @njam
I added a cli-command bin/cm app fill-language-cache, which should be added to the deployment-script after the updates are run. In case any updates modified translations, this call will refill the file-cache.

njam:
Couldn't we just run the normal `fill-caches` again? I understand if caches are already present this should take almost no time, right?

@fvovan could you review?

alexispeter:
I don't think so. Many of the assets prepared in CM_App::fillCache() contain heavy operations that do not use caching. E.g. the javascript-assets cache, the result of the minification, but not the code that is collected by scanning the filesystem.
Using fill-caches after updates have run takes about 12 seconds, using fill-language-cache takes about 60 ms.

fvovan:
Is it a case that `CM_PagingSource_Abstract` can be used with two cache types (local and persistent) at the same time?

alexispeter:
No.

fvovan:
lgtm

alexispeter:
Had a brief talk with @njam concerning app deployment. We'll try resetting the file-cache eagerly instead of lazily. This will prevent situations where none of the caches is present in the case of language-keys created lazily by first-time access, as well as rendering the explicit refilling of the cache, after updates have been run, unnecessary.

alexispeter:
please review @fvovan

alexispeter:
The language-table might still get a lot of reads when a new language-key is "created" by multiple simultaneous requests. The number of affected requests will be smaller though, so it should be less of an issue than before.

fvovan:
lgtm

PR #2129, opened 2016-04-22
Add CSS function for inline image

fvovan:
New LESS function to be added in `CM_Asset_Css::_compileLess()`

Will be used like this:
```less
background-image: image_inline('index/version2/bg.jpg');
```
With an optional "size" argument (so the image would be resized to max. 10x10):
```less
background-image: image_inline('index/version2/bg.jpg', 10);
```

Should be converted to:
```less
background-image: url(data:image/jpeg;base64,/9j/4QAYRXhpZgAASUkqAAgAAAAAAAAAAAAAAP/sABFEdWNreQABAAQAAAAyAAD/7gAhQWRvYmUAZMAAAAABAwAQAwMGCQAAAcMAAAH0AAACJv/bAIQACAYGBgYGCAYGCAwIBwgMDgoICAoOEA0NDg0NEBEMDg0NDgwRDxITFBMSDxgYGhoYGCMiIiIjJycnJycnJycnJwEJCAgJCgkLCQkLDgsNCw4RDg4ODhETDQ0ODQ0TGBEPDw8PERgWFxQUFBcWGhoYGBoaISEgISEnJycnJycnJycn/8IAEQgACgAKAwEiAAIRAQMRAf/EAJcAAQEAAAAAAAAAAAAAAAAAAAUGAQEBAAAAAAAAAAAAAAAAAAAEBhAAAQMFAQAAAAAAAAAAAAAAABECAwETMwQkBREAAQEIAwAAAAAAAAAAAAAAAAERITFBUYECMmFCAxIAAgEFAAAAAAAAAAAAAAAAADEBQVGRAjITAAEDAwUAAAAAAAAAAAAAAAEAYYEQEUEhcZGhsf/aAAwDAQACEQMRAAAAtEppR8t//9oACAECAAEFAeo//9oACAEDAAEFAU8dP//aAAgBAQABBQGZrY9yzUdnP//aAAgBAgIGPwGiP//aAAgBAwIGPwF782h5P//aAAgBAQEGPwFMclcyVSHNj32ud9T/2gAIAQIDAT8QV//aAAgBAwMBPxB35+Zf/9oACAEBAwE/ECJCIaYFa4MJjwhuu9fLUv/Z);
```

Maybe additionally to the CSS compilation cache, we should cache this conversion with a file cache, so it doesn't slow down development too much (when the CSS changes, but not the images).

---
Some example code:
```php
$file = new CM_File('/home/vagrant/sk/library/VR/layout/default/resource/img/index/version2/bg.jpg');
$img = new CM_Image_Image($file->read());
$img->resize(10, 10);
$img->setFormat(CM_Image_Image::FORMAT_GIF);
$url = 'data:image/gif;base64,' . base64_encode($img->getBlob());
```

@christopheschwyzer wdyt?

### DIFF
--- a/library/CM/Asset/Css.php
+++ b/library/CM/Asset/Css.php
@@ -122,18 +122,16 @@ class CM_Asset_Css extends CM_Asset_Abstract {
                 $size = (int) $values[1][1];
             } else {
                 $path = $values[0];
-                $size = null;
+                $size = 0;
             }
 
             $imagePath = $render->getLayoutPath('resource/img/' . $path, null, null, true, true);
-            $cacheKey = CM_CacheConst::App_Resource . '_md5:' . md5($imagePath);
-            if (!empty($size)) {
-                $cacheKey .= '_size:' . $size;
-            }
-            $imageBase64 = CM_Cache_Persistent::getInstance()->get($cacheKey, function () use ($imagePath, $size) {
+            $cache = CM_Cache_Persistent::getInstance();
+
+            $imageBase64 = $cache->get($cache->key(__METHOD__, md5($imagePath), '_size:' . $size), function () use ($imagePath, $size) {
                 $file = new CM_File($imagePath);
                 $img = new CM_Image_Image($file->read());
-                if (!empty($size)) {
+                if ($size > 0) {
                     $img->resize($size, $size);
                 }
                 $img->setFormat(CM_Image_Image::FORMAT_GIF);
--- a/library/CM/Asset/Css.php
+++ b/library/CM/Asset/Css.php
@@ -124,10 +124,8 @@ class CM_Asset_Css extends CM_Asset_Abstract {
                 $path = $values[0];
                 $size = 0;
             }
-
             $imagePath = $render->getLayoutPath('resource/img/' . $path, null, null, true, true);
             $cache = CM_Cache_Persistent::getInstance();
-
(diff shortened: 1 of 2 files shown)


fvovan:
@njam thanks!
I saw this function but supposed it works only with templates.
please review.
I remember you wrote a snippet for easy and fast cache get-set, but can't find it. Is it in CM?

njam:
You can pass a getter closure to `Cache::get()`, see https://github.com/cargomedia/cm/blob/master/library/CM/Cache/Abstract.php#L50

fvovan:
thanks. it turns out i searched at `CM_Cache_Storage_Abstract`
btw. why sometimes (at `CM_Asset_Css`) we use storage class directly?

njam:
- Let's use `layout/default/resource/img/logo.png` for the specs, it's probably gonna be there longer
 - Since this is quite a large image, maybe just compare the beginning of the inline data?
- Cache key should include $size
 - You could use `$cache::key()` to generate a key. Maybe also good to assign $cache to a variable.

fvovan:
But the key contained `$size` , if it was defined https://github.com/cargomedia/cm/pull/2129/commits/65190699421f68904a99648d4026fc5541ef2e88#diff-26fdebd5605d77ab4ce404e6c20eede5L130
okay, now it always presents there.

njam:
lgtm

PR #2131, opened 2016-04-25
Page navigation doesn't inject "pagePlaceholder" anymore

vogdb:
How it should work:
1. When you start navigating to a different page we replace the current page with the "placeholder" (empty).
2. After a short timeout we inject a spinner into the placeholder
3. Once loaded we replace the placeholder with the actual content

Looks to me like step 1) doesn't work anymore.
I think this was lost in 2e60c6fd7ca9b631e0c466de8772b3c9c70a82d4.

cc @christopheschwyzer


vogdb:
@zazabe please review

zazabe:
lgtm!

PR #2134, opened 2016-04-27
Page navigation doesn't inject "pagePlaceholder" anymore II

vogdb:
Added: https://github.com/cargomedia/cm/pull/2131
Reverted: https://github.com/cargomedia/cm/pull/2133

Looks like the spinner timeout didn't clear. Spinner would show up *after* successful page load.

cc @zazabe

### DIFF
--- a/client-vendor/after-body/promise-utils/promise-throttler.js
+++ b/client-vendor/after-body/promise-utils/promise-throttler.js
@@ -42,11 +42,15 @@
     var promise;
 
     return function() {
-      var cancelLeading = options.cancelLeading && promise && promise.isPending() && promise.isCancellable();
+      if (!promise || !promise.isPending()) {
+        promise = fn.apply(this, arguments);
+        return promise;
+      }
+      var cancelLeading = options.cancelLeading;
       if (cancelLeading) {
         promise.cancel();
       }
-      var addToQueue = options.queue && promise && promise.isPending();
+      var addToQueue = options.queue;
       if (cancelLeading || addToQueue) {
         var args = arguments;
         var self = this;
@@ -57,9 +61,6 @@
           });
         });
         return promise;
-      }
-      if (!promise || !promise.isPending()) {
-        promise = fn.apply(this, arguments);
       }
       return promise;
     };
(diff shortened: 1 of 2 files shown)


vogdb:
I made a PR here but just to demonstrate the problem. If you try to run those changes, you will get the next output:
```
preparing a new request
starting a new request
preparing a new request
starting a new request
finally cancels the request
finally cancels the request
```
So you can see that we have a two `finally` in a row.

zazabe:
I can imagine it's because `cancel()` is async, the `promiseThrottler` should use `finally` on the cancelled promise to execute the new one I suppose...

Reproduced differently:
```js
var getPromise = function() {return new Promise(function(r) {setTimeout(function() {r();}, 1000)});};
var throttler = promiseThrottler(function() { 
  var now = Date.now();
  console.log('do something...'); 
  return getPromise().finally(function() {
    console.log('stopped', Date.now() - now);
  }); 
}, {cancelLeading: true});
throttler();
setTimeout(function() {throttler(); }, 500);
> do something...
> do something...
> stopped 502
> stopped 1001
```

vogdb:
@zazabe thank you so much for the hint. Please review.

zazabe:
In general, I think `nonameThrottler` could be simplified, but maybe i'm missing some cases, please enlighten me if I'm wrong :)

- why not executing `fn` if there's no promise in the first place,  like `if(!promise) { promise = fn(this, arguments); } else {//do cancel/queue stuff}`?
- maybe `isCancellable` could be removed, to get some hints if `cancelLeading` is not supported
- even not sure if it's necessary to check if the leading promise is pending or not, `finally` is executed also if the promise is not pending, isn't it?
- is it possible to drop the `queue` option? I mean, it could be the default behaviour when `cancelLeading: false`

vogdb:
Agree with everything except

> - is it possible to drop the `queue` option? I mean, it could be the default behaviour when `cancelLeading: false` 

Because of http://underscorejs.org/#throttle

vogdb:
Forgot to say, @zazabe please review

zazabe:
lgtm!

- 👍 I didn't think about the "cancelLeading" case
- 🎁 I added QUnit to CM, with a test for the `promiseThrottler`, https://github.com/cargomedia/cm/pull/2140  ;)

PR #2140, opened 2016-04-29
Introduce QUnit js test framework

zazabe:
Introduce JS unit test with QUnit on CM

usage descibed in https://github.com/zazabe/cm/blob/js-unit-test/tests/client/README.md


zazabe:
@vogdb please review

Note: the `promiseThrottler` test reproduce the #2134 issue

vogdb:
lgtm

vogdb:
well, one note here

vogdb:
In `QUnit.test("cancel leading"` should be:
```js
      assert.deepEqual(check, ['call:foo', 'finally:foo', 'call:bar', 'resolve:bar', 'finally:bar']);
```
and not
```js
      assert.deepEqual(check, ['call:foo', 'finally:foo', 'call:foo', 'resolve:bar', 'finally:bar']);
```

zazabe:
@vogdb please review

💝 > https://travis-ci.org/cargomedia/cm/jobs/127206070#L1321 ;)

zazabe:
- not sure if it's a good idea to have `bin/qunit`, could we have some troubles with composer?
- require https://github.com/cargomedia/puppet-packages/issues/1297 to run tests inside the box

vogdb:
ok. lgtm

zazabe:
OK, works in the box now:
```
vagrant@www:~/cm$ ./bin/qunit

> promise-utils/promise-throttler
cancel trailing: ✔
cancel leading: ✔
cancel queue: ✔

Took 458ms to run 3 tests. 3 passed, 0 failed.
```

@vogdb @tomaszdurka what do you think about having `bin/qunit`? is it a problem for composer?

vogdb:
I can't say anything about `bin/qunit` + composer :(

njam:
Afaik it's ok - composer honours additional files in its "bin" dir (doesn't modify/delete them).

PR #2136, opened 2016-05-02
Mechanism to transform/encode PHP objects to Javascript objects

vogdb:
Are those changes breaking btw?


njam:
I like the approach!
- There are definitely more usages of such data on the frontend I guess?
 - There are probably also more usages on the server where we should encode it like this?
- Maybe it would be useful to *not* use `toArray()` in `encodeJavascript()`, but something new/additional, so that we can define the data that should be available.
 - Then maybe longterm this can replace `toArray()` completely (as we have `toArrayIdOnly()` to serialize/unserialize objects)

@zazabe wdyt?

vogdb:
> - There are definitely more usages of such data on the frontend I guess?
 - There are probably also more usages on the server where we should encode it like this?

Yes to both questions but we should start from small.
> - Maybe it would be useful to *not* use `toArray()` in `encodeJavascript()`, but something new/additional, so that we can define the data that should be available.

please look at the latest commit.

@zazabe please review.

zazabe:
About the code:
- docu for `cm.factory.*`, `encodeJavascript()`, `


In general, I think it could be a bit more generic and transparent if the `data` property is removed, some propositions:
 - server side:  
   - Remove `encodeJavascript` and inject the `__javascript_encodable__` flag when the interface is implemented or use a trait instead, and override the `toArrayIdOnly()` to add the flag directly.     
   - remove the `setObjectProperty`, because `setObject` is already encoding data in JSON
 - client side
   - `cm.factroy.create` should pass the full data object vs `data['data']`
   - use `cm.factory.create` to decode `cm.ajax` response

By this way, any data sent to the client will be encoded (ajax response, ws pushed data) transparently, and I don't think it will break other CM parts... 

About `toJavascript`, I don't see so much the diff with `toArray`, if the `data` property is removed...

But maybe, in the future, we could add a way to decide which fields have to be expose on the fly, depending on the context, something like this:
```php
CM_Params::encode($model) == [
    '_type'                    => 100,
    '_id'                      => ['id' => 1],
    '_class'                   => 'Model',
    '__javascript_encodable__' => true,
    'id'                       => 1,
    <toArray()>...
]
CM_Params::encode($model->expose(['foo', 'bar'])) == [
    '_type'                    => 100,
    '_id'                      => ['id' => 1],
    '_class'                   => 'Model',
    '__javascript_encodable__' => true,
    'id'                       => 1,
    'foo'                      => 1,
    'bar'                      => [...]
]
```

njam:
Another thought: We're using `CM_JavascriptEncodable` to determine if a class has a JS representation, and that we should instantiate it. We could also determine that from the fact that there is a JS file with the right name, then we wouldn't need to use the interface, wdyt?

zazabe:
Or we can even remove the `__javascript_encodable__` flag and just instantiate the JS object if it exists in `cm.model.types`, in this case, there's nothing to do on the server side i suppose...

njam:
👍

zazabe:
maybe `cm.factory.create` should handle Model embedded in Model recursively, or maybe we could have a backbone `CM_Model_Abstract` that will call `cm.factory.create` on `set`, wdyt?

vogdb:
@njam 
> Another thought: We're using `CM_JavascriptEncodable` to determine if a class has a JS representation, and that we should instantiate it. We could also determine that from the fact that there is a JS file with the right name, then we wouldn't need to use the interface, wdyt?

I see that this is quite a heavy operation. Are you sure to use it?
```php
$ref = new ReflectionClass('CM_Model_User');
$ref->getFileName();
```

vogdb:
Or maybe we have access to `cm.model.types` on back-end? So I can access it from `CM_Params`?

njam:
> to determine if a class has a JS representation, and that we should instantiate it.

I thought about keeping a list of files in APC cache. Read it once and then just check it.

But Michel is right, that we can just do the check on the client side, right?

vogdb:
> But Michel is right, that we can just do the check on the client side, right?

Yep. just can't start testing those cause of my problem with Linked CM. Hope to resolve it soon.

zazabe:
the code looks good in general, few notes:

- `cm.factory.*` docu
- should we introduce a `CM_Model_Abstract`? :
  - add support of Models in Model   
    `cm.factory.create()` called for each attribute (override `set` method?)
  - serialize required data only    
    we could have a large amount of data stored in a client side model but those data will not be used by the server side currently, so i guess it will be safer to just serialize required metadata (`_type`, `_id`?) by default.
  - maybe some other stuff in the future, like WS data binding :)

zazabe:
looks good, now i'm wondering if we should not compute the `_type` + `_id` as the `CM_Model_Abstract` `idAttribute` by default...

zazabe:
- docu (convention)
  - `Boolean` vs `boolean`
  - `@returns` vs `@return`
- not sure, but it seems that a `CM_Model_Abstract` on the server side [needs the `_id` and `_type`](https://github.com/cargomedia/cm/blob/master/library/CM/Model/Abstract.php#L705) attributes, but `toJSON()` returns `id` and `_type` only...

njam:
> needs the _id and _type

I actually wrote a comment about this on Github, but then didn't submit it :) There are two serialization functions:
- toArrayIdOnly: Gives `_type` (`int`) and `_id` (array of `int|string` values)
- toArray: The above plus optionally a `id` (`int|string`)

The use case of `toArrayIdOnly()` is to just serialize and unserialize models. The use case of `toArray` is additionally to provide data of the model for using it on the client side.

Most of the time we should probably use `toArray()` when sending objects to JS, so we have that additional information available. But maybe there will also be use cases where we *only* want serialization/unserialization.

So I think best would be to use `_type` and `_id` for `_compoundId`, and expose `id` optionally if it's available. wdyt?

zazabe:
From what I understand, `fromArray` (`self::factoryGeneric($data['_type'], $data['_id'])`) needs the data exposed by `toArrayIdOnly` (`return array('_type' => $this->getType(), '_id' => $this->getIdRaw());`) to recreate an instance of the model, so I suppose `toJSON` should expose the same data, isn't it?

njam:
Ah yes, definitely!

njam:
- Model/Abstract.js: toJSON: `this.get('_id').id`: The `id` key is the identifier for *most* models, but not for all. For example a location has `id` and `level`. I think here we should use all of the object's keys and values to come up with the compound key.
- Model/Abstract.js: toJSON: I don't think we should include `id` here, as it's not needed to identify a model, but is rather an additional field that's mainly used on the client side (like other model-specific fields, e.g. `user`, `privacy` etc).

zazabe:
lgtm, just 2 small docus: `@return` in `CM_App` and `CM_Model_Abstract.toJSON()` + properties

zazabe:
lgtm

PR #2142, opened 2016-05-04
QUnit: Load suites synchronously to ensure suite.config support

zazabe:


zazabe:
Some issues with #2140 fixed here...

@vogdb please review

vogdb:
lgtm

PR #2144, opened 2016-05-04
Add paging log type and level validation to component

fvovan:
to prevent 500 errors on `/log?level={$someRandomNumber}`

### DIFF
--- a/library/CM/Paging/Log.php
+++ b/library/CM/Paging/Log.php
@@ -30,17 +30,8 @@ class CM_Paging_Log extends CM_Paging_Abstract implements CM_Typed {
             }
         }
 
-        if (null !== $filterType) {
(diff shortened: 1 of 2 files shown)


fvovan:
@tomaszdurka please review

library/CM/Paging/Log.php
```diff
@@ -147,4 +138,17 @@ protected function _getCriteria() {
 
         return $criteria;
     }
+
+    /**
+     * @param int $filterType
+     * @return bool
+     */
+    public static function hasType($filterType) {
```
tomaszdurka: Maybe `isValidType`?
Also rename `$filterType` to `$type`.

tomaszdurka:
Rest lgtm.

PR #2145, opened 2016-05-04
Improve QUnit dependency management

zazabe:


zazabe:
@vogdb sorry for all these reviews, but the previous version was not so flexible for app depending on it, should be fine now... please review

vogdb:
lgtm

PR #2149, opened 2016-05-05
Introduce CM Media lib

zazabe:
- introduce some basic objects: `Event`, `Observer`
- introduce `Media` and `Audio` /  `Video` 
- expose new objects in `cm.lib`
- replace `mejs.MediaElement` usage in `CM_App.createAudio` by `Media.Audio`   
  so we rely only on the `<audio>` HTML5 tag, which should be fine: http://caniuse.com/#search=audio%20element
- add more Media unit tests   
  mainly used to test the Media support on different browser/devices, skipped by `bin/qunit`
- experimental support of the Web Audio API


zazabe:
@vogdb please review

cc @njam

vogdb:
lgtm. Just curious. Can we replace `cm.setupVideo` by `Media.Video` ?

zazabe:
We could probably... but it could break things so I will leave it for another PR ;)

PR #2152, opened 2016-05-06
Check if the HTTP server is up before running QUnit tests

zazabe:
`bin/qunit` may fail with this error:
```
Unable to access network: fail
phantomjs://code/phantomjs-runner.js:109
```

### DIFF
--- a/bin/qunit
+++ b/bin/qunit
@@ -1,30 +1,15 @@
-#!/usr/bin/env bash
+#!/bin/bash -e
 
 SOURCE="${BASH_SOURCE[0]}"
-while [ -h "$SOURCE" ]; do # resolve $SOURCE until the file is no longer a symlink
+while [ -h "$SOURCE" ]; do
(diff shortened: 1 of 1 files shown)


zazabe:
@njam please review

njam:
Using a bash `trap` could be nice here, to ensure the php server is killed when the script exits:
```sh
trap "kill ${SERVER_PID}" EXIT
```
See `build-ci.sh` for an example.

Another nice pattern to implement a timeout is using the `timeout` command, e.g.:
```sh
timeout --signal=9 3 bash -c "while ! (nc -z 127.0.0.1 8080 &>/dev/null); do sleep 0.1; done"
```
If the while doesn't finish after 3 seconds it will kill it with signal "9", and the whole `timeout` command will exit with a status other than 0.

zazabe:
nice tips, thanks!

@njam please review

njam:
lgtm

PR #2137, opened 2016-05-06
Add CLI command to generate favicon images

fvovan:
Via https://github.com/cargomedia/cm/pull/1977

Goal:
1. Store square image file `favicon.svg` in `<namespace>/layout/<theme>/resource/img/favicon.svg`
2. New CLI command that generates PNG images for each namespace and each theme (in `layout/default/resource/img/meta/`)
 - Move `CM_Css_Cli` to `CM_Frontend_Cli`

In the CLI command we would define a list of supported configurations:
```php
$configs = [
    'square-144.png'                       => ['height' => 144, 'width' => 144, 'transparent' => false],
    'splashscreen-400-300-transparent.png' => ['height' => 400, 'width' => 300, 'transparent' => true, 'iconSize' => 0.5],
];
```
This way we can pre-fill the cache of all these transformed images before a new deployment.
The values are:
- height: px
- width: px
- transparent: whether to make the background transparent or not
- iconSize: How big the `favicon.svg` should be placed compared to the smaller size of the output image. Default: 1

When rendering the PNG *without* the `transparent` flag, the image's background should be set to the CSS variable `@colorBrand` (to get it into PHP see `library/CM/SmartyPlugins/function.lessVariable.php`).

Let me know if you want to pick it up so we can talk it through briefly. Will require some adjustements to `CM_Image` (loading svg).


fvovan:
@njam please review
I guess we need to adjust `$config` for generated images. will it be located right in class or in one of our application configs?
I think it's good idea to add a test for `createFromSVG()` as well but I need some test file. Do we have any conventions/policies about test images?

njam:
@christopheschwyzer could you add a `layout/default/resource/img/favicon.svg` for CM?
Then we could copy that file also to `tests/data/img/`, and use it unit tests.

- CM_Image_Image: __construct: How about using only 1 argument (as it's exclusive anyway)? `string|Imagick $image`
- CM_Image_Image: __construct: readImageBlob: Exception messages says "Cannot load Imagick instance", but actually reading the image blob failed.
- CM_Image_Image: createFromSVG: $skipXmlHeader: So loading an SVG would fail if the `<?xml>` header is missing? If so, could we just automatically prepend it if it is missing?

njam:
@christopheschwyzer could you also provide the config for all image formats we need?

fvovan:
Thanks!

>CM_Image_Image: __construct: How about using only 1 argument (as it's exclusive anyway)? string|Imagick $image

I thought about it, but instead of type hinting for  `Imagick` parameter we'll have to do dynamic type checking. But agree, it's clearer  when param is exclusive.

>CM_Image_Image: __construct: readImageBlob: Exception messages says "Cannot load Imagick instance", but actually reading the image blob failed

I changed it, actually it was the same as before PR, but since we added new argument the message became ambiguent.

>CM_Image_Image: createFromSVG: $skipXmlHeader: So loading an SVG would fail if the <?xml> header is missing? If so, could we just automatically prepend it if it is missing?

agree.

Okay, I'm waiting for new test image and the rest of config

christopheschwyzer:
@fvovan I couldn't push to your repo :/ However, here are the requested changes: https://github.com/christopheschwyzer/CM/tree/favicon-config - Still need to adjust the file names…

@njam Also found this repo for the favicon sizes: https://github.com/audreyr/favicon-cheat-sheet The topic is very confusing - e.g. still not sure if we should add an `.ico` version just to be save. There are some formats we current have. Need to investigate further.

fvovan:
@njam please review.

njam:
Initial notes:
- generateFavicon: Shouldn't we use `Render::getLayoutPath()` instead of `getThemeDir()` to find the SVG? This way we also use the fallback mechanism.
- generateFavicon: Would be nice to make this function shorter, and reduce complexity. Ideas:
 - Define *default* config values, merge them with passed values in a single place
 - Extract config into separate function
 - Extract LESS variable retrieval to separate function
 - Extract code that creates one output PNG from configuration

fvovan:
> This way we also use the fallback mechanism.

what is fallback mechanism?

> Define default config values, merge them with passed values in a single place

what's the desired format of passing additional configs to cli command

njam:
> what is fallback mechanism?

I mean the fact that `getLayoutPath()` loops through all of the site's modules and themes. On the other hand `getThemeDir()` only looks in the first module and theme.

> what's the desired format of passing additional configs to cli command

I didn't mean to pass config on the command line. Just that we merge the `$config` with a `$defaults` first to create the final `$config`, and then use it.

njam:
As discussed, let's keep only looking at the top module and theme, so using `getThemeDir()`!

Instantiate CM_File instead of $svgPath string.

fvovan:
I did all the improvements.
There were two tests that used color checking to verify if image was created or composed properly.
But for some reasons when I run tests on vagrant it returned `srgb(255,0,0)` as a result for `getColorAsString()` and on travis the same call gave `rgb(255,0,0)`  (eg another colorspace).
I've changed tests to check only color values, but it seems weird  anyway.

njam:
- CM_Frontend_Cli: Als instantiate `CM_File` for `$themeDir`, then use `join()` to create the `$svgFile` instance. It's safer because you don't need to worry about directory separators, if a string already includes one or not.
- CM_Frontend_Cli: Why multiply $svgResolution by 1.3? Is it not fitting 100% for the value "1"? Might this be because the SVG has a quite large margin (see https://jsfiddle.net/hczeh6pr/)? I removed the multiplication and the result looks fine!
- CM_Frontend_Cli: It's looking in the wrong path (`vendor/cargomedia/cm/layout/default/favicon.svg` instead of `vendor/cargomedia/cm/layout/default/resource/img/favicon.svg`).
 - Same for output paths
- CM_Frontend_Cli::_getFaviconConfigList: Now there's config validation in this function, and defaults in the other function. How about doing everything here?
 - We could even have a wrapper class, maybe it's simpler -  `CM_Frontend_FaviconConfig`, and then return an array of such instances?

fvovan:
I think wrapper class for config is an overkill here because it  currently would be used only once at private method. so I just moved config merging to this helper.
I found another kinda issue thing: when I iterate over all sites using `CM_Site_Abstract::getAll()` I get alias sites, and in this case, generation of same images happens again.

njam:
> I get alias sites, and in this case, generation of same images happens again.

Oh! So either we:
- Determine unique module/theme combinations first, then only run it for those (but with a specific site)
- We just generate it again, and trust that the result will be the same

wdyt?

fvovan:
So the only relevant determinant of their uniqueness in this case is `themeDir`, right?
since `variables.less` is a part of theme as well, I think it could be filtered and run with the first iterated site. and it should give the same results if we rerun it again  (when not filtering).
But for the clearer output (text output with the results file generated) i would prefer to filter

njam:
Yes agree, so let's collect unique themeDirs (or maybe clearer faviconPaths).

fvovan:
It's better to collect themeDirs, because they can be reused when writing created files to them.
please reivew

njam:
lgtm!

@christopheschwyzer could you test it before we merge it?

christopheschwyzer:
cool - works fine!

PR #2150, opened 2016-05-10
Detect timezone offset and store it in a cookie

fvovan:
- detect timezone using  ([`Date.getTimezoneOffset`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/getTimezoneOffset))
- store offset in cookie so it's accessible on server-side

### DIFF
--- a/library/CM/App.js
+++ b/library/CM/App.js
@@ -494,7 +494,7 @@ var CM_App = CM_Class_Abstract.extend({
         wordSeparator: " ",
         numbers: []
       };
-      $.cookie('timezoneOffset', (new Date()).getTimezoneOffset());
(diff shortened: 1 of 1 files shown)


fvovan:
@tomaszdurka  please review

tomaszdurka:
Let's keep simple hour offset (instead of minutes). @fvovan agree?

fvovan:
Sometimes it's fractional values, not sure if it's good idea to store numbers like 2.5 or even 5.45 .

njam:
There are half-hour timezones btw, so it would be a float.

Usually I would suggest to use SI units, which makes it easy because it's a well-defined standard. So how about passing the offset in seconds?

tomaszdurka:
Agree to secs then.

PR #2151, opened 2016-05-10
Read timezone from offset cookie

fvovan:
- read offset from cookie
- instantiate timezone using `DateTime::createFromFormat('O', $offset)->getTimezone();`
- use this timezone in Frontend_Environemnt (precedence over location-based and bootloader timezones)

### DIFF
--- a/library/CM/Http/Request/Abstract.php
+++ b/library/CM/Http/Request/Abstract.php
@@ -432,6 +432,20 @@ abstract class CM_Http_Request_Abstract {
     }
 
     /**
+     * @return DateTimeZone|null
+     */
+    public function getTimeZone() {
+        if ($timezoneOffset = $this->getCookie('timeZoneOffset')) {
+            //timezoneOffset is minutes behind UTC
+            $timezoneOffset = (int) $timezoneOffset * -1;
+            $offsetHours = intval($timezoneOffset / 60);
+            $offsetMinutes = intval($timezoneOffset % 60);
+            return DateTime::createFromFormat('O', sprintf("%+03d%02d", $offsetHours, $offsetMinutes))->getTimezone();
+        }
+        return null;
+    }
+
+    /**
      * @return string
      */
     public function getUserAgent() {
--- a/library/CM/Http/Request/Abstract.php
+++ b/library/CM/Http/Request/Abstract.php
@@ -436,10 +436,10 @@ abstract class CM_Http_Request_Abstract {
      */
     public function getTimeZone() {
         if ($timeZoneOffset = $this->getCookie('timezoneOffset')) {
-            //timezoneOffset is minutes behind UTC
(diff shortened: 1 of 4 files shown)


fvovan:
@tomaszdurka please review. Seems it's the only place where TZ can be injected into new `CM_Frontend_Environment` 
there are some other usages in `CM_RenderAdapter_Layout::fetch()` like 
```
$environmentDefault = new CM_Frontend_Environment($this->getRender()->getEnvironment()->getSite());
$renderDefault = new CM_Frontend_Render($environmentDefault);
```
but i guess it's for  some sort of special rendering.

fvovan:
@tomaszdurka  if you began please suspend it, because it will be changed in https://github.com/cargomedia/cm/issues/2153

tomaszdurka:
I've actually did review this, but didn't put comment. Please put those two topics aside on separate pull-requests.
One note I wanted to add is regarding the offset calculation. I would propose to make offset cookie already in hours. So it's much easier on server-side. Passing number of minutes is strange.

fvovan:
@tomaszdurka please review

library/CM/Http/Request/Abstract.php
```diff
@@ -432,6 +432,20 @@ public function getLocation() {
     }
 
     /**
+     * @return DateTimeZone|null
+     */
+    public function getTimeZone() {
```
tomaszdurka: Extract `Location::getTimeZone()` into `_getTimeZoneFromOffsetCookie`?

fvovan:
@tomaszdurka  please review

library/CM/Http/Request/Abstract.php
```diff
@@ -509,6 +516,20 @@ private function _getLanguageBrowser() {
...
     /**
+     * @return DateTimeZone|null
+     */
+    private function _getTimeZoneFromCookie() {
+        if ($timeZoneOffset = $this->getCookie('timezoneOffset')) {
+            //timezoneOffset is seconds behind UTC
+            $timeZoneOffset = (int) $timeZoneOffset * -1;
+            $offsetHours = floor($timeZoneOffset / 3600);
+            $offsetMinutes = floor($timeZoneOffset % 3600 / 60);
+            return DateTime::createFromFormat('O', sprintf("%+03d%02d", $offsetHours, $offsetMinutes))->getTimezone();
```
tomaszdurka: Small last question. Is this on purpose to not have `:` between hours and minutes? I guess it works both ways.

tomaszdurka:
rest lgtm

library/CM/Http/Request/Abstract.php
fvovan: hmm. actually it works but probably it shouldn't.
php documentation says that `O` format does not contain the colon but `P` contains it

PR #2155, opened 2016-05-10
Jobworker: clear runtime-cache before each run

alexispeter:
Runtime cache may contain outdated cache-entries from a previous job.


alexispeter:
please review @njam

njam:
lgtm

PR #2156, opened 2016-05-12
Set minLevel for NewRelic log handler to ERROR

njam:
This way we can keep using NR error alerts, which has nice features (like alert if the percentage of pageviews with errors exceeds a certain threshold).
Since there's no way to distinguish warnings from errors in NR I think this is the only way to do it.

@fvovan @tomaszdurka wdyt?


fvovan:
sgtm.
The only concerns are we may lose some warnings analytics (probably useless and could be seen at loggly now), and if we treat some errors as warnings (unlikely happens).

njam:
Yes, we should use Loggly for that.

tomaszdurka:
Ok with me.

PR #2157, opened 2016-05-12
improved phpunit-assertions for model-class

alexispeter:
Improves the display of failed equality-assertions between model-classes.

`Failed asserting that SK_User(197) matches expected SK_User(196).`

Also works for arrays, making debugging pagings etc much easier.
image

Sideeffects: Instances of CM_Comparable cannot be compared with assertEquals() anymore without requiring the instance to actually be the same.

### DIFF
--- a/tests/helpers/CMTest/library/CMTest/Comparator/BetterArrayComparator.php
+++ b/tests/helpers/CMTest/library/CMTest/Comparator/BetterArrayComparator.php
@@ -1,0 +1,8 @@
+<?php
+
+class CMTest_Comparator_BetterArrayComparator extends \SebastianBergmann\Comparator\ArrayComparator {
+    
+    public function __construct() {
+        $this->exporter = new CMTest_Exporter();
+    }
+}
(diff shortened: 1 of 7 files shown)


njam:
Couldn't we have a comparator that accepts `CM_Comparable`?

alexispeter:
We could.

alexispeter:
please review @njam

njam:
Nice!

- Comparator/Comparable.php: assertEquals: Update "model" reference
- CMTest/Exporter.php: Would it make sense to use `CM_Debug_VariableInspector::getDebugInfo()`?

alexispeter:
> CMTest/Exporter.php: Would it make sense to use `CM_Debug_VariableInspector::getDebugInfo()`?

That depends on the extent you want to use it. It could make sense to use it for objects implementing `CM_Debug_DebugInfoInterface`. But I'm not comfortable with blindly using it for anything, and making sure it doesn't break anything would be way too much effort for too little benefit.

njam:
ok, let's not use it then.

alexispeter:
please review @njam

njam:
ok, let's use it then ;)

lgtm

alexispeter:
I meant it would be good to use it for objects that use the interface, just not as a general dumper for everything ;)

PR #2153, opened 2016-05-13
Detect timezone from environment's location

fvovan:
- Loop over [`DateTimeZone::listAbbreviations()`](http://php.net/manual/en/datetimezone.listabbreviations.php) and look for closest to Environment's Location ([`DateTimeZone::getLocation`](http://php.net/manual/en/datetimezone.getlocation.php))
- Use this timezone to set frontend environment's timezone (precedence over bootloader's timezone)


fvovan:
@tomaszdurka please review.
it contains the rest of PRs but after they will be merged, here will be only current changes

library/CM/Http/Response/Abstract.php
```diff
@@ -117,17 +117,23 @@ public function getRender() {
...
         if (null === $currency && null !== $location) {
             $currency = CM_Model_Currency::findByLocation($location);
         }
-        $clientDevice = new CM_Http_ClientDevice($this->getRequest());
-        return new CM_Frontend_Environment($this->getSite(), $viewer, $this->getRequest()->getLanguage(), null, null, $location, $currency, $clientDevice);
+        $clientDevice = new CM_Http_ClientDevice($request);
+
+        $timeZone = $request->getTimeZone();
+        if (null === $timeZone && null !== $location) {
+            $timeZone = $location->getTimeZone();
```
tomaszdurka: What about putting this inside `getTimeZone` on request? Would simplify this code.
Also extract current `Location::getTimeZone()` into `_getTimeZoneFromOffsetCookie`.

fvovan: hmm. why `Location::getTimeZone()` ? it is extracted from cookies on `Request` level
Do you mean `Request::getTimeZone()`?

not sure it's good idea because theoretically `Location::getTimeZone()` can be used without request (for example in daemons when we send emails and get user's timezones). And you can easily control the precedence

tomaszdurka: Sure i meant request's one...

fvovan: I would not move `Location` related code intp `Request`

fvovan:
@tomaszdurka  please review

tomaszdurka:
Hey, generally idea looks good!

Suggestions in bullets:
- we have class called `CM_Geo_Point` - I would use as much defined structures instead of float-arrays
- use 2 `CM_Geo_Point` for `distance` method
- move distance method to `CM_Geo_Point`
- make `CM_Model_Location::getCoordinates` return `CM_Geo_Point` or add another method, we could also instantiate `CM_Geo_Point` in new codebase

fvovan:
Hmm. It's better to use `CM_Model_Location::distance` then.
`CM_Util::distance` is not related to  `Geo`, because it does not take into account the curvature of the Earth surface and in general the result can not be measured in any absolute units (like meters).

>make CM_Model_Location::getCoordinates return CM_Geo_Point

I think it could be done later in another PR, because refactoring will affect a lot of occurrences, Elastic filters, etc.
@tomaszdurka please review

tomaszdurka:
Looks good to me.

Small note: It's practice to name functions as verbs. What about changing `distance` to `calculateDistanceTo`?

fvovan:
Ah true. Thanks. I renamed it to `getDistance` to be consistent with `CM_Model_Location` existing method

tomaszdurka:
Oh I understand. Better than `distance` still probably getters should be reserved for non-argument methods. Your choice. Otherwise looks good.

fvovan:
okay, then `Location::getDistance` can be renamed during refactoring of its returning type

tomaszdurka:
lgtm

PR #2158, opened 2016-05-13
Rework Context

tomaszdurka:
- Replace back CM_Log_Context_App with CM_Log_Context
- remove CM_Log_Context_App
- add CM_Log_Context_Extra
- store everything on Context
- adjust usages


tomaszdurka:
@fvovan please review

Not sure in the end why I've put out `extra` into separate class. Probably can be reverted.

library/CM/ExceptionHandling/Handler/Abstract.php
```diff
@@ -101,6 +101,8 @@ private function _getPrintSeverityMin() {
      */
     protected function _logException(Exception $exception) {
         $logLevel = CM_Log_Logger::exceptionSeverityToLevel($exception);
-        $this->getServiceManager()->getLogger()->addMessage('Application error', $logLevel, new CM_Log_Context_App(null, null, $exception));
+        $context = new CM_Log_Context();
```
fvovan: I think new signatures became worse. imo passing ready context was much better.

library/CM/Log/Formatter/Text.php
```diff
@@ -40,8 +40,9 @@ public function renderContext(CM_Log_Record $record) {
             }
             $data['extra'] = implode(', ', $extraText);
         }
-        if ($record->getContext()->getAppContext()->hasException()) {
-            $data['exception'] = $this->_renderException($record->getContext()->getAppContext()->getSerializableException());
+        if ($exception = $record->getContext()->getException()) {
+            $serializableException = new CM_ExceptionHandling_SerializableException($exception);
```
fvovan: What if we hide `new CM_ExceptionHandling_SerializableException()` inside context's method?

library/CM/Log/Context/Extra.php
```diff
@@ -0,0 +1,45 @@
+<?php
+
```
fvovan: I think we need either to bring it back  to "just array representation" or to do something with `type`, if `extra` is a class now, `type` could be its attribute. But probably it's simpler just to revert and consider it as "flexible and easy data storage".

library/CM/ExceptionHandling/Handler/Abstract.php
tomaszdurka: I disagree. Having complicated constructor signature with a lot of nulls is much worse than simply using setter for each of them. It also makes code more readable. Also extracting something into variable helps read the code. Plus we don't have so many usages.

library/CM/Log/Formatter/Text.php
tomaszdurka: I thought about that, but this is only used in some special cases where we want to store exception. From Context point of view we don't know it and simple getException does the job.

library/CM/Log/Context/Extra.php
tomaszdurka: Not sure about importance of `type`, but I agree that Extra class is something unneeded. It also works in a bit different way than other parts of Context - you need to dump it with `toArray`. I will revert it.

tomaszdurka:
@fvovan please review

fvovan:
lgtm

tomaszdurka:
Adjusting sk in https://github.com/cargomedia/sk/issues/3360

tomaszdurka:
@fvovan please confirm

tomaszdurka:
Nothing changed - confirming on my own :)

PR #2159, opened 2016-05-17
FormField and other time-related stuff

alexispeter:
FormField, params getter and smarty helper to input and display time.
Required for https://github.com/cargomedia/sk/pull/3356
Code adapted from https://github.com/denkmal/denkmal.org.

`function.date_time.php` was modified to work with DateTime- or timeStamp-parameters.
`CM_FormField_Time` was modified to reject times like 22:60 or 24:00 as they are the same as 23:00 or 0:00 of the next day respectively.

please review @njam


njam:
@fvovan could you review?

How about we add timezone support, now that we have it in CM?

alexispeter:
Why not, wasn't sure if this was already fully integrated.

njam:
It's being built here: https://github.com/cargomedia/cm/pull/2153
The result will be that the user's timezone will be in the `CM_Frontend_Environment`. There's already a (default) timezone there today. So you could use it already now.

njam:
One more thought: a point in time is only fully defined if the *time* is combined with a *date*. E.g. if an event is added on the day before DST for the next day, it should be "offset" by 1 hour.

So how about introducing a `FormField_DateTime`, that asks for date *and* time?

alexispeter:
I don't think this is necessary in this particular case, though it might come in handy in other cases.
The idea is to use one formfield for the date and two for start- and end-time, since the interval should never be more than a few hours, the date of the seconds time can be safely assumed to either be the same date or the next.

njam:
A "DateTime" formfield would have the advantage that it would encapsulate the logic of combining a date and a time to a "datetime" correctly. See [here](https://github.com/denkmal/denkmal.org/blob/master/library/Denkmal/library/Denkmal/Form/EventAdd.php#L68) for how it's done in denkmal.

How about adding a *DateTime* field, and synchronizing the dates with JS? So if the user enters the start *date* it would automatically set the *end* date to the same value?
Otherwise we could think about a *DateTimeInterval* field, which would consist of a date and two times?
screen shot 2016-05-18 at 12 53 37
https://www.denkmal.org/add

alexispeter:
> A "DateTime" formfield would have the advantage that it would encapsulate the logic of combining a date and a time to a "datetime" correctly. See here for how it's done in denkmal.

Using a date-field and two time-fields actually **is** how it is done in denkmal.
If you absolutely insist, we can do some kind of DateTimeInterval field that combines the three. Intervals would be limited to 24 hours.

njam:
I think a "DateTimeInterval" makes sense. Let's keep it <24h for now.
(We can always add a "+1 day" button in the future - or change the UI completely. But having a logical representation for such an input value seems valid.)

library/CM/FormField/Time.php
```diff
@@ -0,0 +1,18 @@
+<?php
+
+class CM_FormField_Time extends CM_FormField_Text {
+
+    public function validate(CM_Frontend_Environment $environment, $userInput) {
+        $userInput = parent::validate($environment, $userInput);
+
+        if (!preg_match('/^(\d{1,2})(?:[:\.](\d{2}))?$/', $userInput, $matches)) {
```
fvovan: Do we need to allow '.' (dot) as a delimiter ?

library/CM/FormField/Time.php
```diff
@@ -0,0 +1,18 @@
...
+            throw new CM_Exception_FormFieldValidation(new CM_I18n_Phrase('Invalid time.'));
+        }
+        $hour = (int) $matches[1];
+        $minute = array_key_exists(2, $matches) ? $matches[2] : 0;
```
fvovan: Please add casting to int.

fvovan:
I think it will be clearer to input (and/or to store) either "DateTime of event beginning + its length (interval) " or "DateTime of beginning + DataTime of ending".
Otherwise it would be quite clumsy for events starting around midnight and will be even harder to manipulate them in storage (although storage representation is not defined) 

May be add this new field into Example component?

library/CM/FormField/Time.php
alexispeter: It doesn't hurt.

alexispeter:
please review @fvovan

fvovan:
@alexispeter please merge upstream

alexispeter:
please review @fvovan

library/CM/FormField/DateTimeInterval.js
```diff
@@ -0,0 +1,45 @@
...
+      this.trigger('change');
+    });
+  },
+
+  isEmpty: function(value) {
+    return _.isEmpty(value.day) || _.isEmpty(value.month) || _.isEmpty(value.year) || _.isEmpty(value.start) || _.isEmpty(value.end);
+  },
+
+  getInput: function() {
+    return this.$('select');
```
fvovan: Why it's `select` but not `select:first`?

library/CM/FormField/DateTimeInterval.php
```diff
@@ -0,0 +1,83 @@
...
+    }
+
+    public function validate(CM_Frontend_Environment $environment, $userInput) {
+        $dd = (int) trim($userInput['day']);
+        $mm = (int) trim($userInput['month']);
+        $yy = (int) trim($userInput['year']);
+        $start = trim($userInput['start']);
+        $end = trim($userInput['end']);
+
+        $base = new DateTime($yy . '-' . $mm . '-' . $dd, $environment->getTimeZone());
```
fvovan: What happens if is throws?
I saw form validator catches only `CM_Exception_FormFieldValidation` from its inputs.

library/CM/FormField/Time.php
fvovan: what about casting?

fvovan:
should we validate '31th February' or similar mismatches on frontend?

library/CM/FormField/DateTimeInterval.js
alexispeter: Because there are multiple select elements in this form field.

fvovan: hmm. but it also contains `<input>` elements.

library/CM/FormField/DateTimeInterval.php
alexispeter: Invalid entries in the time-fields are detected and thrown as CM_Exception_FormFieldValidation. Invalid entries in the date-fields are prevented by the limited input capabilities.

alexispeter:
> should we validate '31th February' or similar mismatches on frontend?

I wouldn't. The 31st of February is interpreted as the 2nd or 3rd of March, which is ok.

alexispeter:
please review @fvovan

fvovan:
please add a type casting in 'CM/FormField/Time.php for
  `$minute = array_key_exists(2, $matches) ? $matches[2] : 0;`
and lgtm

alexispeter:
Actually, I just forgot to remove it.

fvovan:
ah
lgtm

PR #2107, opened 2016-05-19
Remove FastClick

christopheschwyzer:
Fast-Tap is default in iOS 9.3 / Safari 9.1!
https://developer.apple.com/library/mac/releasenotes/General/WhatsNewInSafari/Articles/Safari_9_1.html

Let's keep this ticket to track adoption, and maybe finally to remove FastClick?

See also https://github.com/ftlabs/fastclick/issues/450

cc @christopheschwyzer @vogdb


christopheschwyzer:
@njam please review

njam:
lgtm

PR #2164, opened 2016-05-21
Allow site URLs to include path

njam:
This changeset allows a site's URL to contain a *pathname*.
E.g. `http://www.denkmal.org/basel` Instead of just `http://www.denkmal.org`

Notes:
- History values in the browser contain the *full* fragment (e.g. `/basel/page1?foo=bar`)
- The value sent to `ajax_loadPage()` starts where the siteUrl ends (e.g. `/page1?foo=bar`)
- Any non-page requests (like CSS, JS etc) still go to the *root* of the URL (e.g `/vendor-css/en/…`)

Breaking: Signature of `Site::match()` changed.

### DIFF
--- a/library/CM/App.js
+++ b/library/CM/App.js
@@ -182,7 +182,7 @@ var CM_App = CM_Class_Abstract.extend({
     if (!options['sameOrigin'] && cm.options.urlCdn) {
       url = cm.options.urlCdn + url;
     } else {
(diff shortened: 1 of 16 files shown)


njam:
@fvovan please have a look and/or let's discuss tomorrow

fvovan:
lgtm

PR #2166, opened 2016-05-21
Fix timezone passed to formatter, allow to override timezone

njam:


njam:
@fvovan please review

cc @tomaszdurka

fvovan:
I'm not sure about timezones named like 'GMT+4'.
According http://php.net/manual/en/timezones.others.php correct name is `Etc/GMT+4`

PR #2167, opened 2016-05-21
Allow to pass timeZone to smarty date helpers

njam:
- date
- date_time

cc @fvovan @tomaszdurka

### DIFF
--- a/layout/default/css/form.less
+++ b/layout/default/css/form.less
@@ -81,6 +81,7 @@ textarea.textinput {
 
   a&, div& {
     line-height: @sizeButton - 2*@buttonBorderWidth;
+    text-align: center;
   }
 }
 
--- a/library/CM/Frontend/Render.php
+++ b/library/CM/Frontend/Render.php
@@ -348,13 +348,21 @@ class CM_Frontend_Render extends CM_Class_Abstract implements CM_Service_Manager
     }
 
     /**
-     * @param int         $dateType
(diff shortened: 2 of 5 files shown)


njam:
Based on https://github.com/cargomedia/cm/pull/2166

@fvovan please review

fvovan:
lgtm

njam:
As discussed, I think it's correct for *Intl*Date:

> This ID must be a valid identifier on ICUʼs database or an ID representing an explicit offset, such as GMT-05:30.
http://php.net/manual/en/intldateformatter.create.php

tomaszdurka:
Why do we allow to pass a different timezone for the helper btw? Is there a use case you guys see you want to on this level change the timezone?

njam:
Yes, for *local/physical* events. Imagine an event happening in Paris at 8pm. You'd want to render it in Paris' timezone, and not in the environment/viewer's timezone!

tomaszdurka:
Yes, I thought about those Denkmal cases when seem this initially over the weekend.

I totally agree to this usage. Not sure if we should make timezone optional on `getDateFormatter` on `Render` though. Thought we should rather recreate instance in smarty helper (using locale, etc from Env). Probably overkill. Maybe if the `$timeZone` argument would be called `$timeZoneOverride` it would be bette :)

Anyway, dw, lgtm.

PR #2168, opened 2016-05-21
Call "checkAccessible()" before rendering any page templates

njam:
To prevent that we render for example "title.tpl" which might access $viewer, for whose existance we chek in checkAccessible

@fvovan please review


fvovan:
don't we need it for `fetchViewTemplate()` too?

njam:
As discussed we need to call it earlier (before prepare()).

PR #2174, opened 2016-05-23
fix  getTimezone() error

fvovan:


fvovan:
error when processing negative fractional offset
@njam, @tomaszdurka  please review

tomaszdurka:
lgtm

PR #2160, opened 2016-05-24
Check performance of mongodb queries to retrieve logs in admin

fvovan:
We noticed the queries to retrieve logs in the admin sometimes time out (30 seconds). My experience is that the page loads quite slowly most of the times.

Can you check if this is because of the mongodb query, and if so how it can be improved?

cc @tomaszdurka @alexispeter


fvovan:
@tomaszdurka please review.

alexispeter:
There is actually a very simple way of just using the criteria to count stuff, at least for non-aggregated pagings. The non-aggregated query for CM_Paging_Log uses only a criteria, a projection and sorting. If and aggregation is used, the whole pipeline has to be processed to get the count. if no aggregation is used, only the criteria is needed. Therefore CM_PagingSource_MongoDb should be instantiated with these three distinct parameters instead of cramming them into the aggregation. 
Optimising the count-query for the aggregated case on the other hand would require a mechanism to customise it.

While optimising the count-query certainly helps mitigate load on mongodb, it won't help if the data-retrieval query takes too long. So I think we should just think about what data we actually need and how best to get it.
The only log I experienced timeouts with was `all=1` and while having a paging which shows log-entries of every type is certainly a good thin, I don't think that showing entries of all time is particularly useful. Reducing them to entries of last week should be enough and not cause any timeouts.

fvovan:
>The non-aggregated query for CM_Paging_Log uses only a criteria, a projection and sorting. If and aggregation is used, the whole pipeline has to be processed to get the count. if no aggregation is used, only the criteria is needed. Therefore CM_PagingSource_MongoDb should be instantiated with these three distinct parameters instead of cramming them into the aggregation. 

In terms of mongodb all queries in CM_Paging_Log are aggregated now. We needed `$projection` for  plain list. Can be possibly redesigned, if we change output format.

>Reducing them to entries of last week should be enough and not cause any timeouts.

Actually we already do it. Records older than 1 week are erased. (except payment notifications)

alexispeter:
When talking about non-aggregated queries I was actually talking about the aggregation of errors, not the aggregation-mechanics of mongodb.

fvovan:
@alexispeter please review

alexispeter:
Aggregated Log should return the exception in `context.exception` so that the format stays the same for both use-cases.

fvovan:
Aggregated log records don't contain context at all, because it's useless to make a grouping by it as context has users and request fields and they  most likely will be different every time

alexispeter:
Then adjust the other case?

fvovan:
Actually the reason we had mongodb's aggregation (which caused performance problems and was removed in PR) is we wanted this  sort of unification. :)
I did it with `_processItem()`
please review

alexispeter:
lgtm

PR #2146, opened 2016-05-25
Handle exceptions in endpoints (cli/http) instead of using global exception handler

fvovan:


fvovan:
I found that global handler is not  actually used for CLI tasks.
Their exceptions are called and handled in various places inside `CM_Cli_CommandManager`, `CM_Process` and `CM_Process_ForkHandler` (in particular exceptions thrown from `CM_Cli_Runnable_Abstract` methods are caught in https://github.com/cargomedia/cm/blob/1b95b051ed11ac0a52a4e11e42de1b9f2f33e677/library/CM/Process/ForkHandler.php#L79

I had to remove 1 test because since exception handler is "installed" directly in `CM_Http_Handler::processRequest()` we just exit there if something goes wrong. and it seems there is no way to test if the process exist.

@tomaszdurka please reivew

tomaszdurka:
I'm bit confused with these changes. Probably need some explanation. Please ping me on chat.

njam:
@tomaszdurka @fvovan could you have a look here? This might solve https://github.com/cargomedia/cm/issues/2203

tomaszdurka:
- Keep `processRequest` as before.
- Add `CM_Bootloader->execute()`
- Let's not add optional severity, but rather:
```
/**
     * @param Exception $exception
     */
    public function handleException(Exception $exception) {
        $severity = $exception instanceof CM_Exception ? $exception->getSeverity() : null;
        $this->handleExceptionWithSeverity($exception, $severity);
    }

    public function handleExceptionWithSeverity(Exception $exception, $severity) {
        $printException = true;
        if ($exception instanceof CM_Exception) {
            $printException = $severity >= $this->_getPrintSeverityMin();
        }

        if ($printException) {
            $this->_printException($exception);
        }

        $this->_logException($exception, $severity);
        
    }
```

fvovan:
@tomaszdurka  please review

library/CM/ExceptionHandling/Handler/Abstract.php
```diff
@@ -98,9 +105,15 @@ private function _getPrintSeverityMin() {
 
     /**
      * @param Exception $exception
+     * @param int|null  $severity
+     * @throws CM_Exception_Invalid
      */
-    protected function _logException(Exception $exception) {
-        $logLevel = CM_Log_Logger::exceptionSeverityToLevel($exception);
+    protected function _logException(Exception $exception, $severity = null) {
```
tomaszdurka: Let's make severity non-optional (simplifies) - we have only one usage where we pass severity (`handleExceptionWithSeverity`).

tomaszdurka:
Rest looks good.

library/CM/ExceptionHandling/Handler/Abstract.php
```diff
@@ -108,7 +108,7 @@ private function _getPrintSeverityMin() {
      * @param int|null  $severity
      * @throws CM_Exception_Invalid
      */
-    protected function _logException(Exception $exception, $severity = null) {
+    protected function _logException(Exception $exception, $severity) {
         if (null === $severity) {
```
tomaszdurka: Remove this `if` too?

fvovan: no. it still can be `null` explicitly.
came from `handleException() -> handleExceptionWithSeverity( ..., null )  `

fvovan:
@tomaszdurka please confirm

tomaszdurka:
lgtm

PR #2176, opened 2016-05-27
fix paging log 

fvovan:


fvovan:
weird, could not reproduce the bug on local env. but I think it should help.
@njam please review

njam:
@alexispeter could you review?

alexispeter:
lgtm

PR #2177, opened 2016-05-29
Add basic service-worker support

njam:
Introducing an HTTP response and an asset to build a service-worker JavaScript file.
Currently the service-worker is not loaded automatically. This is done manually in the Denkmal project for now (in `Denkmal_App.ready()`), but could be moved to CM later on.

@zazabe please review. I recommend looking throught commit by commit.


library/CM/Asset/Javascript/Vendor/Abstract.php
```diff
@@ -2,46 +2,32 @@
...
+                $this->_appendPathGlob($this->_getDistPath());
                 break;
             case 'source':
-                $this->_appendSourceJs($this->_getSourcePath(), $generateSourceMaps);
+                $this->_appendPathBrowserify($this->_getSourcePath());
                 break;
             default:
-                $this->_appendDistJs($this->_getDistPath());
-                $this->_appendSourceJs($this->_getSourcePath(), false);
+                $this->_appendPathGlob($this->_getDistPath());
```
zazabe: The sourcemap generated by browserify will be wrong if there's some extra code in the generated content (like this "glob" here).
I understand that it will not enter this case in debug (`before-body.js` / `after-body.js` are injected only when `debug=false`) but it's less explicit than before with the `$generateSourceMap` parameter...

client-vendor/serviceworker/cm/main.js
```diff
@@ -0,0 +1,4 @@
+self.addEventListener('install', function(event) {
```
zazabe: What is `self` in this context?

zazabe:
The code looks good in general (just see my 2 comments), but I'm not sure how I can test it, I don't see any place where the service worker is registered.

njam:
Oh I wasn't aware. Fixed and added a guard. Please review.

Regarding `self` - it seems to be the ubiquitous reference to the global object in the browser (for both Window and Workers):
https://developer.mozilla.org/en-US/docs/Web/API/Window/self
https://developer.mozilla.org/en-US/docs/Web/API/WorkerGlobalScope/self

You can see how to register the service worker [here](https://github.com/denkmal/denkmal.org/pull/547/files#diff-bfcc277d1cf8b898af20e42d230d8355). I'm not registering it in CM yet, because we need to add some caching for it to be useable under high load. The service-worker needs to be loaded from the same domain as the document, so it's not using the CDN…

zazabe:
Not tested but the code looks good to me!

PR #2179, opened 2016-05-30
Configure trackings per site

fvovan:
Allow to configure `tracking` service by site.

- Add $siteName to params of CM_Service_Tracking_ClientInterface methods
which doesn't accept `$environment` for now

### DIFF
--- a/library/CM/Class/Abstract.php
+++ b/library/CM/Class/Abstract.php
@@ -22,6 +22,13 @@ abstract class CM_Class_Abstract {
     }
 
     /**
+     * @return string[] List of class names
+     */
+    public static function getClassHierarchyStatic() {
+        return self::_getClassHierarchy();
+    }
+    
+    /**
      * @param int $type
      * @return string
      * @throws CM_Class_Exception_TypeNotConfiguredException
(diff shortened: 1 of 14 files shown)


fvovan:
I have one more question.
What we are going to do in case of tracking things which doesn't belong to any site at the moment of their  tracking? For example in `CM_Action_Abstract::_track()` (so far) , should we extract 'actions trackers' to another service, which will be common for all sites of the app?
wdyt @fauvel @njam ?

fvovan:
Second case is `CM_Model_Splittest::_getVariationFixture()`. Even though it's possible to pass SiteId when we process `CM_Model_Splittest_RequestClient`, but `CM_Model_Splittest_User` does not (and most likely should not) contain any relations to `Site`.
So I propose to detach action and split test trackings to separate interfaces and services.

library/CM/Class/Abstract.php
```diff
@@ -22,6 +22,13 @@ public function getClassHierarchy() {
     }
 
     /**
+     * @return string[] List of class names
+     */
+    public static function getClassHierarchyStatic() {
+        return self::_getClassHierarchy();
```
fauvel: This doesn't make much sense imo – you could directly make `getClassHierarchy` static if needed.

fvovan: yes, but sometimes it's used in tests where it's mocked as instance method so I kept it for more compatibility.

njam:
@fvovan @fauvel closeable?

PR #2182, opened 2016-06-01
Ensure css-vendor contains css from  *-body folders only

zazabe:


zazabe:
@njam please review

njam:
lgtm

PR #2186, opened 2016-06-06
Allow disabled attribute for button_link

christopheschwyzer:


christopheschwyzer:
@tomaszdurka please review
@njam fyi

njam:
I was just discussing with Michel a way to expose possible viewSession transitions to the client. I think this "disabled" on the server side will then not be needed anymore.
See https://github.com/cargomedia/sk/issues/3344

christopheschwyzer:
Not sure I understand. I've planned to use disabled buttons in https://github.com/cargomedia/sk/pull/3325. This might be good to have anyways?

njam:
I think it will not be needed in https://github.com/cargomedia/sk/pull/3325, if we make it more dynamic (change it with JS).
In general this parameter for `{button_link` is a bit strange, because it only applies to the `<button>`-case. Therefore I think it might be best to *not* do add it?

PR #2187, opened 2016-06-07
Make CM_Params::encode() work with JsonSerializable

njam:
- Remove `ArrayConvertible::toArray()` in favor of  `JsonSerializable::jsonSerialize()`
- Rename `ArrayConvertible::toArrayIdOnly()` to `toArray()`

### DIFF
--- a/library/CM/Params.php
+++ b/library/CM/Params.php
@@ -536,34 +536,37 @@ class CM_Params extends CM_Class_Abstract implements CM_Debug_DebugInfoInterface
      * @param mixed        $value
      * @param boolean|null $json
      * @throws CM_Exception_Invalid
-     * @return string
+     * @return array|string
      */
     public static function encode($value, $json = null) {
         if (is_array($value)) {
-            $value = array_map('self::encode', $value);
-        }
-        if ($value instanceof CM_ArrayConvertible) {
+            $result = array_map('self::encode', $value);
+        } elseif ($value instanceof CM_ArrayConvertible || $value instanceof JsonSerializable) {
             $class = get_class($value);
-            $value = $value->toArray();
-            $value = array_map('self::encode', $value);
-            $value = array_merge($value, array('_class' => $class));
-        }
-        if ($value instanceof JsonSerializable) {
-            $class = get_class($value);
-            $value = $value->jsonSerialize();
-            if (is_array($value)) {
-                $value = array_map('self::encode', $value);
-                $value = array_merge($value, array('_class' => $class));
+            $result = [];
+            if ($value instanceof CM_ArrayConvertible) {
+                $array = $value->toArrayIdOnly();
+                $array = array_merge($array, array('_class' => $class));
+                $result = array_merge($result, $array);
             }
+            if ($value instanceof JsonSerializable) {
+                $array = $value->jsonSerialize();
+                if (is_array($array)) {
+                    $result = array_merge($result, array_map('self::encode', $array));
+                }
+            }
+        } else {
+            $result = $value;
         }
+
         if ($json) {
-            if (is_object($value)) {
-                $value = 'null';
+            if (is_object($result)) {
+                $result = 'null';
             } else {
-                $value = CM_Util::jsonEncode($value);
+                $result = CM_Util::jsonEncode($result);
             }
         }
-        return $value;
+        return $result;
     }
 
     /**
--- a/library/CM/Params.php
+++ b/library/CM/Params.php
@@ -543,15 +543,24 @@ class CM_Params extends CM_Class_Abstract implements CM_Debug_DebugInfoInterface
             $value = array_map('self::encode', $value);
         }
         if ($value instanceof CM_ArrayConvertible) {
-            $array = $value->toArray();
-            $array = array_map('self::encode', $array);
-            $value = array_merge($array, array('_class' => get_class($value)));
+            $class = get_class($value);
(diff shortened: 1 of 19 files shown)


alexispeter:
Blindly treating the result of jsonSerialize() as an array might be dangerous, the interface specifies a `mixed` return value.

njam:
Good point, please review!

zazabe:
I think it's fine, just a small question (maybe I missed something): 
`CM_Model_Abstract::jsonSerialize()` is not used when a `CM_Model_Abstract` is encoded with `CM_Params::encode()`, isn't it?
So is there any use case where `json_encode` is called directly on a Model?
Because it could be a bit confusing, I could imagine that somebody would override this method to define which fields have to be exposed, as discussed here: https://github.com/cargomedia/cm/pull/2136#issuecomment-217108899

njam:
Agree the situation is a bit messy. My suggestion would be to replace `toArray()` midterm by `jsonSerialize()`. Then `ArrayConvertible` would only have `toArrayIdOnly()` and `fromArray()`.
wdyt?

So maybe `JsonSerializable` should have precedence over `ArrayConvertible` in `Params::encode()`?

alexispeter:
I think we should view `ArrayConvertible` as a special case of `JsonSerializable` for objects that can be decoded, `ArrayConvertible` should have precedence.

njam:
Ok, new proposition:
- **ArrayConvertible**: Ability to serialize and deserialize with a minimal amount of data (usually class and id). Functions to implement: `toArrayIdOnly()` and `fromArray()`
- **JsonSerializable**: Ability to serialize for frontend with additional data that represents the object. Functions to implement: `jsonSerialize()`
- `Params::encode()` will consider both interfaces, and combine the data if necessary.

This will be a breaking change, but I think it's worth for the extra clarity.
I've pushed the code. wdyt?

I'm not 100% sure about the semantics of `JsonSerializable`. Maybe it should be stricter with a custom interface à la `Frontend_JsonSerializable`? Or can we assume that usually json serializing is done for frontend use cases?

zazabe:
Agree, seems more clear like that!

- there's certainly more use cases for `JsonSerializable` than only encoding for the frontend, so keeping the native interface seems good to me.
- not sure why https://github.com/cargomedia/cm/pull/2187/files#diff-9eb7b58609621eb1ecb61a8ef779df42R698 is needed, probably some backward compatibility?

njam:
Yes, the `id` field is used sometimes on the client side, so I'd say let's keep it in `jsonSerialize()`. But agree, this could be simplified later so that we only use `toArrayIdOnly()` for identification.

alexispeter:
Can we rename to`ArrayIdOnly` to `toArray` since we're removing `toArray`?

library/CM/Params.php
```diff
@@ -536,25 +536,37 @@ private function _getInt($param) {
...
+            $result = [];
+            if ($value instanceof CM_ArrayConvertible) {
+                $array = $value->toArrayIdOnly();
+                $array = array_merge($array, array('_class' => $class));
+                $result = array_merge($result, $array);
+            }
+            if ($value instanceof JsonSerializable) {
+                $array = $value->jsonSerialize();
+                if (is_array($array)) {
+                    $result = array_merge($result, array_map('self::encode', $array));
```
alexispeter: The serialised value should be added to the result even if it's not an array.

njam: Not sure what you mean. If the result is scalar, like `12` - we can't merge that with an array in a meaningful way?

njam:
@alexispeter agree, did the renaming. Please review

library/CM/Params.php
alexispeter: I expressed myself poorly, what I meant is that if you pass a `JsonSerializable` object to Params::encode() and jsonSerialize() returns a scalar value, then the function should return this value. The current implementation would return an empty array, because it is assumed with `$result = [];` that instances of `CM_ArrayConvertible` or `JsonSerializable` always return arrays.
This unfortunately makes this whole merging business a bit tricky, as you pointed out. Maybe we should create our own interface after all and restrict the return-value to array?

alexispeter:
renaming lgtm

library/CM/Params.php
njam: Yeah, both have ups and downs. Using the standard interface is *nice*. But then the contract is a *bit* different, so it's a bit quirky. But since the use case of *not* returning an array in `jsonSerialize()` is quite limited I think it might be fine. wdyt?

alexispeter:
I'm not too sure about having CM_Params automatically merge both serialisations. Usually you only need one, even though the object might implement both interfaces. E.g. serialising a user to pass it to a gearman-job only requires the ArrayConvertible-part, whereas "sending" it to a browser may only require the JsonSerialisable-part.

library/CM/Params.php
alexispeter: Ok.

njam:
Sending it to the browser also can require the "ArrayConvertible", in case we send it back as a parameter.

Maybe we should introduce `ArraySerializable` (instead of using `JsonSerializable`)…?

alexispeter:
But wouldn't any implementation of `JsonSerializable` necessarily include all of the properties contained in `ArraySerializable`? Those are the most significant properties of an object after all.

njam:
As discussed both ways have up and downsides. We will stick with `JsonSerializable` for now.

PR #2189, opened 2016-06-08
Client-Server model binding (experimental)

zazabe:
can be tested on https://admin.url/example?tab=binding


zazabe:
@njam please review

zazabe:
closable i guess...

PR #2193, opened 2016-06-10
Add getErrors in CM_Http_Response_View_Form

dlondero:
Allow to access errors in CM_Http_Response_View_Form.


dlondero:
@njam please review

njam:
Would be nice to know what types these are. I think they are strings, can you confirm? If so, let's change the docu to `string[]`.

dlondero:
@njam yes, strings. Error message and sometimes fieldName as well.

njam:
Ok, lgtm

PR #2190, opened 2016-06-13
In cm.factory.create, check the instance type instead of a cid property

vogdb:
In https://github.com/cargomedia/cm/blob/master/library/CM/App.js#L292, it's not so clear what is this `data.cid`. 
I think it would be better to explicitly check if `data` in not an instance of `Backbone.Model` or `Backbone.Collection`, wdyt @vogdb?

### DIFF
--- a/library/CM/App.js
+++ b/library/CM/App.js
@@ -290,11 +290,8 @@ var CM_App = CM_Class_Abstract.extend({
     _isCmObject: function(data) {
       var className = data && data['_class'];
       var isClass = className && cm.model.types[className] && window[className];
-      if (isClass) {
-        var isBackbone = data instanceof Backbone.Model || data instanceof Backbone.Collection;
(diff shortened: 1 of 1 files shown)


vogdb:
As always took some time to test it. @zazabe please review

zazabe:
maybe having `return isClass && !isBackbone;` will be a bit more nice...

vogdb:
Ok :( `instanceof` is a heavy operation

vogdb:
@zazabe please look

zazabe:
ok, didn't thought about the perf, but you think it could have some impact?

vogdb:
well, I just have this habit from old times. Lets yearn for readability!

vogdb:
@zazabe good to merge?

zazabe:
yes!

PR #2191, opened 2016-06-13
Make "Integer" form field inherit from "Text"

christopheschwyzer:
Currently `CM_FormField_Integer` is a separate form field using the "noUiSlider" library to show a slider to select a value. This has more downsides than advantages.

How about we:
- Make `CM_FormField_Integer` inherit from `CM_FormField_Text` (similar like `CM_FormField_Float`)
- Move the "noUiSlider" functionality over to `CM_FormField_Distance` - where it makes sense

@christopheschwyzer wdyt? wanna do the change?
cc @alexispeter


christopheschwyzer:
@njam please review
@tomaszdurka fyi 
…or the other way around.

tomaszdurka:
Generally looks good, but...
- Integer should only validate when integer, not numeric. Otherwise rename to "Number"?
- Maybe let's keep min/max on `Integer`? Use case could be integer "age" field.
- Is there any reason distance-slider-field inherits from text-field. Maybe it should inherit straight from Hidden one?

christopheschwyzer:
@tomaszdurka please review

library/CM/FormField/Location.js
```diff
@@ -24,7 +24,7 @@ var CM_FormField_Location = CM_FormField_SuggestOne.extend({
   },
 
   /**
-   * @returns {CM_FormField_Integer|Null}
+   * @returns {CM_FormField_Number|Null}
```
tomaszdurka: Shouldn't this be Distance field now?

library/CM/FormField/Number.php
```diff
@@ -14,13 +10,13 @@ public function validate(CM_Frontend_Environment $environment, $userInput) {
         if ($value < $this->_options['min'] || $value > $this->_options['max']) {
             throw new CM_Exception_FormFieldValidation(new CM_I18n_Phrase('Value not in range.'));
         }
+
         return $value;
     }
 
     protected function _initialize() {
         $this->_options['min'] = $this->_params->getInt('min', 0);
         $this->_options['max'] = $this->_params->getInt('max', 100);
```
tomaszdurka: These two above could be probably optional, but let's leave it as is.

library/CM/Form/Example.php
```diff
@@ -10,7 +10,7 @@ protected function _initialize() {
         $this->registerField(new CM_FormField_Float(['name' => 'float']));
         $this->registerField(new CM_FormField_Money(['name' => 'money']));
         $this->registerField(new CM_FormField_Url(['name' => 'url']));
-        $this->registerField(new CM_FormField_Integer(['name' => 'int', 'min' => -10, 'max' => 20, 'step' => 2]));
+        $this->registerField(new CM_FormField_Number(['name' => 'int']));
```
tomaszdurka: You could keep the min/max values.

tomaszdurka:
Rest lgtm.

christopheschwyzer:
Makes sense. @tomaszdurka please verify

tomaszdurka:
lgtm (didn't test it manually, I assume it works just fine :) )

tomaszdurka:
@christopheschwyzer just realized `CM_FormField_Number` we've created is the same as `CM_FormField_Float`.

christopheschwyzer:
Well one returns `(int) $value` the other `(float) $value` :)

tomaszdurka:
So it's should be Integer and probably validation should throw on other values.

PR #2195, opened 2016-06-14
Allow any ArrayConvertible to be stored with model schema

tomaszdurka:
How about we allow to store any instance of `CM_ArrayConvertible` in `CM_Model_Schema_Definition`, by json stringifying the array structure and storing it in the persistence layer?
We're doing something similar for models already.

@alexispeter @tomaszdurka wdyt?

### DIFF
--- a/library/CM/Model/Schema/Definition.php
+++ b/library/CM/Model/Schema/Definition.php
@@ -48,8 +48,8 @@ class CM_Model_Schema_Definition {
                             $value = $value->getTimestamp();
                             break;
                         default:
-                            if (!class_exists($type) && !interface_exists($type)) {
-                                throw new CM_Model_Exception_Validation('Field type `' . $type . '` is not a valid class/interface');
+                            if (!class_exists($type)) {
+                                throw new CM_Model_Exception_Validation('Field type `' . $type . '` is not a valid class');
                             }
                             $className = $type;
                             if (!$value instanceof $className) {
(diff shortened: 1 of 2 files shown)


tomaszdurka:
@njam want to review?

njam:
- Schema/Definition.php: this should *not* use `CM_Params::encode()`, because those are internally calling `jsonSerialize()` if the object implements `JsonSerializable` (recent change). For this use case we want the most minimalistic data that identifies the object, so the output of only `toArray()`. This can be retrieved by `CM_Params::encodeObjectId()`.
 - Maybe add a test with an object that also implements `JsonSerializable`
- DefinitionTest.php: How about testing that encode+decode are compatible - like sending the output of one to the input of the other?

njam:
I'm also wondering if we should really include `_class` in the serialized value. As we discussed it's not really necessary, because we know the *base* class from the schema, and can just call `BaseClass::fromArray($data)` which should be able to create the right instance.
Storing the `_class` in the database has the disadvantage that one needs to update lots of records if the class gets renamed. So I would rather prefer to *not* store it.
wdyt?

njam:
Three ways:
- `CM_Params::encode($object)`
- `CM_Params::encodeObjectId($object)`
- `$object->toArray()` **<--- should be sufficient here**

tomaszdurka:
@njam please review

njam:
lgtm

PR #2199, opened 2016-06-14
Polyfill the Performance API

zazabe:


zazabe:
@vogdb please review

zazabe:
cc @njam

vogdb:
@zazabe you are so impatient! only 22 minutes have passed and you are already calling for njam :D

vogdb:
lgtm

PR #2197, opened 2016-06-14
Add an index on "cm_log.message"?

fvovan:
Via https://github.com/cargomedia/sk-analytics/issues/397

There should not be so many different values for `message`. In the long run we rather want to keep the message static, and move variable parts to other parts.

@fvovan @alexispeter wdyt?
cc @eillasti

### DIFF
--- a/resources/db/mongo/collections.json
+++ b/resources/db/mongo/collections.json
@@ -26,6 +26,14 @@
       "options": {
         "name": "context_extra_type_1_level_1_createdAt_1"
       }
+    },
+    {
+      "key": {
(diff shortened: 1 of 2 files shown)


fvovan:
@alexispeter please review

alexispeter:
lgtm 

index-creation will probably take a long time and should be done manually on production, agree?

fvovan:
for dev environment it takes 18-25 secs on 1.3M records collection.
but could be different on prod due sharding and current load.
so I inserted `return`

PR #2201, opened 2016-06-14
Make slider a display option for Integer FF

tomaszdurka:


tomaszdurka:
@njam can you check?

njam:
- CM_FormField_Integer: adding a test for rendering and validating would be nice
- CM_FormField_Distance: shouldn't this set the display to "slider" by default?
- CM_FormField_Distance: Checkin min/max probably not needed because it's done in Integer?

tomaszdurka:
Closing in favor of #2210

PR #2207, opened 2016-06-15
Use SSL in "manifests/site.pp"

ppp0:
Via https://github.com/cargomedia/puppet-packages/pull/1293

@ppp0 could you adjust it? Streamline with [`denkmal's site.pp`](https://github.com/denkmal/denkmal.org/blob/master/puppet/environments/development/manifests/site.pp) for example?
cc @kris-lab


ppp0:
@njam please review

njam:
Let's add a comment what domain this cert is valid for (or store it in a file, then it's clear from the file name).

ppp0:
@njam please review

njam:
We could use just `node default` here, if there's only 1 node. But this works as well (as long as the hostname is set, which is the case at least for the virtualbox vagrant provider).

Tested and works?
lgtm.

ppp0:
Yes, tested and works for me.. But not without the empty default node. I copied this from https://github.com/cargomedia/sk/blob/master/puppet/environments/development/manifests/site.pp#L1 :)

njam:
Yes if there's a non-default node you also need the default node.
But if there's only 1 node you could use *only* the default node.
See https://github.com/denkmal/denkmal.org/blob/master/puppet/environments/development/manifests/site.pp

ppp0:
Can we please merge it? It I remove the default, then I keep getting
```
 Error: Evaluation Error: Error while evaluating a Function Call, Could not find template 'cm_ssl/*.dev.cargomedia.ch.key' at /tmp/vagrant-puppet/environments/development/manifests/site.pp:8:17 on node www.cm.dev.cargomedia.ch
```

This `cm_ssl` is a fake module just there so that the function `template()` can find the file.

ppp0:
@njam last look, please

njam:
lgtm

PR #2202, opened 2016-06-15
Replace SplEnum with own implementation

fvovan:
After experiencing a lot of segmentation faults that were obviously caused by SplEnum, we decided to implement our own solution.

There are some limitations as to what we can do, e.g. SplEnum supports typecasts for all types, whereas we can only do this for strings. That shouldn't prevent us from using different types though.

cc @njam 
@fvovan maybe this is something for you?

### DIFF
--- a/library/CM/Type/Enum.php
+++ b/library/CM/Type/Enum.php
@@ -1,0 +1,63 @@
+<?php
+
+abstract class CM_Type_Enum {
+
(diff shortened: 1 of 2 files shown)


fvovan:
@alexispeter @njam please review.
I did not find the way to mock constants

njam:
lgtm

PR #2210, opened 2016-06-16
Move FF Integer's slider functionality into FF Slider

tomaszdurka:


tomaszdurka:
@njam wanna review?

njam:
lgtm!

tomaszdurka:
@christopheschwyzer can you take a look if any of the existing Integer form-fields (cm/sk) should be changed to Slider? Thanks.

tomaszdurka:
NOTE: Keep in mind that `Integer` doesn't have min/max now.
@njam @christopheschwyzer should I add it? Maybe as option.

njam:
min/max would be good to have!
E.g. for `Admin_Form_Splitfeature`

christopheschwyzer:
It's all in Admin. I don't use it so can't say. Should be fine.

njam:
There is validation on the model about the integer's range. But it will result in an uncaught exception. We should avoid that, by validating the value on the form level.

@tomaszdurka can you add the config options?

tomaszdurka:
Validation for the integer/float has been added in https://github.com/cargomedia/cm/pull/2212.

@christopheschwyzer I've replaced all the FF Integer usages with FF Slider. All the cases seemed right for slider to me.

PR #2204, opened 2016-06-16
Fluentd handler: add site reference

fvovan:
Currently exception log entries in Loggly have no reference to the *site* they were thrown on.

For the HTTP request there's only `uri: /form/en/190`.

I see two options:
- Add full URI with hostname (see [guidelines](https://github.com/cargomedia/guides/tree/master/logging))
- Add additional "siteName" field to app specific data

@fvovan @tomaszdurka wdyt?

### DIFF
--- a/library/CM/Log/Handler/Fluentd.php
+++ b/library/CM/Log/Handler/Fluentd.php
@@ -77,6 +77,9 @@ class CM_Log_Handler_Fluentd extends CM_Log_Handler_Abstract {
             if ($ip = $request->getIp()) {
                 $formattedRequest['ip'] = (string) $ip;
             }
+            if ($request->hasHeader('host')) {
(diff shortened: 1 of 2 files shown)


fvovan:
@tomaszdurka  please review

library/CM/Log/Handler/Fluentd.php
```diff
@@ -77,6 +77,9 @@ protected function _formatRecord(CM_Log_Record $record) {
             if ($ip = $request->getIp()) {
                 $formattedRequest['ip'] = (string) $ip;
             }
+            if ($request->hasHeader('host')) {
+                $formattedRequest['hostname'] = $request->getHost();
```
tomaszdurka: Should this be only for fluentd? I see we store all the headers for mongodb so it should be ok.

tomaszdurka:
lgtm

PR #2200, opened 2016-06-16
cast string fields in CM_Model_Schema_Definition

fvovan:
like we do for numeric fields.


fvovan:
@njam please review

njam:
lgtm

PR #2154, opened 2016-06-16
Janus: Send context via http requests

tomaszdurka:
Currently we send Context when connecting via websockets to Janus service (we do that on client-side https://github.com/cargomedia/cm/blob/2743980c9a58805a7c5df773cea5d7786647c4f6/library/CM/App.js#L246).
Let's do the same when sending plain http requests. 

--------------------------------------------
We send janus http requests using `CM_Janus_HttpApiClient#_request`. Let's add context parameter here.
Context can contain data about stream (reason for being stopped) or any other valid data.


tomaszdurka:
@fvovan please take a look

library/CM/Log/ContextFormatter/Cargomedia.php
```diff
@@ -0,0 +1,90 @@
...
+        ];
+        return array_merge($hash, $this->getContext($context));
+    }
+
+    /**
+     * @param CM_Log_Context $context
+     * @return array
+     */
+    public function getContext(CM_Log_Context $context) {
+        $hash = [];
```
fvovan: I would rather name it as `$result`. `$hash` is too general

fvovan:
`request` key should be actually `httpRequest`. it was fixed in recent PR
may it's good idea to prefix methods with `format` rather than `get`?

tomaszdurka:
- Yes, I did the merge of upstream after my changes which resulted in mistakes.
- I am always a but sceptic about calling Formatters method `format*`. Mhmm.

fvovan:
>I am always a but sceptic about calling Formatters method format*. Mhmm.

Agree but not sure if it's worse that calling non getters with `get`.

tomaszdurka:
@fvovan please review

Please ignore `get*` namings for the moment (just read your messages, agree).

tomaszdurka:
@njam any thoughts about the Cargomedia formatter?

njam:
As long as it's in line with the [guidelines](https://github.com/cargomedia/guides/tree/master/logging) :)

Quickly looking at it I can see that "referrer" should be "referer", and "user_agent" should be "useragent".
Please check everything again.

library/CM/Janus/HttpApiClient.php
```diff
@@ -4,12 +4,16 @@ class CM_Janus_HttpApiClient {
 
     /** @var \GuzzleHttp\Client */
     protected $_httpClient;
+    
+    /** @var CM_Log_ContextFormatter_Interface */
+    protected $_contextFormatter;
 
     /**
      * @param \GuzzleHttp\Client $httpClient
```
fvovan: please adjust the doc

library/CM/Log/ContextFormatter/Cargomedia.php
```diff
@@ -0,0 +1,80 @@
...
+
+class CM_Log_ContextFormatter_Cargomedia implements CM_Log_ContextFormatter_Interface {
+    
+    /** @var string */
+    protected $_appName;
+
+    /**
+     * @param string $appName
+     */
+    public function __construct($appName) {
```
fvovan: is casting needed here?

fvovan:
The rest looks good except mentioned by Reto wrong keys, `get..` and `$hash`

tomaszdurka:
@fvovan please check

tests/library/CM/Http/Request/AbstractTest.php
```diff
@@ -296,6 +296,10 @@ public function testGetTimeZoneFromCookie() {
         $this->assertNull($timeZone);
     }
 
+    public function testFactoryFromGlobals() {
```
fvovan: is it needed?

tests/library/CM/Log/ContextFormatter/CargomediaTest.php
```diff
@@ -0,0 +1,60 @@
+<?php
+
```
fvovan: please delete one line

fvovan:
the rest lgtm

PR #2212, opened 2016-06-17
Add optional min/max for FF float

tomaszdurka:


tomaszdurka:
@njam please review

njam:
- Validation: Since the min/max is not necessarily known to the user, how about we make the message something like `Value cannot be greater than ${maximum}`?

rest lgtm

tomaszdurka:
@njam please check

njam:
lgtm

PR #2196, opened 2016-06-17
Add "location" to CM_Janus_Server

tomaszdurka:
For https://github.com/cargomedia/sk/issues/3433

Store a `GeoPoint` (lat, lon).
For picking the *closest* server to a given geo point.

@tomaszdurka @fvovan


tomaszdurka:
@fvovan please review
@njam fyi / aka small review

tomaszdurka:
I've reworked `CM_Janus_Configuration` into `CM_Janus_ServerList`.

Reason was I would need to add a lot of method combinations for:
- getting closest server by channel definiton
- getting closest server by plugin name
- getting random server by plugin name
- finding random server by plugin name
- etc.

Instead decided to implement ServerList class to wrapp all those operations of filtering (returning subset) and finders/getters for single instance. While implementing I've noticed that Configuration class is used solely for the server list - so replaced it.

Still small chunk of code for finding closest server incoming.

tomaszdurka:
@fvovan please review `filterByClosestDistanceTo` implementation now.

library/CM/Janus/ServerList.php
```diff
@@ -0,0 +1,129 @@
...
+
+    /**
+     * @param CM_StreamChannel_Definition $channelDefinition
+     * @return CM_Janus_ServerList
+     * @throws CM_Class_Exception_TypeNotConfiguredException
+     * @throws CM_Exception_Invalid
+     */
+    public function filterByChannelDefinition(CM_StreamChannel_Definition $channelDefinition) {
+        $streamChannelClassName = CM_Model_StreamChannel_Media::_getClassName($channelDefinition->getType());
+        $streamChannelClass = new ReflectionClass($streamChannelClassName);
```
fvovan: probably it's better to use http://php.net/manual/en/function.is-subclass-of.php instead of Reflection

library/CM/Janus/ServerList.php
```diff
@@ -0,0 +1,129 @@
...
+            }
+        }
+        return null;
+    }
+
+    /**
+     * @param CM_Geo_Point $location
+     * @return CM_Janus_ServerList
+     */
+    public function filterByClosestDistanceTo(CM_Geo_Point $location) {
```
fvovan: I'm not aware of janus servers topology  and thus  not sure about if this method will do the work.
I mean if we have only 1 janus server per "place" , it will always return only single one.
Is that okay? should it rather return let's say "top x" sorted?

fvovan:
the rest looks good

library/CM/Janus/ServerList.php
tomaszdurka: Well I thought first about returning them sorted, but I thought then we would always have the same one on the top (sorted by distance and default order). This way we return set of closest servers (assuming we will have multiple servers in the same location). We can then run `getRandom` method on the set. That's flexible and aligned with rest of the method.

tomaszdurka:
@fvovan please confirm

fvovan:
lgtm

PR #2213, opened 2016-06-19
Refactor response- and site-matching

njam:
Changes:
- **Replace `Response::match` with `Response::createFromRequest`**: At the moment HTTP response routing works by finding the first response who's `match($request)` returns *true*. Then the *request* is passed to the response's constructor, where it is further processed (for example in `/library-css/en/80/1465656311/all.css` the language is removed, the site id "80" is removed etc). Changes:
 - Introduce a `Reponse::createFromRequest` which does both the request processing (removing some path parts) and the matching. This function is expected to return a *response* instance, if the given request can be handled.
 - Introduce a `CM_Http_Response_Factory` which loops over all responses and returns the first which is a match (using `createFromRequest`)
 - Remove the configurability of the default *page* response (the catch-all response). This has been replaced by a getter `Response::catchAll`
- **Pop a site url's path from the request**, when processing it. So when a site has a url with a path, like `http://example.com/foo`, then a request to `http://example.com/foo/bar` should have `/foo` removed from the path before going into the page response processing. This is accomplished with a new `Request::popPathSiteByMatch()`
- **Move all non-resource responses to default site URL.** Ajax, form, RPC, upload etc. endpoints are now running on the site's URL, and detecting the current site from that URL (instead of passing the site as a parameter in the path).

For example:
Old Ajax URL:
```
https://www.any-site.com/ajax/80
```
New Ajax URL:
```
https://www.denkmal.org/ajax
```

TODO:
- Check (hidden) dependencies:
 - RPC: cm-janus. It's backwards compatible! https://github.com/cargomedia/cm-janus/issues/276


njam:
@tomaszdurka @fvovan it's finished, please have a look.
I've summarized the main changes in the ticket description.

SK adjustment is here: https://github.com/cargomedia/sk/pull/3466

tomaszdurka:
@fvovan could you review this please?

library/CM/Http/Request/Abstract.php
```diff
@@ -219,6 +219,26 @@ public function popPathSite() {
     }
 
     /**
+     * @return CM_Site_Abstract
+     */
+    public function popPathSiteByMatch() {
+        $siteFactory = new CM_Site_SiteFactory();
+        $site = $siteFactory->findSite($this);
+        if (null === $site) {
+            $site = CM_Site_Abstract::factory();
```
fvovan: What is the point of `$site = CM_Site_Abstract::factory();`  ?
Is it a case that it's still valid request if the site is not found by `findSite()`?

library/CM/Http/Response/Resource/Css/Library.php
```diff
@@ -12,7 +12,16 @@ protected function _process() {
...
     }
 
-    public static function match(CM_Http_Request_Abstract $request) {
-        return $request->getPathPart(0) === 'library-css';
+    public static function createFromRequest(CM_Http_Request_Abstract $request, CM_Site_Abstract $site, CM_Service_Manager $serviceManager) {
+        if ($request->getPathPart(0) === 'library-css') {
+            $request = clone $request;
+            $request->popPathPart(0);
+            $request->popPathLanguage();
+            $site = $request->popPathSite();
```
fvovan: is this `$site`  different with passed in `createFromRequest()` arguments one? or you merely wanted to "annotate'"  `popPathSite()` needed  before `$deployVersion = $request->popPathPart(0);` ?

and do we need `null` string check in `CM_Http_Request_Abstract::popPathSite()`?

library/CM/Site/Abstract.php
```diff
@@ -110,6 +110,13 @@ public function getUrl() {
     }
 
     /**
+     * @return CM_Http_UrlParser
+     */
+    public function getUrlParser() {
+        return new CM_Http_UrlParser($this->getUrl());
```
fvovan: may be move initialization inside constructor?

library/CM/Http/Request/Abstract.php
njam: Usually it should not happen. But it can happen in two cases I think:
- The web server has a vhost configured for a domain that we don't expect in the application
- There is *no* `host` header at all (because of HTTP1.0 or because it's in unit tests)

library/CM/Site/SiteFactory.php
```diff
@@ -0,0 +1,55 @@
+<?php
+
+class CM_Site_SiteFactory {
```
fvovan: may be call it just CM_Site_Factory ?

library/CM/Http/Response/Resource/Css/Library.php
njam: This is the *other* case for URLs like
```
http://example.com/library-css/en/123/foo.ext
```
Here the **site id** is passed in the URL (123). This pattern allows us to serve files for a *different site* on any of the site's domains. It is used for serving files on the CDN.
This "null" case is currently used in some requests like RPC, where we don't really care about the site. I think it will be unused after this is merged, and related projects are updated.

library/CM/Http/Response/RPC.php
```diff
@@ -24,7 +24,14 @@ protected function _process() {
         $this->_setContent(json_encode($output));
     }
 
-    public static function match(CM_Http_Request_Abstract $request) {
-        return $request->getPathPart(0) === 'rpc';
+    public static function createFromRequest(CM_Http_Request_Abstract $request, CM_Site_Abstract $site, CM_Service_Manager $serviceManager) {
+        if ($request->getPathPart(0) === 'rpc') {
+            $request = clone $request;
+            $request->popPathPart(0);
+            $request->popPathLanguage();
```
njam: An RPC call uses the "ajax" function on the client, which passes along the language.
So it will be here and we need to pop it.
But it is not available in the RPC function itself.

library/CM/Site/Abstract.php
njam: In some sites `getUrl()` is overridden (Denkmal), so I think we need to do it lazily. But it's a very cheap constructor call!

library/CM/Site/SiteFactory.php
njam: I think it's better to include the type we're creating in the name again, since the first "Site" occurrence is basically a module name. In case there would be other classes in `CM/Site` it would then not be clear anymore for which one the factory is?
That's the case for example in `CM/Http/ResponseFactory.php`.

library/CM/Site/Abstract.php
```diff
@@ -179,30 +191,57 @@ protected function _setModule($name) {
...
-            $hostList[] = $urlCdn->getHost();
+            $matchList[] = [
+                'host' => $urlCdn->getHost(),
+                'path' => $urlCdn->getPath(),
+            ];
         }
 
-        return in_array($request->getHost(), $hostList, true);
+        return Functional\some($matchList, function ($match) use ($host, $path) {
+            $path = new Stringy\Stringy($path);
```
fvovan: I would put it outside of callback

library/CM/App.js
```diff
@@ -1399,9 +1398,10 @@ var CM_App = CM_Class_Abstract.extend({
       }
 
       var urlSite = cm.getUrl();
+      var urlBase = cm.options.urlBase;
       if (0 === url.indexOf(urlSite)) {
```
fvovan: shouldn't `urlBase` be checked here instead of `urlSite` ?

fvovan:
review is finished.

library/CM/App.js
njam: `urlSite` is actually correct, but agree it's a bit confusing.
If the path to load does *not* overlap with the url of the site, then we should always do a full page load, i.e. go into the `else`. If there *is* an overlap, then we request the "full path" including any site-path from the server. Because we can't know for sure yet that it's going to be part of the current site.
Example:
- There's site A with URL `example.com/foo`
- There's site B with URL `example.com/foo/bar`
- The client is on `/foo` and requests `/foo/bar.

In that scenario the site will actually change, but the client doesn't know it beforehand. The server will detect the new site, and respond with an "externalRedirect", so the client can do a reload.

fvovan:
lgtm

njam:
I added two more helper functions for checking and popping a request's path prefix, which will be useful in SK: https://github.com/cargomedia/cm/pull/2213/commits/42938cc9d746d09c6bb3bdbe7e62fcadddef33f8

PR #2211, opened 2016-06-20
CM_Util::jsonEncode: fails with warning

fvovan:
Addressing https://github.com/cargomedia/cm/issues/2221

### DIFF
--- a/library/CM/Log/Handler/MongoDb.php
+++ b/library/CM/Log/Handler/MongoDb.php
@@ -115,6 +115,32 @@ class CM_Log_Handler_MongoDb extends CM_Log_Handler_Abstract {
             $formattedRecord['expireAt'] = new MongoDate($expireAt->getTimestamp());
         }
 
+        $formattedRecord = $this->_sanitizeRecord($formattedRecord); //TODO remove after investigation
+
+        return $formattedRecord;
+    }
+
+    /**
+     * @param array $formattedRecord
+     * @return array
+     */
+    protected function _sanitizeRecord(array $formattedRecord) {
+        $nonUtfBytesList = [];
+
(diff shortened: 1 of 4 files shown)


fvovan:
@njam  please review.
I did both options - log  malformed entries to `extra` and  suppress warnings (if `json_encode` was called outside of log handler).

fvovan:
i reverted `@` because it turned out that it's version and params specific.
5.4 with `@json_encode` signalize `json_last_error()` for `pack("H*" ,'c32e')` param but passes any resource like we have in tests 
```
$resource = fopen(sys_get_temp_dir(), 'r');
CM_Util::jsonEncode(['foo' => $resource]);
```
while 5.5 and 5.6's gives a `json_last_error()` with it.

njam:
As discussed:
- Add try/catch around `json_*`
- Remove variables from exception messages

 then we can release it

njam:
Would still be good to include the $value as an extra in the exception, wdyt?
And is there a way to retrieve the specific error that caused the failure? So we could include it?
And could you also do it for decode?

fvovan:
>Would still be good to include the $value as an extra in the exception, wdyt?
And is there a way to retrieve the specific error that caused the failure? So we could include it?

`$value` is anyway can be found at trace. but I included it to `meta` for convenience sake.
error message or json error code is added as well.
btw if we migrate to >=5.5 we will be able to replace error code with `json_last_error_msg()` 

unfortunately there are cases that will not be perfectly handled eg if we try to encode an object with malformed fields, encoder will throw, but no `$value` caused this error will be shown at log listing.

@njam please review

njam:
lgtm

I opened a new ticket with the original description, to track the progress after we merge this: https://github.com/cargomedia/cm/issues/2221
Let's rather open a separate PR in the future if a change will not (necessarily) completely solve the issue.

PR #2216, opened 2016-06-21
Janus server: return different URL for watching and publishing

tomaszdurka:
Part of https://github.com/cargomedia/puppet-cargomedia/issues/1080

Watching will use a different endpoint (`/edge`).

cc @tomaszdurka

### DIFF
--- a/library/CM/Janus/Server.php
+++ b/library/CM/Janus/Server.php
@@ -76,6 +76,13 @@ class CM_Janus_Server {
     }
 
     /**
+     * @return string
+     */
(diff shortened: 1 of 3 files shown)


tomaszdurka:
@njam please review

tomaszdurka:
As discussed we will use GET parameter instead of different url. Subscribe only address will contain `subscribeOnly=1` query.

njam:
lgtm

did you also want to update usages here?

tomaszdurka:
There are no usages in CM.

PR #2222, opened 2016-06-23
Logger backup handler reliability (NR)

fvovan:
Vladimir:
> I have a suggestion regarding NR handler, 
What if we exclude it from handlers list by default?
the thing is, since it presents on dev environment but never actually sends any data, all mongo handler's exceptions go to nowhere (because backup handler believes that record was processed correctly) what is not convenient. 
Another way is checking if NR is enabled in `CMService_NewRelic_Log_Handler->isHandling()`

cc @tomaszdurka

### DIFF
--- a/library/CMService/Newrelic.php
+++ b/library/CMService/Newrelic.php
@@ -21,7 +21,7 @@ class CMService_Newrelic extends CM_Class_Abstract {
     }
 
     public function setConfig() {
-        if ($this->_getEnabled()) {
+        if ($this->getEnabled()) {
             newrelic_set_appname($this->_appName);
         }
     }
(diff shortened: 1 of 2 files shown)


fvovan:
@njam please review

library/CMService/Newrelic.php
```diff
@@ -93,6 +93,18 @@ public function setCustomMetric($name, $milliseconds) {
     }
 
     /**
+     * @return bool
+     */
+    public function isEnabled() {
+        try {
+            $isEnabled = $this->_getEnabled();
+        } catch (CM_Exception_Invalid $e) {
```
njam: I would not catch here. If the service is enabled but not installed I think it's fine to throw? Maybe just make `getEnabled()` public?

njam:
Is the handler code missing a usage of `isEnabled()`?

fvovan:
>Is the handler code missing a usage of isEnabled()?

oh. bad me. I accidentally  reverted it with 4b5f011.

fvovan:
please review

njam:
lgtm

PR #2220, opened 2016-06-23
Add CM_Janus_ConnectionDescription

tomaszdurka:


zazabe:
@tomaszdurka could you review my changes?

tomaszdurka:
- use URI.js for parsing urls?
- add context to server url on the janus-client not in server?
- do we need StreamChannel js representation?

zazabe:
- context removed
- URI.js  (47kb), not needed now i think, could be introduced later...
- `StreamChannel`, yes, because it implements `CM_ArrayConvertible`

tomaszdurka:
Looks good.

Regarding StreamChannel:
I thought having JS class for ArrayConvertible PHP class is just an option. I saw channel-js it have been used in SK, but could be replaced by `ConnectionDescription`.
I don't mind having JS classes we won't use on client-side, but just thought we should not really pass this object to client-side. Do as you think :)

zazabe:
The problem is, when you have to send back an object to the server, `fromArray` is mandatory in order to be converted back into a PHP object. `ConnectionDescription` doesn't have such method and I don't think we can really recreate it on the server side with what we send to the client, or we must pass all `CM_Janus_Server` properties...

PR #2223, opened 2016-06-24
Add Vector types

fvovan:
For storing geometric coordinates. 
```
CM_Geometry_Vector2 implements CM_Geometry_Vector
CM_Geometry_Vector3 implements CM_Geometry_Vector
CM_FormField_Vector(CM_Geometry_Vector $vector)
````

### CODE REVIEW

**Comment by fvovan** [comment]
I decided to make separate class per each `FormField`
IMO it looks better than numerous conditions.
and I had to use different schema for `Vector` and `FormField` classes. In the first one components of vector are called `x` `y` and in the second they are `xCoordinate`. It was done to avoid possible name conflicts in jquery selectors. It that fine or is it better to unify them?

@njam @tomaszdurka wdyt?

**Comment by njam** [comment]

**Comment by tomaszdurka** [comment]
Looks good although have some questions/suggestions:

- Maybe we should not have `getX`, `getY` on CM_Geometry_Vector interface, but rather something like getCoordinates? For me even thought they both have these methods, it's not really part of interface. 
- Regarding the naming of coordinates. Wouldn't it be simpler to call them all x/y and keep x/yCoordinate for input names only?
- Also you can use a more generic blocks in templates. You've introduced `item-after` block which is working fine, but if you would have 4d vector you would need another `item-after-2` block in the 3d template. There is simple solution to that - use `append` option and use single main block (as below).

`FormField/Geometry_Vector2/default.tpl'`
```smarty
{block name='content'}
  {tag el="input" name="{$name}[xCoordinate]" type="text" value=$xCoordinate class="textinput"}
  {tag el="input" name="{$name}[yCoordinate]" type="text" value=$yCoordinate class="textinput"}
{/block}
```

`FormField/Geometry_Vector3/default.tpl'`
```smarty
{extends file=$render->getLayoutPath('FormField/Geometry_Vector2/default.tpl', 'CM')}

{block name='content' append}
  {tag el="input" name="{$name}[zCoordinate]" type="text" value=$zCoordinate class="textinput"}
{/block}
```

fvovan:
>Maybe we should not have getX, getY on CM_Geometry_Vector interface, but rather something like getCoordinates? For me even though they both have these methods, it's not really part of interface.

But then `coordinates` will be  the array of `[ 'x' => 1, 'y' => 2] `. And... may be just implement `ArrayAccess` instead?

>Regarding the naming of coordinates. Wouldn't it be simpler to call them all x/y and keep x/yCoordinate for input names only?

Agree.

tomaszdurka:
Hmm, im not sure about the need of `getCoordinates`, just thought it could somehow add for needing this interface. I can't see usage of `getX`, `getY` only for 3d Vector. Thought `getCoordinates` could be a method to get a list of them, in order to pass them along somewhere else. Probably `ArrayConvertible` is something like that. Maybe let's drop the interface then?

fvovan:
> I can't see usage of getX, getY only for 3d Vector. Thought getCoordinates could be a method to get a list of them, in order to pass them along somewhere else.

I think in this case we should pass the whole object. Wasn't it the main purpose of their corresponding classes creating? And since these classes are considered to be in `Geometry` namespace, i think  x, y, z getters are okay. But agree - interface is not needed. One class can just extend another.
Fine flexibility probably is not needed, because unlikely we will ever introduce 5 dimensional vectors for geometric purposes. 
Probably it makes sense to reconsider these classes to be more generic, not bound to `geometry`, not bound to particular primitive types. And then they could be parents of `CM_Geometry_Vector2` and `CM_Geometry_Vector3`. But again the question is will we ever need it?

tomaszdurka:
Let's keep `CM_Geometry_Vector2`, `CM_Geometry_Vector3`.
If you want to keep interface it would be to only force `ArrayConvertible` implicitly then.

fvovan:
I thinks there is no need to keep the interface. Let `CM_Geometry_Vector2` be implementing `ArrayConvertible`

fvovan:
@tomaszdurka  please review

tomaszdurka:
lgtm

PR #2227, opened 2016-06-24
make model::equals() not final again

alexispeter:
So this functionality can be mocked in tests.
Particular use-case: Comparing a mocked model to the real model should return true.


alexispeter:
please review @njam

njam:
lgtm

PR #2229, opened 2016-06-27
remove mongoDB log collection validation.

fvovan:
Index is not always properly recreated after reloading app and it's not so important for development env.


fvovan:
@alexispeter please review
@tomaszdurka @njam FYI

alexispeter:
lgtm

PR #2221, opened 2016-06-29
CM_Util::jsonEncode: fails with warning

fvovan:
Addressing https://github.com/cargomedia/cm/issues/2235

### DIFF
--- a/library/CM/Http/Request/Abstract.php
+++ b/library/CM/Http/Request/Abstract.php
@@ -577,6 +577,14 @@ abstract class CM_Http_Request_Abstract {
      */
     public static function factory($method, $uri, array $headers = null, array $server = null, $body = null) {
         $method = strtolower($method);
+
+        $uri = self::_sanitizeUtf($uri);
+        foreach ($server as &$serverValue) {
+            if (is_string($serverValue)) {
+                $serverValue = self::_sanitizeUtf($serverValue);
+            }
+        }
+
         if ($method === 'post') {
             return new CM_Http_Request_Post($uri, $headers, $server, $body);
         }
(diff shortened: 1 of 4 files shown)


fvovan:
I managed to reproduce it but probably it's not the case for unit testing.
Basically the exception occurs either  when fluentd tries to log 404 error or when GA encodes the content of the page. So it's hard to mock without real environment.
And the testing is just checking if 'sanitized' key appears in admin 404 error log record.
We may continue keep an eye on it.

`$_server` had to be sanitized as well because of `php_self` and few other uri related values inside.

@njam please review

njam:
How about doing this in the constructor, to cover all use cases of this class?

Could we add a unit test by using this command you showed to create an invalid string? Then show that it doesn't fail? Or it's only possible in php-fpm?

BTW it would be good to *not* transform this to a PR, so we can keep the ticket open until it's fully resolved. I opened a new one: https://github.com/cargomedia/cm/issues/2235

fvovan:
>Could we add a unit test by using this command you showed to create an invalid string? Then show that it doesn't fail? Or it's only possible in php-fpm?

Yes, it's  a problem with `display_errors`. Besides, received 500 error is a consequence of `CMService_GoogleAnalytics_Client->getHtml(CM_Frontend_Environment)` call came from some included smarty template. and GA on test env. seems to be not configured the same way as for production.

I added the test checking if wrong utf value was converted to valid one inside of constructor.

>BTW it would be good to not transform this to a PR, so we can keep the ticket open until it's fully resolved. I opened a new one: #2235

Oh sorry. I did it "automatically".

njam:
lgtm

PR #2237, opened 2016-06-30
Add a method to clear CM_Janus_ConnectionDescription js instance

zazabe:


zazabe:
@tomaszdurka please review

tomaszdurka:
type 0 is kinda strange, lgtm otherwise

zazabe:
Is it better?

tomaszdurka:
Yeah, better. Still not 100% sure about the clearing, you talked it over with @njam too?

zazabe:
Maybe http://backbonejs.org/#Model-clear could be used instead...

tomaszdurka:
Indeed this looks much better!

zazabe:
I think we should keep those PR changes, because if we use the native `clear`, all getters (like `getIceServerList`) will return `null`...

tomaszdurka:
I would say it's not bad. You mean that it should return empty array? I actually think empty array can be also a valid value. Null is clearly different than empty array.

tomaszdurka:
On the other side we should probably do not run getters on empty models. That's why I am not sure if clearing model is enough. Isn't this that there is no (e.g.) server for this ConnectionDescription and it should be null? Rather then nulling properties of the server?

zazabe:
Finally, I don't need to do this `clear` at all. I thought I would need to keep those instances persistent but actually not... :)

tomaszdurka:
Hooray!

PR #2236, opened 2016-07-01
No record logged if mongodb handler fails

fvovan:
Default config:
```
                'handlersLayerConfigList' => [
                    ['logger-handler-mongodb', 'logger-handler-newrelic'],
                    ['logger-handler-file-error']
                ],
```

If the mongodb handler fails, it seems there is no record logged.
Is this a bug, because if newrelic is disabled (on development), it should go to file, and additional log a logger error?

cc @tomaszdurka

### DIFF
--- a/library/CM/Log/Context.php
+++ b/library/CM/Log/Context.php
@@ -99,8 +99,8 @@ class CM_Log_Context {
      * @throws CM_Exception_Invalid
      */
     public function setExtra(array $extra) {
-        $flattenContext = Functional\flatten($extra);
-        if (Functional\some($flattenContext, function ($el) {
+        $flattenExtra = Functional\flatten($extra);
+        if (Functional\some($flattenExtra, function ($el) {
             return is_object($el);
         })) {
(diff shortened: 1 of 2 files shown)


fvovan:
@njam please review

njam:
@tomaszdurka please review

tomaszdurka:
@vogdb so this was causing the problems on mongodb? What about flattening extra only for the mongodb (and all other handlers which cannot handle nested data)?

vogdb:
me O_O ?

fvovan:
I guess it's me.
@tomaszdurka flattening here is done only to traverse all values. As you can see after checking original `$extra` is used

tomaszdurka:
Was to be @fvovan :>

fvovan:
@tomaszdurka i'm not sure if you noticed my message, because it was sent just few seconds earlier than mine, but i think i would be better just ignore nested values where they are not supported.
Because flattening generally either drops all keys or potentially can lead to values shadowing (when the same keys will appear more that 1 time)

tomaszdurka:
true, lgtm

PR #2240, opened 2016-07-05
Unpublish streamchannel AFTER publisher is deleted

alexispeter:
Because it shouldn't exist anymore when onUnpublish() is called.


alexispeter:
please review @njam

njam:
- PublishTest.php: Rename `$onDeleteBefore`

Otherwise good

PR #2243, opened 2016-07-06
add metaInfo for Exceptions in contextFormatter

fvovan:


fvovan:
is it okay to have `metaInfo`  key for it?
There is little issue regarding displaying values of Exceptions meta keys, each its value is mapped with `CM_Debug_VariableInspector::getDebugInfo()`. So for example strings get additional quotes around them (please see changes in test file). Probably it will create some problems for analyzing but on the other hand I guess this mapping is really needed to handle non primitive instances.
@njam @tomaszdurka wdyt?

fvovan:
@tomaszdurka  please review

tomaszdurka:
lgtm

fvovan:
@tomaszdurka I decided to not unify `getMeta()`  to  `getMetaInfo()` now because of existing mongoDB `meta`  mapping. It's also used in template and if we change it we temporary will lose existing values.
Some transitional code could be added to template, but it probably is not worth because of redundant complexity.

tomaszdurka:
ok

PR #2238, opened 2016-07-06
URL encoded requests lead to json encoding error

fvovan:
Related to https://github.com/cargomedia/cm/issues/2235

Example URL:
/campaign?redirectUnsupported=false&version=1&af=%%aff_id%%&af_source=offerit&oid=%%click_hash%%&utm_source=%%aff_id%%&utm_medium=dfsdf&utm_content=v1&utm_campaign=

> it is %af symbols  (twice) 
they are treated as hex urlencoded values being invalid utf sequences
the question is should we sanitize it?

> urldecoding happens here 
https://github.com/cargomedia/cm/blob/master/library/CM/Http/Request/Abstract.php#L318

```
Cannot json_encode value.
#0 /home/example/releases/20160704104342/vendor/cargomedia/cm/library/CM/Params.php(626): CM_Util::jsonEncode(Array, NULL)
#1 /home/example/releases/20160704104342/vendor/cargomedia/cm/library/CM/Frontend/GlobalResponse.php(194): CM_Params::jsonEncode(Array)
#2 /home/example/releases/20160704104342/vendor/cargomedia/cm/library/CM/Frontend/GlobalResponse.php(151): CM_Frontend_GlobalResponse->_getTreeNodeInitJs(Object(CM_Frontend_TreeNode))
#3 /home/example/releases/20160704104342/vendor/cargomedia/cm/library/CM/Frontend/GlobalResponse.php(138): CM_Frontend_GlobalResponse->_getJs()
#4 /home/example/releases/20160704104342/tmp/smarty/VR_Layout_Index_252_en^5218301f29f7b29cd9a7e5ec0e5146ab7ffa46f3.file.default.tpl.php(182): CM_Frontend_GlobalResponse->getHtml()
#5 /home/example/releases/20160704104342/vendor/smarty/smarty/libs/sysplugins/smarty_internal_templatebase.php(182): content_577a3ecda622d3_83527630(Object(Smarty_Internal_Template))
#6 /home/example/releases/20160704104342/vendor/cargomedia/cm/library/CM/Frontend/Render.php(85): Smarty_Internal_TemplateBase->fetch()
#7 /home/example/releases/20160704104342/vendor/cargomedia/cm/library/CM/Frontend/Render.php(472): CM_Frontend_Render->fetchTemplate('library/VR/layo...', Array, Array)
[...]

PR #2246, opened 2016-07-07
Filter arrays when sanitizing passed data

fvovan:
follow up of #2238

### DIFF
--- a/library/CM/Http/Request/Abstract.php
+++ b/library/CM/Http/Request/Abstract.php
@@ -320,7 +320,18 @@ abstract class CM_Http_Request_Abstract {
         $querySanitized = [];
         foreach ($query as $key => $value) {
             $key = self::_sanitizeUtf($key);
-            $querySanitized[$key] = self::_sanitizeUtf($value);
+
+            if (is_array($value)) {
+                array_walk_recursive($value, function (&$innerValue) {
+                    if (is_string($innerValue)) {
+                        $innerValue = self::_sanitizeUtf($innerValue);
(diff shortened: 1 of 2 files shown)


fvovan:
@fauvel please review.
@zazabe fyi

fauvel:
Could be more elegant (e.g. recursive implementation of `_sanitizeUtf`), but lgtm!

library/CM/Http/Request/Abstract.php
```diff
@@ -320,7 +320,18 @@ public function setUri($uri) {
         $querySanitized = [];
         foreach ($query as $key => $value) {
             $key = self::_sanitizeUtf($key);
-            $querySanitized[$key] = self::_sanitizeUtf($value);
+
+            if (is_array($value)) {
+                array_walk_recursive($value, function (&$innerValue) {
+                    if (is_string($innerValue)) {
```
fauvel: If you're still not sure that it must be a string at this stage... maybe you'd better use a safe, recursive implementation for `_sanitizeUtf`!

fvovan: i'm almost 100% sure it is a string, but added it just to double check, 
i'll test it without checking condition a bit later.

njam: What happens if we pass numbers? Will they be string or integer here?

fvovan: They will be casted to strings by `mb_parse_url()`

PR #2247, opened 2016-07-07
Introduce JsonSerialized JS model

zazabe:


njam:
lgtm

Just not sure about the classname. Usually we try to introduce groups/modules of files within `library/`.
How about for example `CM/Frontend/JsonModel.js`?

njam:
Or as discussed `CM/Frontend/JsonSerializable.js`

zazabe:
@vogdb please review

vogdb:
- `CM_StreamChannel_Definition` still uses `CM_JsonSerialized_Abstract`.
- `CM_StreamChannel_Definition` Why do we add `bind` and `unbind`? They were not presented before and `JsonSerializable` does not implement/override them.
- CM_Frontend_JsonSerializable.equals: `var keys = _.union(this.keys(), jsonSerialized.keys());`. Should two instances be equal if their keys have `_.difference > 0` ?
- in CM_Frontend_JsonSerializable.sync

  ```js
  this.set(key, externalValue);
  var attrs = {};
  _.each(this.changedAttributes(), function(val, key) {
    attrs[key] = val instanceof CM_Frontend_JsonSerializable ? val.toJSON() : val;
  });
  _.extend(resultTarget, attrs);
  ```
  Why we can't do?
  ```js
  this.set(key, externalValue);
  var attrs = {};
  attrs[key] = val instanceof CM_Frontend_JsonSerializable ? val.toJSON() : val;
  _.extend(resultTarget, attrs);
  ```
  Only `key` should be changed by the previous `this.set`, isn't it?

zazabe:
Sorry, I changed a bit the implementation because I'm testing it at the same time on the app... 

- _CM_StreamChannel_Definition still uses CM_JsonSerialized_Abstract._   
  Fixed
- _CM_StreamChannel_Definition Why do we add bind and unbind? They were not presented before and JsonSerializable does not implement/override them._   
   Sorry, pushed by mistake, reverted (it will be another PR)
- _CM_Frontend_JsonSerializable.equals: var keys = _.union(this.keys(), jsonSerialized.keys());. Should two instances be equal if their keys have _.difference > 0 ?_   
  It's not only about having the same keys but also about checking their values and catch missing keys in both instances, but maybe I misunderstand your comment...
- _in CM_Frontend_JsonSerializable.sync ..._  
  Agree, could be simplified!

vogdb:
@zazabe tests fail

vogdb:
@zazabe tests still faile

vogdb:
Also I still don't get the logic of default values. For example lets take the test you have `"sync with defaults"`. Lets put this into it:
```js
var fooRemovedDefault = foo.clone();
fooRemovedDefault.unset('foo');
result = foo.sync(fooRemovedDefault);
console.log(result);
console.log(foo.get('foo'));
```

The output will be
```js
null
defaultValue
```

Is it impossible to remove default value key from the object?

zazabe:
that's was the propose of this "default" support... but actually it's not needed anymore and I agree, it makes no sense at the end. Defaults should be used only for the init, here it's used more like something to preserve attributes...

Anyhow, let me remove that feature :)

About those tests that fail: seems to be related to https://github.com/ariya/phantomjs/issues/10522, because Travis is using `PhantomJS 1.9.8` ;(

zazabe:
@vogdb please review

zazabe:
@vogdb wait, I'm doing some adjustments...

vogdb:
oh, no. Adjustments! 😢

zazabe:
ok, no more adjustments! could you review again @vogdb ?

vogdb:
in `install-phantomjs.sh` should we replace `...setup-XXX)` with `...setup-{VERSION})`? Honestly I don't know the syntax of bash.

zazabe:
`mktemp` will replace those `XXX` by a random string, it doesn't matter so much but I think it's good to have a unique folder for such setup operation.

vogdb:
lgtm then!

PR #2127, opened 2016-07-07
Make internals.js more readable

alexispeter:
Each entry should go on a separate line, to make it easier to spot differences.


alexispeter:
please review @tomaszdurka

tomaszdurka:
Can you tell me what is difference between `jsonEncode` and `encode`. I understand these are scalar values and there should be no real difference, right?

alexispeter:
The difference is that jsonEncode() supports pretty-printing, i.e. each value on a separate line.

tomaszdurka:
Ok then lgtm.

PR #2248, opened 2016-07-08
Janus RPC: invalid session should be warning

fvovan:
Currently when we receive RPC calls with invalid session information we get this exception:
```php
CM_Exception_Nonexistent: Session `e443d82711136b1668032a1e680cd2ba` does not exist. in /home/example/releases/20160707172200/vendor/cargomedia/cm/library/CM/Session.php on line 35
0. {main} at /home/example/serve/public/index.php line 0
1. CM_Http_Handler->processRequest(CM_Http_Request_Post) at /home/example/releases/20160707172200/public/index.php line 8
2. CM_Http_Response_Abstract->process() at /home/example/releases/20160707172200/vendor/cargomedia/cm/library/CM/Http/Handler.php line 28
3. CM_Http_Response_RPC->_process() at /home/example/releases/20160707172200/vendor/cargomedia/cm/library/CM/Http/Response/Abstract.php line 48
4. CM_Http_Response_Abstract->_runWithCatching(Closure, Closure) at /home/example/releases/20160707172200/vendor/cargomedia/cm/library/CM/Http/Response/RPC.php line 21
5. CM_Http_Response_RPC->{closure}() at /home/example/releases/20160707172200/vendor/cargomedia/cm/library/CM/Http/Response/Abstract.php line 273
6. call_user_func_array([], []) at /home/example/releases/20160707172200/vendor/cargomedia/cm/library/CM/Http/Response/RPC.php line 18
7. [internal function] at line
8. CM_Session->__construct('e443d82711136b1668032a1e680cd2...') at /home/example/releases/20160707172200/vendor/cargomedia/cm/library/CM/Janus/RpcEndpoints.php line 92
```

This should be a warning only, as it's based on user input.

I see two ways to accomplish that:
- A: Catch the exception in the RPC functions, and handle it locally
- B: Add it to the config's `$config->CM_Http_Response_RPC->exceptionsToCatch`

Either way we should probably introduce a specific exception `CM_Session_UnknownSessionIdException`?

@tomaszdurka @fvovan wdyt?

### DIFF
--- a/library/CM/Exception/UnknownSessionId.php
+++ b/library/CM/Exception/UnknownSessionId.php
@@ -1,0 +1,5 @@
+<?php
+
+class CM_Exception_UnknownSessionId extends CM_Exception {
+
+}
--- a/library/CM/Janus/RpcEndpoints.php
+++ b/library/CM/Janus/RpcEndpoints.php
@@ -21,7 +21,7 @@ class CM_Janus_RpcEndpoints {
 
         $server = $janus->getServerList()->findByKey($serverKey);
         $sessionParams = CM_Params::factory(CM_Params::jsonDecode($sessionData), true);
-        $session = self::_getSession($sessionParams);
+        $session = self::_getSession($sessionParams->getString('sessionId'));
         $user = $session->getUser(true);
 
         $channelKey = (string) $channelKey;
@@ -89,7 +89,7 @@ class CM_Janus_RpcEndpoints {
 
         $server = $janus->getServerList()->findByKey($serverKey);
         $sessionParams = CM_Params::factory(CM_Params::jsonDecode($sessionData), true);
-        $session = self::_getSession($sessionParams);
+        $session = self::_getSession($sessionParams->getString('sessionId'));
         $user = $session->getUser(true);
 
         $channelKey = (string) $channelKey;
@@ -210,14 +210,14 @@ class CM_Janus_RpcEndpoints {
     }
 
     /**
-     * @param CM_Params $sessionParams
+     * @param string $sessionId
      * @return CM_Session
-     * @throws CM_Exception_Nonexistent
+     * @throws CM_Exception_UnknownSessionId
      */
-    protected static function _getSession(CM_Params $sessionParams) {
+    protected static function _getSession($sessionId) {
         try {
-            return new CM_Session($sessionParams->getString('sessionId'));
-        } catch (CM_Exception_Nonexistent $e) {
+            return new CM_Session((string) $sessionId);
+        } catch (CM_Exception_UnknownSessionId $e) {
             $e->setSeverity(CM_Exception::WARN);
             throw $e;
         }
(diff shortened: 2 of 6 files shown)


fvovan:
@njam please review.

njam:
- I would keep `->getString('sessionId')` outside, to document the parameters that are used in the "rpc_*" function, wdyt?
- Let's also introduce this specific exception so we can catch it with confidence? Something like `CM_Session_UnknownSessionIdException`?

fvovan:
>Let's also introduce this specific exception so we can catch it with confidence? Something like CM_Session_UnknownSessionIdException?

Do you mean new exception for rethrowing?

njam:
I thought to throw `CM_Session_UnknownSessionIdException` directly in `CM_Session`. The widespread use of `CM_Exception_Nonexistent` is a bit dangerous (because we translate it to 404). In this case I think a different exception is easy to introduce and better. wdyt?

fvovan:
>I thought to throw CM_Session_UnknownSessionIdException directly in CM_Session. The widespread use of CM_Exception_Nonexistent is a bit dangerous (because we translate it to 404). In this case I think a different exception is easy to introduce and better. wdyt?

hmm. how do we want to translate this new Exception for Page responses? Isn't it the same 404?

njam:
I don't think `CM_Exception_Nonexistent` from *sessions* is or should lead to a 404.

fvovan:
>I don't think CM_Exception_Nonexistent from sessions is or should lead to a 404.

new type sgtm then.
but what will be instead of 404? just server error?

njam:
I think error only happens in the RPC response, so 404 is not relevant here.
Note that the in `CM_Session::getUser()` we throw a different exception, which I think is the one relevant for page/ajax requests?

fvovan:
I added `CM_Exception_UnknownSessionId` 
and should it be a warning only for `CM_Janus_RpcEndpoints` ? if no , we could originally throw it as a warning without modifying of severity.
@njam wdyt

njam:
btw what do you think of adding a `CM_Params::getSession($key)`? This is the usual way we handle user input, and we could then throw `CM_Exception_InvalidParam`.

fvovan:
makes sense!
@njam please review

njam:
Let's also allow to get existing session objects through `CM_Params::getSession()`. This is quite useful for retrieving parameters in a view from a template for example.
```
{some_function session=$session}
```

You can use `CM_Params::getObject()` for that I think, with a custom getter.

fvovan:
>Let's also allow to get existing session objects through CM_Params::getSession().

hmm, not sure I got you right.. 
so far we have `CM_Params::getSession($key)` to retrieve existing sessions by key.
Do you suggest to also allow some other existing sessions with this method, or just to modify  `CM_Params::getSession($key)` to use `CM_Params::getObject()`?

njam:
Currently `getSession("foo")` works if the params contains `{foo: 12345}` (where "12345" is a session-id). But it should also work if params contains `{foo: CM_Session(12345)}`, so already a session object instead of a session identifier. This is what `getObject()` does.

fvovan:
@njam  thanks for the elaboration.
please review.

njam:
lgtm

PR #2250, opened 2016-07-08
Include mainPaths content in browserify cache key

njam:


njam:
@zazabe @fauvel please review

njam:
@zazabe @fauvel please review

zazabe:
- not sure if `$cacheKeyMainPaths` is needed, do you think we could have diff `$mainPaths` but with the same concatenated `$involvedFiles` content?

njam:
As discussed it's needed because the list of entry points affects directly the generated code.

fauvel:
lgtm! shortening the cache key by calling md5 iteratively definitely makes sense, too.

PR #2252, opened 2016-07-08
Introduce CM_Paging_List

zazabe:


zazabe:
@vogdb please review

vogdb:
**pain in the ass mode** 💩 💩 💩 

- `CM_Frontend_AbstractTrait.applyImplementation`

  ```js
    _.each(this.getTraitProperties(), function(traitPropertyName) {
      var property = this.getTraitProperty(traitPropertyName);
  ```
  why not ?
  ```js
    _.each(this.traitProperties, function(traitPropertyName, property) {
  ```

- `CM_Frontend_AbstractTrait.isImplementedBy`. The same as above.
- `CM_Frontend_AbstractTrait`. Name of property is named as in `traitPropertyName` in `applyImplementation` and as `traitProperty` in `isImplementedBy`. It misleads.
- Quite often I see constructions like `_(CM_Frontend_AbstractTrait).clone()`. Why we can't do `_.clone(CM_Frontend_AbstractTrait)` ? I guess that I am missing something here.
- Double `assert.ok(foo.bar());` in `QUnit.test("applyImplementation"`
- In tests. I thought when we use `define` we ought to declare dependencies as arguments

  ```js
  define(["CM/Frontend/AbstractTrait", "CM/Frontend/SynchronizableTrait"], function() {
  ```

  ```js
  define(["CM/Frontend/AbstractTrait", "CM/Frontend/SynchronizableTrait"], function(CM_Frontend_AbstractTrait, CM_Frontend_SynchronizableTrait) {
  ```
- `CM_Frontend_JsonSerializable.sync` Why did you replace?

  ```js
  if (this.compatible(localValue)) {
  ```
  with
  ```js
  if (this.compatible(localValue)) {
  ```
- `CM_Paging_List.equals` should be improved. Please try use this data in tests of nested:

  ```js
  var list = new CM_Paging_List([
    new CM_Frontend_JsonSerializable({
      id: 1,
      foo: 1,
      '1.1': new CM_Paging_List([
        new CM_Frontend_JsonSerializable({id: 'a', bar: 1})
      ])
    }),
    new CM_Frontend_JsonSerializable({id: 2})
  ]);
  var clone = new CM_Paging_List([
    new CM_Frontend_JsonSerializable({
      id: 1,
      foo: 1,
      '1.1': new CM_Paging_List([
        new CM_Frontend_JsonSerializable({id: 'a', bar: 1})
      ])
    }),
    new CM_Frontend_JsonSerializable({id: 2})
  ]);
  ```
  Shall we use 
  ```js
  return list.every(function(item) {
  ```
  instead of
  ```js
  return list.any(function(item) {
  ```
  ?

zazabe:
> I thought when we use define we ought to declare dependencies as arguments   
 
Those libs are not CommonJS/AMD modules, we can't get them as arguments of the `define` callback.
Also, I agree that using `define` without any module definition could looks strange, but `require` is async and using it in tests will lead to some troubles with the `baseUrl`, because [it is changed for each `suite`](https://github.com/cargomedia/cm/blob/master/tests/client/bootstrap.js#L67).   
On the other hand, declaring tests with `define()` will ensure that all dependencies are loaded in the correct context before executing the test: 
  - [with `require()`](https://cloud.githubusercontent.com/assets/401598/16759688/6f71b15c-4818-11e6-9559-87b652c15ce5.png)
  - [with `define()`](https://cloud.githubusercontent.com/assets/401598/16759689/6f72925c-4818-11e6-8472-a3963b9f176b.png)   
  
> `CM_Frontend_JsonSerializable.sync` Why did you replace?

Just to trig the "remove" event on any instance that support triggering events...

> CM_Paging_List.equals should be improved.   

Good catch!


Should be fine, please kindly review again @vogdb

vogdb:
- CM_Frontend_AbstractTrait. I would still replace

  ```js
  _.each(this.traitProperties, function(property, traitPropertyName) {
  ```
  with
  ```js
  _.each(this.traitProperties, function(property, propertyName) {
  ```
  or
  ```js
  _.each(this.traitProperties, function(traitProperty, traitPropertyName) {
  ```

- `equals` still requires an improvement. Try this test:

  ```js
  QUnit.test("equals2", function(assert) {
    var list = new CM_Paging_List([
      new CM_Frontend_JsonSerializable({id: 2}),
      new CM_Frontend_JsonSerializable({id: 2}),
      new CM_Frontend_JsonSerializable({id: 3})
    ]);

    var clone = new CM_Paging_List([
      new CM_Frontend_JsonSerializable({id: 3}),
      new CM_Frontend_JsonSerializable({id: 3}),
      new CM_Frontend_JsonSerializable({id: 2})
    ]);

    assert.notOk(list.equals(clone));
  });
  ```

zazabe:
> `equals` still requires an improvement

Please consider da0b31b

vogdb:
> Please consider da0b31b

So, it is impossible to have two instances with the same data?

vogdb:
what about this then?
```js
  QUnit.test("equals: redundant NOT ids", function(assert) {
    var list1 = new CM_Paging_List([
      new CM_Frontend_JsonSerializable({notId: 2}),
      new CM_Frontend_JsonSerializable({notId: 2}),
      new CM_Frontend_JsonSerializable({notId: 3})
    ]);

    var list2 = new CM_Paging_List([
      new CM_Frontend_JsonSerializable({notId: 3}),
      new CM_Frontend_JsonSerializable({notId: 3}),
      new CM_Frontend_JsonSerializable({notId: 2})
    ]);

    assert.ok(list1.size(), 2);
    assert.ok(list2.size(), 2);
    assert.ok(list1.equals(list2));
  });
```

zazabe:
Same `id`, same data should be fine. Also, please not that the ordering has no impact. see 740b3a3d45f3b6ace4af628fc17e8e66333ef457

zazabe:
ok, this test will failed:
```
var list1 = new CM_Paging_List([
      new CM_Frontend_JsonSerializable({foo: 2}),
      new CM_Frontend_JsonSerializable({foo: 3})
    ]);

    var list2 = new CM_Paging_List([
      new CM_Frontend_JsonSerializable({foo: 2}),
      new CM_Frontend_JsonSerializable({foo: 3})
    ]);
```

I'll fix it

vogdb:
Well, this test passes for me. I think that we have lgtm here!

vogdb:
ah, no. It failes >_<. Need to refresh with cache clear

zazabe:
OK, more tests, more reliable, please review :)

vogdb:
lvgtm!

PR #2254, opened 2016-07-11
Use more specific naming for CM_StreamChannel_Definition binding methods

zazabe:


zazabe:
@tomaszdurka please review

tomaszdurka:
Could be. Maybe `addStreamEventListener` ? :)

zazabe:
ok, just for you ;)

tomaszdurka:
<3

zazabe:
I added some methods to bind actions directly, @vogdb could you take a look?

vogdb:
Can we use `/^(\d+)\s+(.+)$/` instead of `/^(\S+)\s+(.+)$/` and remove `isNumber` check? Have you tried? If yes then lgtm

zazabe:
The first match is a string, then we get the type number from `cm.action.types`, so the `RegExp` and the `isNumber` condition sgtm, wdyt?

vogdb:
lgtm!

PR #2253, opened 2016-07-11
Logger: pass "httpRequest.query" to Loggly

fvovan:
Currently in Loggly we have the following keys in `httpRequest`: ip, hostname, uri, referer, useragent, method.

It would be good to also have `query`, which allows to identify for example what ajax request was sent. This information is available in mongodb/admin.

@fvovan @tomaszdurka  wdyt? If you agree, could you also open a PR on *guides*?

### DIFF
--- a/library/CM/Http/Request/Abstract.php
+++ b/library/CM/Http/Request/Abstract.php
@@ -278,6 +278,17 @@ abstract class CM_Http_Request_Abstract {
     }
 
     /**
+     * @return array
+     */
+    public function findQuery() {
+        try {
+            return $this->getQuery();
+        } catch (CM_Exception_Invalid $e) {
+            return [];
+        }
+    }
+
+    /**
      * @param array $query
      */
     public function setQuery(array $query) {
--- a/library/CM/Log/ContextFormatter/Cargomedia.php
+++ b/library/CM/Log/ContextFormatter/Cargomedia.php
@@ -40,6 +40,11 @@ class CM_Log_ContextFormatter_Cargomedia implements CM_Log_ContextFormatter_Inte
                 'uri'    => $request->getUri(),
                 'method' => $request->getMethodName(),
             ];
+            try {
+                $formattedRequest['query'] = $request->getQuery();
+            } catch (CM_Exception_Invalid $e) {
+                //CM_Http_Request_Post can throw.
+            }
             if (array_key_exists('http_referer', $serverArray)) {
                 $formattedRequest['referer'] = (string) $serverArray['http_referer'];
             }
--- a/library/CM/Log/ContextFormatter/Cargomedia.php
+++ b/library/CM/Log/ContextFormatter/Cargomedia.php
@@ -40,11 +40,7 @@ class CM_Log_ContextFormatter_Cargomedia implements CM_Log_ContextFormatter_Inte
                 'uri'    => $request->getUri(),
                 'method' => $request->getMethodName(),
             ];
-            try {
-                $formattedRequest['query'] = $request->getQuery();
-            } catch (CM_Exception_Invalid $e) {
-                //CM_Http_Request_Post can throw.
-            }
+            $formattedRequest['query'] = $request->findQuery();
             if (array_key_exists('http_referer', $serverArray)) {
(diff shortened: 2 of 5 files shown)


fvovan:
@njam please review
guides changes are here https://github.com/cargomedia/guides/pull/58/

tomaszdurka:
agree

njam:

library/CM/Log/ContextFormatter/Cargomedia.php
```diff
@@ -40,6 +40,11 @@ public function formatContext(CM_Log_Context $context) {
                 'uri'    => $request->getUri(),
                 'method' => $request->getMethodName(),
             ];
+            try {
+                $formattedRequest['query'] = $request->getQuery();
+            } catch (CM_Exception_Invalid $e) {
+                //CM_Http_Request_Post can throw.
+            }
```
tomaszdurka: I see more and more usages of try/catch for normal control-flow. Let's try not use it this way. Let's either add non-throwing `hasQuery` or make `getQuery` not throw.

tomaszdurka:
See comments!

fvovan:
I would not change existing signatures.
what about `findQuery()` method?
@tomaszdurka please have a look

library/CM/Http/Request/Abstract.php
```diff
@@ -278,6 +278,17 @@ public function getQuery() {
     }
 
     /**
+     * @return array
+     */
+    public function findQuery() {
+        try {
+            return $this->getQuery();
+        } catch (CM_Exception_Invalid $e) {
```
tomaszdurka: This is still using catch for normal flow. What abut implementing `findQuery` on both `Abstract` and `Post` and return null when cannot extract in `Post`? Then `getQuery` could throw on null.

fvovan: So do you suggest to make `findQuery()` to be an alias for `getQuery()` for all types of requests except  `Post`.
And this implementation will be moved to `post`, right?
I would not make any changes to current `getQuery()` signature but rather prefer to wrap it.

tomaszdurka: Nono, I rather meant that `findQuery` should return `[]|null`, while `getQuery` should return `[]` or throw.

I thought that we could use `Post::getQuery` code for new `Post::findQuery`, but instead of throwing we could return `null` - as it's invalid query.
New `getQuery` would be simple proxy for `findQuery` with option throw when query is null.

fvovan: I'm not sure, if "deferred" exception creating would be good.
Currently we can have only `Cannot extract query from body` error and may treat all `null` as this error.
But in case of other errors we might need to have 'error codes'. We could _return_ `Exception` (not throw it), but that may be weird.
should we concern about other possible kinds of failures?

tomaszdurka: Yes agree that it's not great to treat all nulls as errors.

Quick thought:

Why do we consider invalid query in first place. Maybe rather encode the query in the constructor (or better outside) and pass along correct query. We would throw exception there. Wdyt?

fvovan: >Maybe rather encode the query in the constructor (or better outside) and pass along correct query. We would throw exception there.

hmm.  do you mean _decode_ query? In `Request` constructor? 
Probably it would be good. But it's again about changing of class behaviour. It will fail faster and i guess in most cases it's okay, but in general it could be differently.

However I think we should not inject already parsed query from outside as it is the area of class responsibility along with cookies, headers etc

tomaszdurka: Ok. I give up. 
I still think it's wrong to catch/return in here, but I cannot easy-convince you to the heavy lifting. I also agree that currently Post class is responsible by decoding the data depending on the encoding and probably that is bad.

Let's say it's good.

fvovan: Sorry Tomasz. I don't 100% like current solution as well, but as you can see we can't get rid of `try catch` in some obvious way, and tries to invent a tricky workaround here and `CM_Http_Request` reworking  may be not worth of 2 invocations from the logger. 
I believe it would be better to do when we have some really important reasons.

Issue #2235, opened 2016-06-29
CM_Util::jsonEncode: fails with warning


njam:
> E_WARNING: json_encode(): Invalid UTF-8 sequence in argument

```
#0 [internal function]: CM_ExceptionHandling_Handler_Abstract->handleErrorRaw(2, 'json_encode(): ...', '/home/example/...', 540, Array)
#1 /home/example/releases/20160616110844/vendor/cargomedia/cm/library/CM/Util.php(540): json_encode(Array, 0)
#2 /home/example/releases/20160616110844/vendor/cargomedia/cm/library/CM/Params.php(620): CM_Util::jsonEncode(Array, NULL)
#3 /home/example/releases/20160616110844/vendor/cargomedia/cm/library/CM/MongoDb/Client.php(46): CM_Params::jsonEncode(Array)
#4 /home/example/releases/20160616110844/vendor/cargomediaedia/cm/library/CM/Log/Handler/MongoDb.php(45): CM_MongoDb_Client->insert('cm_log', Array, Array)
#5 /home/example/releases/20160616110844/vendor/cargomedia/cm/library/CM/Log/Handler/Abstract.php(40): CM_Log_Handler_MongoDb->_writeRecord(Object(CM_Log_Record))
#6 /home/example/releases/20160616110844/vendor/cargomedia/cm/library/CM/Log/Handler/Layered.php(76): CM_Log_Handler_Abstract->handleRecord(Object(CM_Log_Record))
#7 /home/example/releases/20160616110844/vendor/cargomedia/cm/library/CM/Log/Handler/Layered.php(34): CM_Log_Handler_Layered->_addRecordToLayer(Object(CM_Log_Record), 0)
#8 /home/example/releases/20160616110844/vendor/cargomedia/cm/library/CM/Log/Logger.php(145): CM_Log_Handler_Layered->handleRecord(Object(CM_Log_Record))
#9 /home/example/releases/20160616110844/vendor/cargomedia/cm/library/CM/Log/Logger.php(92): CM_Log_Logger->_addRecord(Object(CM_Log_Record))
#10 /home/example/releases/20160616110844/vendor/cargomedia/cm/library/CM/ExceptionHandling/Handler/Abstract.php(104): CM_Log_Logger->addMessage('Application err...', 400, Object(CM_Log_Context_App))
[...]

PR #2258, opened 2016-07-12
Improve utf-8 sanitizing for logger handlers

fvovan:
https://github.com/cargomedia/cm/issues/2235

>As discussed let's sanitize all the log entry values for all affected handlers (mongodb and fluentd and file?).
In mongodb for records that had values in invalid encoding we'll add a key sanitizedFields that contains a key/value hash of all sanitized values for debugging.
loggerNotifications: { sanitizedFields: {key1: value, key2: value} }
This should displayed next to the "context" in the admin.

### DIFF
--- a/layout/default/Component/LogList/default.tpl
+++ b/layout/default/Component/LogList/default.tpl
@@ -61,9 +61,14 @@
       {if !empty($log.context)}
         <div class="toggleNext">Context</div>
         <div class="toggleNext-content">
-          {if (!empty($log.context))}
-            <pre>{print_r($log.context, true)}</pre>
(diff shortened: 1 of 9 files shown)


fvovan:
@tomaszdurka please review

tomaszdurka:
Looks good, but does this covers logging sanitation problems? I believe we talked together about something like that.

fvovan:
>but does this covers logging sanitation problems? 

Yes, initial point was to enable sanitization for fluentd.

#2235 is kind of meta issue which will be resolved when all json encode errors are eliminated.

Issue #1778, opened 2015-05-27
Introduce CM_TransferObject for easier frontend data sending/receiveing


tomaszdurka:
Small draft of TransferObject and its children.
```php
class CM_TransferObject {

    /** @var CM_ArrayConvertible */
    private $_object;

    /**
     * @param CM_ArrayConvertible $object
     */
    public function __construct(CM_ArrayConvertible $object) {
        $this->_object = $object;
    }

    /**
     * @return string
     */
    public function encode() {
        return CM_Params::encode($this->_object);
    }

    /**
     * @param string $encodedObject
     * @return static
     */
    public static function decode($encodedObject) {
        $decodedObject = CM_Params::decode($encodedObject);
        return new static($decodedObject);
    }
}


class CM_TransferModelObject extends CM_TransferObject {

    /**
     * @param CM_Model_Abstract $model
     */
    public function __construct(CM_Model_Abstract $model) {
        parent::__construct($model);
    }
}

class CM_ConversationTransferModelObject extends CM_TransferModelObject {

    public function __construct(SK_Entity_Conversation $conversation, SK_User $viewer) {
    }
}
```

njam:
We should also consider different use cases:
- Encoding for serialization/deserialization *only*
- Add additional fields for frontend (JS)

We've solved it for now by introducing `ArrayConvertible::toArrayIdOnly()` additional to `toArray()`, see https://github.com/cargomedia/cm/pull/1984

PR #2259, opened 2016-07-15
Introduce CM_Frontend_JsonSerializable

zazabe:
Useful to pass auto-converted `CM_Frontend_JsonSerializable` JS model to the client.

### DIFF
--- a/library/CM/Frontend/JsonSerializable.php
+++ b/library/CM/Frontend/JsonSerializable.php
@@ -1,0 +1,29 @@
+<?php
+
+class CM_Frontend_JsonSerializable implements JsonSerializable {
+
+    /** @var array */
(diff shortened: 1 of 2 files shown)


zazabe:
@tomaszdurka please review

zazabe:
@tomaszdurka 💋

tomaszdurka:
Is this something like transferObject? Generally looks good.

zazabe:
Yes, a simplified version of https://github.com/cargomedia/cm/issues/1778, @njam told me about it :)

Do you think we should support `CM_ArrayConvertible` also?

zazabe:
ok, let's merge it like this, we can always improve it if needed

PR #2262, opened 2016-07-20
Set user online on CM_Model_StreamChannel_Message_User subscription

zazabe:
`CM_User_OfflineJob` could be executed just before the user subscription when the page is reloaded. Then, the user will be considered as offline and he will not receive any published events from the `CM_Model_StreamChannel_Message_User`...

Let's consider that when a user subscribes to the stream channel, he is actually online!

### DIFF
--- a/library/CM/Model/StreamChannel/Message/User.php
+++ b/library/CM/Model/StreamChannel/Message/User.php
@@ -8,8 +8,11 @@ class CM_Model_StreamChannel_Message_User extends CM_Model_StreamChannel_Message
     }
 
     public function onSubscribe(CM_Model_Stream_Subscribe $streamSubscribe) {
-        if (!$streamSubscribe->getUser()->getOnline()) {
-            $streamSubscribe->getUser()->setOnline(true);
+        if ($this->hasUser()) {
+            $user = $streamSubscribe->getUser();
+            if ($user && !$user->getOnline()) {
+                $user->setOnline(true);
+            }
         }
     }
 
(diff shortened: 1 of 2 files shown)


zazabe:
@alexispeter please review

cc @njam @tomaszdurka

alexispeter:
* Add an if-clause analogous to onUnsubscribe
* Test could be shorter, no need to for the first subscribe/unsubscribe-part
```php
public function testOnSubscribe() {
        $user = CMTest_TH::createUser();
        /** @var CM_Model_StreamChannel_Message_User $channel */
        $channel = CM_Model_StreamChannel_Message_User::createStatic([
            'key'         => CM_Model_StreamChannel_Message_User::getKeyByUser($user),
            'adapterType' => CM_MessageStream_Adapter_SocketRedis::getTypeStatic(),
        ]);

        $this->assertSame(false, $user->getOnline());
        $subscribe = CMTest_TH::createStreamSubscribe($user, $channel);
        CMTest_TH::reinstantiateModel($user);
        $this->assertSame(true, $user->getOnline());
    }
```

zazabe:
@alexispeter could you check again?

alexispeter:
Sorry, I meant "Add an if-clause to onSubscribe(), analogous to the one in onUnsubscribe() `$this->hasUser()`

zazabe:
ok, like this?

alexispeter:
lgtm

tomaszdurka:
We already set user to online whenever he loads any page. See `CM_Session->setUser()`. Not sure if we need extra one. But agree, it's nice to tie to `user-ws-channel`.

zazabe:
@tomaszdurka yes, but in some cases, eg. if the page is reloaded, it could happen that `CM_User_OfflineJob` is executed just after the `CM_Session->setUser()` code, then the user will be switch back into offline, which will prevent him from receiving stream channel messages...

PR #2255, opened 2016-07-20
CM_Process: ignore PIDs that have been forked elsewhere

fauvel:
Currently `CM_Process` assumes it's the only place where process forks are created. When *any* child process dies (`$pid = pcntl_wait()`) it looks up the died process' PID in a table it keeps for reference about all forked processes (`_getForkHandlerByPid()`).

If a child process is forked anywhere else and this process dies, then this lookup fails.
I suggest we silently ignore dying children that we don't know about.


fauvel:
@alexispeter please review

alexispeter:
lgtm

PR #2263, opened 2016-07-20
Log the HTTP header "referer" only if available

zazabe:
_Found this bug when I tried to use the the logger for some debugging..._

### DIFF
--- a/library/CM/Log/Formatter/Text.php
+++ b/library/CM/Log/Formatter/Text.php
@@ -21,18 +21,20 @@ class CM_Log_Formatter_Text extends CM_Log_Formatter_Abstract {
             ]);
         }
         if (null != $httpRequest) {
-            $headers = $httpRequest->getHeaders();
             $server = $httpRequest->getServer();
-            $httpRequestText = '{type} {path} {proto}, host: {host}, ip: {ip}, referer: {referer}, user-agent: {agent}';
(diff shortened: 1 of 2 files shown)


zazabe:
@fvovan please review

library/CM/Log/Formatter/Text.php
```diff
@@ -29,7 +30,7 @@ public function renderContext(CM_Log_Record $record) {
                 'proto'   => isset($server['SERVER_PROTOCOL']) ? $server['SERVER_PROTOCOL'] : '',
                 'host'    => $httpRequest->getHost(),
                 'ip'      => $httpRequest->getIp(),
-                'referer' => $httpRequest->getHeader('referer'),
+                'referer' => isset($headers['referer']) ? $headers['referer'] : '',
```
fvovan: It's better to check it with `$httpRequest->hasHeader('referer')` rather using arrays directly.

zazabe:
@fvovan is it better like that?

fvovan:
Yes, it's better but for simplicity i would leave formatting template the same and just  modify  former line 33
`'referer' => $httpRequest->hasHeader('referer')? $httpRequest->getHeader('referer') : '',`

zazabe:
ok, as you want ;)

fvovan:
lgtm!

PR #2264, opened 2016-07-22
Add missing seconds to smarty_function_date_time

dlondero:


dlondero:
@tomaszdurka please review

tomaszdurka:
Well, I understood that we should add those only for Admin.

dlondero:
@tomaszdurka please review. Used `show*` flag name like we do in `date` plugin with `showTime` and `showWeekday`.

dlondero:
@tomaszdurka ?

dlondero:
@fvovan please review.

fvovan:
lgtm, but probably `$showSeconds` could be inlined without any lack of readability.

dlondero:
@fvovan agree. Please check.

fvovan:
lgtm

PR #2265, opened 2016-07-25
Support "ManagerAware" jobs

alexispeter:
Set servicemanager inside CM_Jobdistribution_Cli::startWorker() if job is *ManagerAware*

### DIFF
--- a/tests/library/CM/Jobdistribution/DelayedQueueTest.php
+++ b/tests/library/CM/Jobdistribution/DelayedQueueTest.php
@@ -51,4 +51,14 @@ class CM_JobDistribution_DelayedQueueTest extends CMTest_TestCase {
 
         $this->assertSame(3, $instantiateMethod->getCallCount());
     }
+
+    public function test_instantiateJobSetServiceManager() {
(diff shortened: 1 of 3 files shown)


alexispeter:
please review @tomaszdurka

tests/library/CM/Jobdistribution/DelayedQueueTest.php
```diff
@@ -51,4 +51,14 @@ public function testQueueOutstanding() {
 
         $this->assertSame(3, $instantiateMethod->getCallCount());
     }
+
+    public function test_instantiateJobSetServiceManager() {
```
tomaszdurka: Out of curiousity is it our convention to use `_` in test name when method had it?

tomaszdurka:
lgtm

tests/library/CM/Jobdistribution/DelayedQueueTest.php
alexispeter: Well, it's a convention to call your tests `test<nameofthemethod>`, and since the name of the method has an underscore, so does the test.

PR #2266, opened 2016-07-27
smarty_function_date_time: allow to pass timeZone param as string

dlondero:
At the moment in `smarty_function_date_time` timeZone param must be a `DateTimeZone` or null (because used in CM_Frontend_Render::getFormatterDate()). Usage is quite painful as requires to create the object in a component or page, pass it to a template and then use it.

I'd allow timeZone param as a string using one of the [PHP supported timezones](http://php.net/manual/en/timezones.php) in order to create the `DateTimeZone` based on that. Of course w/o BC break because passing an object would still be allowed.

@njam wdyt?


dlondero:
@fvovan please review

fvovan:
lgtm

PR #2269, opened 2016-08-01
Track site session activity

fvovan:


fvovan:
@njam or @tomaszdurka could you review it please

tomaszdurka:
on it

library/CM/Model/User.php
```diff
@@ -215,16 +215,34 @@ public function getEnvironment() {
...
-    protected function _updateLatestActivity() {
+    /**
+     * @return CM_Site_Abstract|null
+     */
+    public function getLastSessionSite() {
+        $siteType = $this->_get('lastSessionSite');
+        if (!$siteType) {
+            return null;
+        }
+        return CM_Site_Abstract::factory($siteType);
```
tomaszdurka: You really think we should return default here?

tomaszdurka: My bad. Looks ok.

tomaszdurka:
This solution will definitely work, but I have my concerns. There is lot of tangling to achieve extra update. Session needs request, request needs session, should it?

What about:
- move `updateLatestActivityThrottled` out of user (keep `_updateLatestActivity` there)
- use throttling mechanism around `_updateLatestActivity` and new `_updateLastSessionSite($site)` in request class where `Session->start()` is called
- no need to store request on session, pass site directly

fvovan:
thanks!
>move updateLatestActivityThrottled out of user (keep _updateLatestActivity there)
use throttling mechanism around _updateLatestActivity and new _updateLastSessionSite($site) in request class where Session->start() is called
no need to store request on session, pass site directly

I would keep `updateLatestActivityThrottled` as it was before then. Basically there is no better place for it.
And then just  add  `_updateLastSessionSite($user, $site)`  after each `Session->start()` (without throttling control). it will be 2 db updates, but session starting is not so frequent.
wdyt?

fvovan:
@tomaszdurka please review

library/CM/Http/Request/Abstract.php
```diff
@@ -576,6 +578,21 @@ public function isSupported() {
     }
 
     /**
+     * @param CM_Session $session
+     */
+    private function _updateLastActivitySite(CM_Session $session) {
+        if (!$this->hasHeader('host') || !$session->hasUser()) {
```
tomaszdurka: You don't need to pass session, you can get user by `Request->getViewer(null)`.

library/CM/Http/Request/Abstract.php
```diff
@@ -576,6 +578,21 @@ public function isSupported() {
...
+            return;
+        }
+        $siteFactory = new CM_Site_SiteFactory();
+        $site = $siteFactory->findSite($this);
```
tomaszdurka: I really don't like this factory with `findSite` method expecting `$request`.

library/CM/Model/User.php
```diff
@@ -302,11 +322,12 @@ protected static function _createStatic(array $data) {
...
-            'activityStamp' => time(),
-            'site'          => $siteType,
-            'languageId'    => $languageId,
-            'currencyId'    => $currencyId,
+            'createStamp'     => time(),
+            'activityStamp'   => time(),
+            'site'            => $siteType,
+            'languageId'      => $languageId,
+            'currencyId'      => $currencyId,
+            'lastSessionSite' => $siteType,
```
tomaszdurka: Having this here btw really hurts. But it's simplest I guess...

fvovan: hmm but why is hurts?

library/CM/Http/Request/Abstract.php
fvovan: Me too, but what is the another way to get the site here?

library/CM/Http/Request/Abstract.php
fvovan: I'm afraid it will not work.
`$this->_session` is initialized after updating `$this->_updateLastActivitySite($session);` call,
And `getViewer()` needs it for proper `hasSession()` call.

library/CM/Http/Request/Abstract.php
fvovan: I did kind of memoization for  `$site`.
value got from `popPathSiteByMatch` will be used.
@tomaszdurka wdyt?

njam:
I think "lastSessionSite" might not be needed, please discuss with Tomasz

tomaszdurka:
As discussed with @fvovan possibly we can drop this!

fvovan:
not needed now

PR #2268, opened 2016-08-01
Fluentd adapter sends incomplete stack-trace

fvovan:
More to the point, the file and line where an exception is instantiated is handled as separate information in exceptions and the adapter just sends the trace without this information.

For example, this is what our exception-handler prints to stdout.
```
CM_Exception_Invalid:  in /home/vagrant/sk/scripts/test.php on line 16
    0. {main} /home/vagrant/sk/scripts/test.php:0
    1. A->foo() /home/vagrant/sk/scripts/test.php:21
    2. A->bar() /home/vagrant/sk/scripts/test.php:12
```
and this is (part of) what the fluentd-logger sends to loggly. Note that `/home/vagrant/sk/scripts/test.php on line 16` is missing.
```php
array(1) {
  'exception' =>
  array(4) {
    'type' =>
    string(20) "CM_Exception_Invalid"
    'message' =>
    string(0) ""
    'stack' =>
    string(111) "#0 /home/vagrant/sk/scripts/test.php(12): A->bar()\n#1 /home/vagrant/sk/scripts/test.php(21): A->foo()\n#2 {main}"
    'metaInfo' =>
    array(0) {
    }
  }
}
```
The missing information could either be inserted into the trace, or sent as separate search-able fields. It wouldn't be bad to be able to search for filename and line, in case of non-unique error-messages.
Another related issue is that Exception::getTraceAsString() prints the trace in reverse order, which can be highly confusing.

fyi @tomaszdurka

### DIFF
--- a/library/CM/ExceptionHandling/SerializableException.php
+++ b/library/CM/ExceptionHandling/SerializableException.php
@@ -104,6 +104,11 @@ class CM_ExceptionHandling_SerializableException {
             foreach (array_reverse($exception->getTrace()) as $row) {
                 $trace[] = self::_extractTraceRow($row);
             }
+            $trace[] = [
+                'code' => '{throw}',
+                'line' => $this->line,
+                'file' => $this->file,
+            ];
             $this->trace = $trace;
         } catch (Exception $e) {
             $this->trace = null;
--- a/library/CM/Log/ContextFormatter/Cargomedia.php
+++ b/library/CM/Log/ContextFormatter/Cargomedia.php
@@ -60,10 +60,11 @@ class CM_Log_ContextFormatter_Cargomedia implements CM_Log_ContextFormatter_Inte
         if ($exception = $context->getException()) {
             $serializableException = new CM_ExceptionHandling_SerializableException($exception);
             $result['exception'] = [
-                'type'     => $serializableException->getClass(),
-                'message'  => $serializableException->getMessage(),
-                'stack'    => $serializableException->getTraceAsString(),
-                'metaInfo' => $serializableException->getMeta(),
+                'type'          => $serializableException->getClass(),
+                'message'       => $serializableException->getMessage(),
+                'stack'         => $serializableException->getTrace(),
+                'stackAsString' => $serializableException->getTraceAsString(),
+                'metaInfo'      => $serializableException->getMeta(),
             ];
         }
         return $result;
--- a/library/CM/Log/ContextFormatter/Cargomedia.php
+++ b/library/CM/Log/ContextFormatter/Cargomedia.php
@@ -59,12 +59,19 @@ class CM_Log_ContextFormatter_Cargomedia implements CM_Log_ContextFormatter_Inte
 
         if ($exception = $context->getException()) {
             $serializableException = new CM_ExceptionHandling_SerializableException($exception);
+            $stack = $serializableException->getTrace();
+            if (!empty($stack)) {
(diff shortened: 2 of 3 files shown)


fvovan:
I added the line where exception in throws to stack trace and made this field to be an array.
Old `stack` is  moved to `stackAsString`. (so it's similar to mongoDB)
@tomaszdurka  please review

library/CM/ExceptionHandling/SerializableException.php
```diff
@@ -104,6 +104,11 @@ private function _extract(Exception $exception) {
             foreach (array_reverse($exception->getTrace()) as $row) {
                 $trace[] = self::_extractTraceRow($row);
             }
+            $trace[] = [
+                'code' => '{throw}',
```
tomaszdurka: Usually this is called `main`?

tomaszdurka: Oh wait, noes. Usually this doesn't fit into exception trace, right?

library/CM/Log/ContextFormatter/Cargomedia.php
```diff
@@ -60,10 +60,11 @@ public function formatContext(CM_Log_Context $context) {
...
             $serializableException = new CM_ExceptionHandling_SerializableException($exception);
             $result['exception'] = [
-                'type'     => $serializableException->getClass(),
-                'message'  => $serializableException->getMessage(),
-                'stack'    => $serializableException->getTraceAsString(),
-                'metaInfo' => $serializableException->getMeta(),
+                'type'          => $serializableException->getClass(),
+                'message'       => $serializableException->getMessage(),
+                'stack'         => $serializableException->getTrace(),
+                'stackAsString' => $serializableException->getTraceAsString(),
```
tomaszdurka: Not sure about this one anymore. I thought that we still will send only stack (as string) by Cargomedia formatter - as defined in our logging guide.

I thought that we will use stack-array to generate stack-string if present and otherwise use stack-string from start.

I'm simply against sending 2 of these in Cargomedia formatter.

library/CM/ExceptionHandling/SerializableException.php
fvovan: `{main} ` is here https://github.com/cargomedia/cm/pull/2268/files/4449284b717b0b1e78789bf2e48e8b38148247fc#diff-94d3d51c9fe64cebc29f778dc1477105L99

yes, these details don't get to trace, because they are direct Exception's attributes.

library/CM/Log/ContextFormatter/Cargomedia.php
fvovan: okay. I constructed string version from the array. please review
But as for me duplication is not really a problem, because it's just repercussion of `Exception`'s nature.
Logging schema could be updated to include optional 'stack as array'
upsides were we could use non reversed order. And sometimes `stackAsArray` and `stack` are slightly different.

alexispeter:
>> I thought we still have the array for CM-admin site.

> Actually we have, but as for me the downside is you can't search without accessing mongodb console.

I didn't mean we should make the mongodb log searchable like this. But maybe loggly could do it ouf-of-the-box if we just sent the right data.

fvovan:
>I didn't mean we should make the mongodb log searchable like this. But maybe loggly could do it ouf-of-the-box if we just sent the right data.

Yes, I meant the same.

fvovan:
@tomaszdurka please review

tomaszdurka:
lgtm

PR #2272, opened 2016-08-05
Use right staleness for CM_PagingSource_Array

alexispeter:
CM_PagingSource_Array::getStalenessChance() should return the staleness of the PagingSource it's based on, if any. Currently, it just returns 0.

### DIFF
--- a/library/CM/PagingSource/Array.php
+++ b/library/CM/PagingSource/Array.php
@@ -64,6 +64,13 @@ class CM_PagingSource_Array extends CM_PagingSource_Abstract {
         return array_slice($this->_getData(), $offset, $count);
     }
 
+    public function getStalenessChance() {
+        if ($this->_source) {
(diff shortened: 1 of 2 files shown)


alexispeter:
please review @tomaszdurka

library/CM/PagingSource/Array.php
```diff
@@ -64,6 +64,13 @@ public function getItems($offset = null, $count = null) {
         return array_slice($this->_getData(), $offset, $count);
     }
 
+    public function getStalenessChance() {
```
tomaszdurka: Shouldn't this be:
```php
protected function _getStalenessChance() {
```
?

tomaszdurka: Oh my bad. Got confused.

tomaszdurka:
Looks good!

PR #2279, opened 2016-08-09
lazy onLoadItemsRaw callback

fvovan:


njam:
I thought we would move the invocation of `_onLoadItemsRaw()` from `_getItemsRaw()` to `_processItem()`. Then we don't need to pass this `$skipCallback`?

fvovan:
>I thought we would move the invocation of _onLoadItemsRaw() from _getItemsRaw() to _processItem(). Then we don't need to pass this $skipCallback?

Yes, it was original idea, but i found one gotcha.
the problem is `_getItemsRaw()` is called from `getItemsRaw()` and it's a place where we would not put `onLoadItemsRaw()`. But i'm afraid in the end some of `getItemsRaw()` external invocations will really rely on this callback.

njam:
As discussed let's move the `_onLoadItemsRaw()` *only* into `CM_Paging_ModelAbstract`, like this:
```php
    protected function _processItem($itemRaw) {
        if (null === $this->_modelList) {
            $this->_modelList = $this->_populateModelList($this->getItemsRaw());
        }
```

fvovan:
@njam please review

njam:
I think we should use a different variable type for the "not initialized" state, like `null` (because empty array is a valid paging state)?

njam:
Otherwise lgtm

njam:
- CM/Util.php: applyOffset: Include "count" in exception metaInfo

Rest lgtm

PR #2282, opened 2016-08-15
Make sure language cache is invalidated when a new javascript version is set

njam:


vogdb:
looks good but before merging. Shouldn't we have smth like `$where .= " AND 'javascriptVersion' = '{$javascriptVersion}'"` instead of `$where .= " AND '{$javascriptVersion}' = '{$javascriptVersion}'"` ?

njam:
It's a trick. It will result in something like `WHERE '12345' = '12345'` which will always be true. But when the version changes then the condition will become `WHERE '123456' = '123456'` and a new cache key will be used.

@tomaszdurka please review too

tomaszdurka:
Ok, so this condition is to use automatic caching mechanism. The hacks!
lgtm

PR #2281, opened 2016-08-15
Include "deployVersion" in RPC responses

vogdb:
On the client side `cm.rpc()` calls `cm.ajax()` which expects a `deployVersion` key present in the response.

Currently that key is only available in the *ajax* response, but not in the *RPC* response.


vogdb:
@njam please review

njam:
lgtm

PR #2275, opened 2016-08-16
Move functionality of CM_Paging_List to CM_Paging_Abstract

zazabe:
Keep only `_construct` of back-end in CM_Paging_List. All the rest should go to CM_Paging_Abstract.
The usecase is to create sync-able SK_Paging_MailboxConversation_Abstract front-end counterpart.

@zazabe please add your thoughts


zazabe:
It works, but all Paging that will go through `CM_Params:encode()` will be exposed like that: `{"_class":"", "items": "<$paging->getItems()>"}`, so the payload could be pretty heavy, wdyt @vogdb @tomaszdurka @njam ?

vogdb:
- Can we exclude items by default?
 - Shall we implement the Paging API for those lists?

zazabe:
> Can we exclude items by default?

we could, be it means that each paging that need a JS representation will have to override this `jsonSerialize` method, with the same code i guess...

> Shall we implement the Paging API for those lists?

pagination on the client side? mmmh... it's not needed right now but i would let you implement that in a new PR :D!

zazabe:
as discussed with @njam, it's preferable to explicitly define `jsonSerializable()` for each `Paging` that we want to expose as a collection on the client side. 

@vogdb please review

vogdb:
well, looks fine

PR #2283, opened 2016-08-17
Add defaults to loadTypekit script

vogdb:


vogdb:

christopheschwyzer:
Shouldn't this be (only 2 arguments)?
```js
 config = _.defaults(config, {
    scriptTimeout: 3000,
    async: true
  });
```

http://underscorejs.org/#defaults

vogdb:
Well, in case of two arguments, we will modify `config`. I thought it's not nice to modify it cause it can be used outside theoretically.

christopheschwyzer:
Isn't `config` modified anyways here: `config = ` - Well, I'm no js dev - I don't understand this fully :) Seems to work, therefore lgtm.

vogdb:
Well, you can play with this
```js
function modifiesArgument(argument) {
  _.defaults(argument, {hello: 'world'});
}
function immutableArgument(argument) {
  _.defaults({}, argument, {hello: 'world'});
}

var sample = {foo: 'bar'};

immutableArgument(sample);
console.log('After immutableArgument', sample);

modifiesArgument(sample);
console.log('After modifiesArgument', sample);
```

christopheschwyzer:
I see. Thanks for the explanation :)

PR #2285, opened 2016-08-19
Ensure that users always subscribe to CM_Model_StreamChannel_Message_User

zazabe:
`CM_Model_StreamChannel_Message_User` is used to [detect the presence of users](https://github.com/cargomedia/cm/blob/master/library/CM/Model/StreamChannel/Message/User.php#L10), so we should always subscribe to it.
On the other hand, user subscribes to [channels on demand](https://github.com/cargomedia/cm/blob/master/library/CM/App.js#L994), when they need to listen to some events.

We should enforce the `CM_Model_StreamChannel_Message_User` subscription, maybe at the `CM_App` level, wdyt @vogdb @njam ?

### DIFF
--- a/library/CM/App.js
+++ b/library/CM/App.js
@@ -985,6 +985,12 @@ var CM_App = CM_Class_Abstract.extend({
     _channelDispatchers: {},
 
     ready: function() {
+      if (!cm.options.stream.enabled) {
+        return;
+      }
+      if (!cm.options.stream.channel) {
+        return;
+      }
       if (cm.options.stream.channel.key && cm.options.stream.channel.type) {
         var channel = cm.options.stream.channel.key + ':' + cm.options.stream.channel.type;
         this._subscribe(channel);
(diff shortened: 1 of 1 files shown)


zazabe:
@vogdb please review

njam:
In `cm.stream.bind()` we have: 
```js
    bind: function(channelKey, channelType, namespace, callback, context, allowClientMessage) {
      var channel = channelKey + ':' + channelType;
      if (!cm.options.stream.enabled) {
        return;
      }
```
So if "stream" is disabled we should not do it. Then again I think it might not be relevant anymore that we can disable that functionality. So maybe it's best to remove this flag?

The if should check if `cm.options.stream.channel` is present, otherwise it will fail to access `.key` and `.type`? The same check is present in `CM_View_Abstract.bindAction()`.

zazabe:
Seems that [ `CM_MessageStream_Service::_adapter`](https://github.com/cargomedia/cm/blob/master/library/CM/MessageStream/Service.php#L5) could be `null`, then `cm.options.stream.enabled ==  false`, so I think we should keep it.

@njam please review

njam:
lgtm

PR #2294, opened 2016-08-22
Integrate Webfontloader

vogdb:


vogdb:
@christopheschwyzer please review. Especially the last commit. Did I put `@transitionDuration: 150ms;` into the right file/line? Should we use `Typekit` in names of `mixins.less`? Shall we rename it to something as `webFontLoader`?

client-vendor/before-body/03-webfontloader/webfontloader.js
```diff
@@ -0,0 +1,9 @@
+(function(d) {
+  var config = document.documentElement.getAttribute('data-web-font-loader');
+  if (config) {
+    WebFontConfig = JSON.parse(config);
+    var wf = d.createElement('script'), s = d.scripts[0];
+    wf.src = 'https://ajax.googleapis.com/ajax/libs/webfont/1.6.16/webfont.js';
```
christopheschwyzer: Host it ourselves?!

christopheschwyzer:
- Variable looks good!
- Yes I would rename it like suggested. 
- I think this part is not necessary:
```css
  html.wf-active &, html.wf-inactive & {
    opacity: 1;
  }
```

client-vendor/before-body/03-webfontloader/webfontloader.js
vogdb: Where? If we include it in `before-body` then it won't be async loaded. It will be always present.

christopheschwyzer: @njam I think we don't have anything like that yet? Shouldn't go into `before-`nor `after-body` - rather something like "on demand". Ideas?

njam: I would say we should either A) load it from Google or B) inline it completely into before-body.
To inline it into before-body on a per-site basis is a bit tricky, so I would go with A), just load it from Google.

vogdb:

christopheschwyzer:
Is this part necessary? Not sure - I'm always confused… `isset($webFontLoaderConfig) &&`

christopheschwyzer:
Changed the `mixins.less` again :)I think it makes sense to show the fallback font if it couldn't be loaded.
Tested. Lgtm!

vogdb:
@christopheschwyzer I've made a change to `isset($webFontLoaderConfig) &&`. If you have no objections, we can merge as soon as build is successful.

christopheschwyzer:
👍

PR #2295, opened 2016-08-22
CM_Paging_ModelAbstract cache misses 

fvovan:
ErrorException: E_NOTICE: Undefined index: a:2:{s:2:"id";s:8:"16771914";s:4:"type";s:1:"2";} in .../CM/Paging/ModelAbstract.php on line 36

Most likely caused by https://github.com/cargomedia/cm/pull/2279

### DIFF
--- a/library/CM/Paging/Abstract.php
+++ b/library/CM/Paging/Abstract.php
@@ -295,12 +295,19 @@ abstract class CM_Paging_Abstract extends CM_Class_Abstract implements Iterator,
         $this->_flattenItems = (bool) $state;
     }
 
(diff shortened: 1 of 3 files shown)


fvovan:
Problem is we don't reset `_modelList` after `_clearItems()` called in parent class `CM_Paging_Abstract`
so I had to override methods that call  `_clearItems()`.
Not sure about `getRandItem()` though.
@njam please review

njam:
Let's do it in `_clearItems()`.

njam:
lgtm

PR #2296, opened 2016-08-22
Make sure to iterate over all items when instantiating models in paging

njam:
_getItemsRaw() includes all items, while getItemsRaw() only includes the current page.


njam:
@fvovan please review

fvovan:
lgtm

PR #2291, opened 2016-08-24
Use arrays in fluentd records.

fvovan:
Via https://github.com/cargomedia/puppet-cargomedia/issues/1341

To reduce number of indexed keys (which is limited by loggly) we need to encode some of not important for searching purposes keys as arrays.
eg instead of 

```
{
  "important_search_key" : { 
    "sub_key1" : {},
    "sub_key2" : {}
  },
  "just_information_key" : { // it's a map 
    "not_needed_sub_key1" : {},
    "not_needed_sub_key2" : {}
  }
} 
```
we want 

```
{
  "important_search_key" : {
    "sub_key1" : {},
    "sub_key2" : {}
  },
  "just_information_key" : [ //here is an array
    { "id" : "not_needed_sub_key1", "value" : {} },
    { "id" : "not_needed_sub_key2", "value" : {} }
  ]
}
``` 

Let's also drop `httpRequest.query.viewInfoList` completely from any logs!
(it's a lot of data and was never used so far)

### DIFF
--- a/client-vendor/after-body/jquery.floatbox/jquery.floatbox.js
+++ b/client-vendor/after-body/jquery.floatbox/jquery.floatbox.js
@@ -86,7 +86,7 @@
       this._getFocusElement().focus();
       this.$floatbox.trap();
 
-      this.$layer.data('floatbox', this);
+      $element.data('floatbox', this);
       $element.trigger('floatbox-open');
     },
     close: function() {
@@ -97,7 +97,7 @@
       if (this.$parent.length) {
         this.$parent.append($element);
       }
-      this.$layer.removeData('floatbox');
+      $element.removeData('floatbox');
       this.$layer.remove();
       $viewport.children('.floatbox-layer:last').addClass('active');
       if (lastFocusedElement) {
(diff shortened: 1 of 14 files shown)


fvovan:
@njam @tomaszdurka please review.
I hope it's okay to be in `ContextFormatter`.

>Let's also drop httpRequest.query.viewInfoList completely from any logs!

should it be deleted only from this handler or from mongo and NR as well? (we currently don't have common preprocessors)

njam:
@tomaszdurka can you review this?

njam:
We're losing log records because of the field limit.
We should deploy this to see if it helps asap.
@tomaszdurka can you review this, so we can release it on Monday?

fvovan:
Actually we lose only some search abilities. Non indexed fields could not be found with smart search 
https://community.loggly.com/customer/portal/questions/8299618-cors-requests-tags-and-maxfieldslimitreached-?b_id=50

njam:
We lose the capability of using the data in a structured way: we can't search for it, we can't aggregate those fields (for stats), and we can't copy the value from the interface.

The only thing we can do is to search the value in its original json stringified form. E.g. search for `"foo":"bar"`
> You would have to run the searches as raw text.

fvovan:
@tomaszdurka please confirm

tomaszdurka:
confirmed!

fvovan:
FInally got an "official response" on Friday.

>Thank you for your patience.
>I tried your both JSON structures at my end. 
>And as per my findings, changing the JSON structure as you suggested in the second case will help use less number of fields.
>Also "just_information_key.id" and "just_information_key.value" will be counted as indexed keys.

PR #2299, opened 2016-08-24
Add delayed job queue to service manager

alexispeter:
depends on https://github.com/cargomedia/cm/pull/2305

To improve testability.


alexispeter:
please review @njam

njam:
- Instead of adding an $overwrite parameter I think it would be better to create a new service manager instance in this test. This way the test also is more isolated from the "global state". wdyt?

alexispeter:
But how would you inject the new service manager into the streamchannel in this example?

njam:
Singleton. Let's try to fix it! I was just discussing it with Tomasz, here our thoughts:

It would be better if the `onUnsubscribe()` implementation wouldn't be in the streamchannel *model*, as we would like to keep models light, meaning they shouldn't have too many dependencies (like a service manager).
We were discussing the possibility of **introducing a separate "streamchannel processor"**, which would would have functions `onSubscribe`/`onUnsubscribe()` and could be attached as *behaviour* to a streamchannel through the config. This one would be servicemanager-aware.
But this would have the downside that we separate the *streamchannel* and the *streamchannel processor* into two files with the current design, which might be difficult to understand.

Alternatively we suggest to **pass the servicemanager to `onUnsubscribe()` et al** as an intermediary solution. Basically we keep the functions in the same file, but upgrade these 4 functions to servicemanager awareness.

Currently `onUnsubscribe()` is called in `Stream::_onDeleteBefore()` which is a problem, because we no context whatsoever there. Roughly we'll need to do the following:
- Make `CM_MediaStreams_Service` servicemanager aware
- Make `CM_MediaStreams_StreamRepository` servicemanager aware
- Move the `onUnsubscribe()` et al calls into `CM_MediaStreams_StreamRepository`.

wdyt? let's do a call to discuss if needed.

alexispeter:
Sounds good all in all. Taking the calls out of the models would also fix other issues, e.g. #2260.
But can we keep the overwrite-param for a convenient way to mock service-instances for those cases in the meantime?

njam:
As discussed let's instead use the global service-manager, once we have the changes from https://github.com/cargomedia/cm/issues/2305, and see if we can live with that.

alexispeter:
please review @njam

njam:
lgtm

Please remember to mark as a breaking change.

alexispeter:
waiting on https://github.com/cargomedia/sk/issues/3832

PR #2303, opened 2016-08-25
Add deprecation note to ServiceManager::getInstance()

njam:
Just to act as a reminder to not use the singleton access everywhere.
Instead we should find ways how to pass the manager whenever we use it.

@tomaszdurka @alexispeter please review


alexispeter:
lgtm

PR #2305, opened 2016-08-25
Unit tests: restore service-manager

alexispeter:
- In `TH::clearEnv()` restore the service-manager that was backed up in `init()`.
- Introduce `SM::setInstance()`
- Introduce `SM::replaceInstance()`

### DIFF
--- a/library/CM/Service/Manager.php
+++ b/library/CM/Service/Manager.php
@@ -117,6 +117,17 @@ class CM_Service_Manager extends CM_Class_Abstract {
 
     /**
      * @param string $serviceName
+     * @param mixed  $instance
+     */
+    public function replaceInstance($serviceName, $instance) {
+        if ($this->has($serviceName)) {
+            $this->unregister($serviceName);
+        }
+        $this->registerInstance($serviceName, $instance);
+    }
+
+    /**
+     * @param string $serviceName
      */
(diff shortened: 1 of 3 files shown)


alexispeter:
please review @njam

alexispeter:
tons of side-effects :>

njam:
@tomaszdurka @zazabe can you review this?

zazabe:
I thought that we would introduce a specific class only used in tests that extends `CM_Service_Manager`, isn't it?

alexispeter:
> I thought that we would introduce a specific class only used in tests that extends CM_Service_Manager, isn't it?

I don't think so.

please review @tomaszdurka 

I used deep cloning to isolate the services that are instantiated during bootstrap from the actual tests, but I don't think it's really needed.

alexispeter:
please review @tomaszdurka

njam:
please review @tomaszdurka

njam:
please review @tomaszdurka
 Show all checks

tomaszdurka:
@zazabe I think i proposed that, but seeing that replaceInstance, could be part of serviceManager and get/set instances are global ones which should be removed by... 2019 :) it's ok.

lgtm

btw it would be great to remove all the staticness of the TH and use it as trait maybe.

PR #2306, opened 2016-08-25
CM_Log_Logger::addMessage() fails if the logger has no global context

fvovan:
If `CM_Log_Logger` is instantiated without any parameter, `CM_Log_Logger::addMessage()` fails with this error:
```
PHP Fatal error:  __clone method called on non-object in library/CM/Log/Logger.php on line 88
```

@fvovan ?


fvovan:
indeed.
@zazabe please review

zazabe:
lgtm

PR #2301, opened 2016-08-26
Allow to open a floatbox in a floatbox

zazabe:
Currently disallowed in `jquery.floatbox.js`:
```js
  $.fn.floatOut = function(options) {
    return this.each(function() {
      if (!$(this).parents('.floatbox-layer').addBack().data('floatbox')) {
        var floatbox = new $.floatbox(options);
        floatbox.show($(this));
      }
    });
  };
```

I think the reason for this check is just to make sure we don't open the *same element* in a floatbox multiple times.
Can we instead store that data attribute on the `$element` that is actually floating?

@zazabe maybe for you?
cc @christopheschwyzer

### DIFF
--- a/client-vendor/after-body/jquery.floatbox/jquery.floatbox.js
+++ b/client-vendor/after-body/jquery.floatbox/jquery.floatbox.js
@@ -86,7 +86,7 @@
       this._getFocusElement().focus();
       this.$floatbox.trap();
 
-      this.$layer.data('floatbox', this);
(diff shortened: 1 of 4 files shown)


zazabe:
@christopheschwyzer wanna review it?

zazabe:
minor change only, please review

christopheschwyzer:
Not even sure the example is necessary :) Let's keep it.

Lgtm!

PR #2307, opened 2016-08-26
Breaking: Allow to register multiple global handlers

zazabe:


zazabe:
@vogdb please review

vogdb:
looks good.

Off topic. Should we have a method to remove handler now when we can attach a multiple of them?

zazabe:
Few improvements + add `cm.error.unregisterHandler`, please review @vogdb

vogdb:
Why not use `reject` instead of `filter` with negation? I think you should delete this method at all. I just couldn't find any caveats in your blessed code but I so wanted to write something smart so I wrote about unregistering.

zazabe:
I did it with love (for you?), so please, don't ask me to remove it now...

vogdb:
petite lgtm ;-)

PR #2311, opened 2016-08-30
ESC key doesn't close floatbox

zazabe:
As reported by @NicolasSchmutz 

Probably a regression via https://github.com/cargomedia/cm/pull/2301 ?

If we can't find a quick fix, let's revert that change because we don't need it yet.


zazabe:

christopheschwyzer:
lgtm

PR #2310, opened 2016-08-30
Breaking: Mail: allow to send over SMTP

zazabe:
Currently `CM_Mail` uses *PHPMailer* in its default configuration, which uses PHP's `mail()` function to send the mail (which in turn uses the `sendmail` program).

In the future we would like to talk to an SMTP server directly from PHP for better error reporting, and easier configurability.

Ideas:
- Separate `CM_Mail` (the individual email, a view) from the sending/transport functionality
- Use an external library: [PHPMailer](https://github.com/PHPMailer/PHPMailer) or [Swift Mailer](https://github.com/swiftmailer/swiftmailer). The latter looks better?
- Move sending of emails to a gearman job, as it can take some time.
 - Drop queueing mechanism (mysql table `cm_mail`), just queue with gearman.
- Allow to configure different transports:
 - production: SMTP server ([current config](https://github.com/cargomedia/puppet-cargomedia/blob/master/modules/cargomedia/files/hiera-data/datacenter/softlayer.json#L232))
 - development: Currently just goes to the Log (see `CM_Mail::_log()`)


zazabe:
@njam please review

zazabe:
_note "delay" with gearman will be done in a follow up_

zazabe:
as discussed with @njam, gearman should be implemented in this PR to avoid troubles in prod (sending through SMTP might be slower than the current version).

@alexispeter could you take a look at this PR to ensure that i'm going in the right direction?

zazabe:
everything is implemented! @alexispeter review please

alexispeter:
I think you're mixing a few concepts. There's CM_Mail which is a representation of a mail but also has sending logic. Then there's CM_Mailer_Message which is injected into swiftmailer and actually holds the mail's data and also provides a some convenience methods for the cm-app.

I would do what @njam suggested and try to split the whole system into individual concepts.
* The mail-template (i.e the subclasses of CM_Mail)
* The mail-object that is created when we send/queue a mail (Any instance of Swift_Mime_SimpleMessage). 
* The mailer-client that sends mails (CM_Mailer_Client).

The mail-template would just be responsible for rendering title, html-body and text-body, nothing more. 
The mailer-client would provide methods to create and send mail-objects and also to create mail-objects from mail templates plus template-variables.

I think doing this would really simplify things.

zazabe:
The idea was to keep the same `CM_Mail` API, with all those `setSubject`, `setSender`, etc...  (oops, `setText()` is missing, used by SK in some places...)

Anyhow, I agree and I think it will not take too much time on CM, but it requires more refactoring on SK... wdyt @njam? should I remove `CM_Mailer_Message` from `CM_Mail`?

alexispeter:
The interfaces of CM_Mail and Swift_Mime_SimpleMessage are similar enough for the template-less usages of CM_Mail to not be too munch of a hassle. Template-based mails will be a bit trickier, since we will have to decouple the constructor arguments from the mail-template.

zazabe:
Currently CM_Mail does two jobs:
- Populating the **message** (from, to, subject)
- Rendering the **view** (body)

We discussed with njam that we will keep these two things in the same file for now and call it **mailable** (`CM_Mail_Mailable`). It will still inherit from the view. (In a later step we can remove that inheritance and *use* the View to render the body.)

Naming suggestion:
- CM_Mail_Mailable (extends `CM_View_Abstract`)
- CM_Mail_Mailer
- CM_Mail_MailerFactory
- CM_Mail_Message
- CM_Mail_Transport_Log
- CM_Mail_SendJob

alexispeter:
* library/CM/Mail.php
 + setRenderLayout() passing null should be equal to passing true instead of false because it is somewhat implied by the method's name. 
 + same for setVerificationRequired()
* library/CM/Mailer/Client.php
 + I think we should log a warning when any recipients failed, not just an exception when all recipients fail. Maybe we shouldn't even throw an exception but just log an error in this case?
 + registering CM_Mailer_Message should probably be done directly in the constructor.
* CM_Mailer_Factory
 + Setting the serviceManager is unnecessary for services, it's done implicitly when the manager instantiates a service.
* library/CM/Mailer/Job/Queue.php
 + ~~It's a bit weird that theoretically one might "prepare" a mail using one mailer but not be able to send it with that mailer (transport) because sending always uses the mailer that is configured on the service-manager. From an sk point of view it's not important because we only ever use one transport, but from a cm point of view, it would probably be good if the job somehow knew about what mailer to use exactly.~~ Tricky stuff, might be better to ignore this for now.
* library/CM/Mailer/Message.php
 + fromArray() A message can only have one sender 
 + fromArray() Use setFoo instead of adding properties one by one
* library/CM/Mailer/Transport/Log.php
 + Should probably have the logger injected. A technicality, feel free to ignore for now.
* tests/helpers/CMTest/library/CMTest/TestCase.php
 + getMockUser() I would remove this and implement it individual testscases as needed.

zazabe:
@alexispeter I did all changes from your comment, also now, `CM_Mail_MailerFactory` has specific methods for each Transport (only Log and Smtp for now), which is a bit more clear i think, anyhow, please review again... ;)

alexispeter:
* library/CM/Mail/Mailable.php
 + When using ternary operators, the confidion should be set up in such a way that the "default" value comes at the end.
* README.md
 + Text occurence of `CM_Mail`
* CM_Mail_MailerTest
 + testSendFailed() can be implemented, you just need to pass `failedRecipients` as reference. Since the transport's method is mocked, the value does not get changed.
```php
$failedRecipients = ['foo@example.com'];
$client->send($message, $failedRecipients);
```
* library/CM/Mail/MailerFactory.php
 + reverse ternary operator condition so the default is at the end.

zazabe:
@alexispeter please review

alexispeter:
lgtm

zazabe:
@njam just thought about that, now if [sending the mail for some recipients fails](https://github.com/zazabe/cm/blob/630f9e836b85db14407b04fde8bab52cfef2d9fb/library/CM/Mail/Mailer.php#L19), it will be logged as a warning, we may have lot of those warning if emails are not correct, wdyt?

alexispeter:
Most emails we send out require email-confirmation, so this should be ok.

njam:
I guess we will only have *immediate* failures, that our SMTP provider (ESP) detects.
For example "mailbox doesn't exist" will only be detected later?

Can you test it with our production config, to find out what errors we will see exactly? Then we can decide how to treat them.
https://github.com/cargomedia/puppet-cargomedia/blob/master/modules/cargomedia/files/hiera-data/datacenter/softlayer.json#L232

zazabe:
About errors (in `CM_Mail_Mailer::send()`):
- unresolvable host  
`Swift_TransportException: Connection could not be established with host foo.bar.com [php_network_getaddresses: getaddrinfo failed: No address associated with hostname #0]`
- wrong port   
`Swift_TransportException: Connection could not be established with host xxx [Connection timed out #110]`
- wrong recipient (due to `Swift_RfcComplianceException`, eg: `foo@bar`)  
  `CM_Mail_Mailer::send()` returns `0` and `$failedRecipients = ["foo@bar"]`
- inexistent recipient (eg: `foo@bar.com`)  
  works, `CM_Mail_Mailer::send()` returns `1` and `$failedRecipients = []`

_tested with https://0bin.net/paste/5+aTXR96sjH+I+Gj#3faFSaN6IBLMyrDHv3kfXw6xAEUSSboccBalRAU7Mno_

njam:
As discussed let's bind to the "throw" event nevertheless to get better visibility.
We can throw an ERROR exception for now, and it we see it's problem we can reconsider.

zazabe:
As checked, swift transport exception are thrown and will be caught by CM, no need to bind to the `exceptionThrown` event.

library/CM/Mail/Message.php
```diff
@@ -22,6 +22,11 @@ public function getText() {
         return $this->_text;
     }
 
+    public function setSender($address, $name = null) {
```
alexispeter: Stuff like this, where you explicitly change the behaviour of a method, should be done in CM_Mail_Mailable instead of CM_Mail_Message.

alexispeter:
rest lgtm

zazabe:
@njam could you review the last 2 commits?

njam:
lgtm

PR #2312, opened 2016-08-31
MongoDB based model constructor calls with invalid id

fvovan:
If we pass malformed `$id` to model's based on mongo constructor we get 
`MongoException`. 
It's not handly to catch both that and `CM_Exception_Nonexistent`
How it would be better?
- to validate it (and throw `CM_Exception_Nonexistent` ) here https://github.com/cargomedia/cm/blob/6cdeb6109ee2b848c47aca119dce8301988e2fca/library/CM/Model/StorageAdapter/MongoDb.php#L18
- to create abstract `CM_Model_MongoBased` (or something) like that, overriding the constructor and implementing some mongo specific method.  Maybe it even should be a trait...

@njam wdyt?

### DIFF
--- a/library/CM/Model/StorageAdapter/MongoDb.php
+++ b/library/CM/Model/StorageAdapter/MongoDb.php
@@ -15,7 +15,7 @@ class CM_Model_StorageAdapter_MongoDb extends CM_Model_StorageAdapter_AbstractAd
         $type = (int) $type;
         $id = (string) $id['id'];
         $collectionName = $this->_getCollectionName($type);
-        if (!$this->_checkIdFormat($id)) {
+        if (!$this->_isValidId($id)) {
             return false;
         }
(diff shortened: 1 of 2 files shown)


fvovan:
@alexispeter please review

library/CM/Model/StorageAdapter/MongoDb.php
```diff
@@ -73,6 +76,14 @@ public function delete($type, array $id) {
     }
 
     /**
+     * @param string $id
+     * @return bool
+     */
+    protected function _isValidId($id) {
+        return (bool) preg_match('/^[0-9a-fA-F]{24}$/', (string) $id);
```
alexispeter: Use MongoId::isValid() instead.

alexispeter:
lgtm

PR #2308, opened 2016-09-01
Use Logger instead of handleException()

fauvel:
We're often using `handleException()` when we want to log an exception using the following pattern:
```php
} catch (CM_Exception_Nonexistent $e) {
    $e->setSeverity(CM_Exception::WARN);
    CM_Bootloader::getInstance()->getExceptionHandler()->handleException($e);
}
```

Now that we have a proper Logger service we can instead just log a message:
```php
$serviceManager->getLogger()->warning('Something bad happened');
```

The difference will be:
- No stack trace available. Usually that's fine.
- Error will not be printed (stdout in CLI, and status 500 code in HTTP)
 - Printing happens only if the exception's severity is above the "printSeverityMin" (error or fatal). In most cases it's probably unexpected behaviour to print it (at least for HTTP).

Let's check all usages, also in SK and decide case to case.
Do we even still need `handleException()`?

cc @zazabe @fvovan

### DIFF
--- a/library/CM/Process/ForkHandler.php
+++ b/library/CM/Process/ForkHandler.php
@@ -76,7 +76,7 @@ class CM_Process_ForkHandler {
             $return = $workload($result);
             $result->setResult($return);
         } catch (Exception $e) {
-            CM_Service_Manager::getInstance()->getLogger()->logException($e);
+            CM_Service_Manager::getInstance()->getLogger()->logException($e, null, 'Cannot execute workload in forked child process');
             $result->setException($e);
         }
 
--- a/library/CM/Process/ForkHandler.php
+++ b/library/CM/Process/ForkHandler.php
@@ -76,7 +76,7 @@ class CM_Process_ForkHandler {
             $return = $workload($result);
             $result->setResult($return);
         } catch (Exception $e) {
-            CM_Bootloader::getInstance()->getExceptionHandler()->handleException($e);
+            CM_Service_Manager::getInstance()->getLogger()->logException($e);
             $result->setException($e);
         }
 
--- a/library/CM/Process/ForkHandler.php
+++ b/library/CM/Process/ForkHandler.php
@@ -76,7 +76,7 @@ class CM_Process_ForkHandler {
             $return = $workload($result);
             $result->setResult($return);
         } catch (Exception $e) {
-            CM_Service_Manager::getInstance()->getLogger()->logException($e, null, 'Cannot execute workload in forked child process');
+            CM_Service_Manager::getInstance()->getLogger()->logException($e);
             $result->setException($e);
         }
 
--- a/library/CM/Process/ForkHandler.php
+++ b/library/CM/Process/ForkHandler.php
@@ -76,7 +76,7 @@ class CM_Process_ForkHandler {
             $return = $workload($result);
             $result->setResult($return);
         } catch (Exception $e) {
-            CM_Service_Manager::getInstance()->getLogger()->logException($e);
+            CM_Service_Manager::getInstance()->getLogger()->addMessage('Forked workload failed', CM_Log_Logger::exceptionToLevel($e), (new CM_Log_Context())->setException($e));
             $result->setException($e);
         }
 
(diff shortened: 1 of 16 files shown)


fauvel:
@tomaszdurka please review:
- This change will disable printing errors, which means that we will either not print to the error stream (in cli mode) or not send the http header 500 (in html mode). Please check that this won't be an issue in the use cases I've refactored here.
- I'm relying on `CM_Service_Manager::getInstance()` when I'm not in the context of a service. Any suggestion on how to avoid this?

njam:
As discussed with Tomasz:
Currently with `logException()` we log the full exception (including the stack trace), but with a generic message.
Instead we suggest to use `warning()` and log a specific message about what went wrong, for example
- `CM/Action/Email.php`: it could be "Unrecognized mail type when creating mail action" (with the type in meta info)
- `CM/MessageStream/Adapter/SocketRedis.php`: "Cannot add subscriber when synchronizing status" (add the exception to the context, for debugging if needed)

fauvel:
@tomaszdurka @njam please review - agree on always using the exception message of `CM_Exception` in the logs instead of a generic "Application error"?

njam:
The idea was to drop `logException()`, use `warning()` directly and write are more specific message in the `catch` (not re-use the one from the original exception).

The reason is that for example in `Adapter/SocketRedis.php` it's not obvious what actually happened and for example why it's a warning and not an error. If we write that out in a new message it will be easier to understand ("Cannot add subscriber when synchronizing status" ). By including the original exception in the meta info we are still able to look through the stack trace if needed.

fauvel:
@njam please review.

I'm not dropping `logException` altogether, because I think it makes sense to use it in cases where we execute a callback and return an error code (forks, jobs, redis client, cli). We might actually write to the error log in this case.

I'm not using `CM_Log_Context::setExtra` to add meta information either, because it is already set on the exception passed to `CM_Log_Context::setException` and logged along with it.

Otherwise I'm using `CM_Log_Logger::warning` to log exceptions in the warning log.

library/CM/Process/ForkHandler.php
```diff
@@ -76,7 +76,7 @@ public function runAndSendWorkload() {
             $return = $workload($result);
             $result->setResult($return);
         } catch (Exception $e) {
-            CM_Service_Manager::getInstance()->getLogger()->logException($e, null, 'Cannot execute workload in forked child process');
+            CM_Service_Manager::getInstance()->getLogger()->logException($e);
```
tomaszdurka: Let's drop the logException method all together. Simple interface on logger and exceptions only as part of context seems more aligned. What about using `error` method here same as for warnings?

library/CM/Redis/Client.php
```diff
@@ -349,7 +349,7 @@ public function subscribe($channels, Closure $callback) {
                     $response = $callback($message->channel, $message->payload);
                 }
             } catch (Exception $e) {
-                $this->getServiceManager()->getLogger()->logException($e, null, 'Cannot execute callback for redis message');
+                $this->getServiceManager()->getLogger()->logException($e);
```
tomaszdurka: Same as above :)

fauvel:
> Let's drop the logException method all together. Simple interface on logger and exceptions only as part of context seems more aligned. What about using error method here same as for warnings?

@tomaszdurka why would you like to ignore the exception severity and put everything into the error log?

tomaszdurka:
Ok, I agree to extract level from exception - true to that. I am still sceptic for these shorthands like `logException`. Currently it is used in two places:
- logging exception with severity, imo here a clean usage for addMessage()
- logging exception with level of the exception

I would not create method for that, but just do it inline. Also having a custom message for it sounds fun! 🐼

fauvel:
@tomaszdurka please review - is this what you meant?

fauvel:
@tomaszdurka please review!

tomaszdurka:
that looks good to me!

fauvel:
Tagged 1.200.1

PR #2314, opened 2016-09-01
remove '{"message" : { $exists : true } }' clause from CM_Paging_Log

fvovan:
It's just useless and impacting performance.


fvovan:
@njam  please review

njam:
lgtm

PR #2315, opened 2016-09-05
Show seconds in smarty_function_date_range

dlondero:


dlondero:
@tomaszdurka please review. Combining `date` and `date_time` smarty plugins in order to be able to show seconds and pass a timezone when needed.

tests/library/CM/SmartyPlugins/function.dateRangeTest.php
```diff
@@ -40,8 +40,8 @@ public function testRender() {
             'params'   => ['start' => $start, 'showTime' => true],
             'expected' =>
                 (CMTest_TH::getVersionICU() < 50)
-                    ? '<span class="date-range">2/1/03 12:34 PM – now (<span class="date-period">' . $yearsAgo . ' years</span>)</span>'
-                    : '<span class="date-range">2/1/03, 12:34 PM – now (<span class="date-period">' . $yearsAgo . ' years</span>)</span>'
```
tomaszdurka: Why are we losing "PM"?

dlondero: Because `smarty_function_date_time` uses `H` to format hours (0-23).

dlondero:
@tomaszdurka ?

tomaszdurka:
Sorry was reviewing this already, bot distracted by something. Give me a minute.

tomaszdurka:
Well it works fine I agree.

Maybe it would be better to have some more flexible methods for the future with possibility to pass time/date-formats. We could allow to set default values for this plugins and set those up in the configuration of the framework.

```smarty
{date_time time=$time format='HH:ii:ss'}
```

```php
$smarty->getPlugin(function(SmartyPluginDateTime $plugin) {
  $plugin->setDefaultFormat('HH:ii:ss');
})
```
Above would require some changes in the plugins architecture we use.

Anyway let's say it's good!

PR #2319, opened 2016-09-05
ES location index gets emptied

fauvel:
The "location" search index suddenly doesn't return any results.
It happened once before, and now again as reported in https://github.com/cargomedia/sk-support/issues/29.

Querying the index' status shows 0 documents:
```
curl -XGET 'http://localhost:9200/location/_status'
[…]
      "docs": {
        "num_docs": 0,
        "max_doc": 0,
        "deleted_docs": 0
      },
[…]
```
Full output: https://gist.github.com/njam/c99a516a903122054fd66ae2d7a2bf1f

I guess it happens after the weekly MaxMind upgrade.

@fauvel can you investigate?
It looks to me like the previous call to `CM_Model_Location::createAggregation()` probably truncates the table `cm_tmp_location`, and when the search index is created the mysql slave doesn't have the table yet?

Didn't we want to re-create that table with a different name, and then just rename it?

### DIFF
--- a/library/CM/Db/Db.php
+++ b/library/CM/Db/Db.php
@@ -178,6 +178,19 @@ class CM_Db_Db extends CM_Class_Abstract {
     }
 
     /**
+     * @param string $table
+     * @param string $replacement
+     */
+    public static function replaceTable($table, $replacement) {
+        $client = self::getClient();
+        $tableQuoted = $client->quoteIdentifier($table);
+        $replacementQuoted = $client->quoteIdentifier($replacement);
+        $temporaryTableQuoted = $client->quoteIdentifier(uniqid($table . '_'));
+        self::exec("RENAME TABLE {$tableQuoted} TO {$temporaryTableQuoted}, {$replacementQuoted} TO {$tableQuoted}");
+        self::exec("DROP TABLE {$temporaryTableQuoted}");
+    }
+
+    /**
      * @param string            $table
      * @param string|array      $fields Column-name OR Column-names array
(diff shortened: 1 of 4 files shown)


fauvel:
@njam please review

fauvel:
@njam please review. Is it reasonable to think that at least `cm_tmp_location_new` will already have been created on the maintenance database, when the whole database update has just completed on the master? Or should I explicitly wait for the replication to have started in the first place?

fauvel:
I've just added a [commit](95dcd1e) to illustrate, we can revert it if you prefer.

fauvel:
I've just added [two commits](https://github.com/cargomedia/cm/compare/89b6377a481f8b6b53411836399b9f2be4c0cc4e...48655560ef7c583a25248347a4411e2d950d40b4) to illustrate, we can revert them if you prefer.

njam:
`$waitForDb` looks very weird (difficult to understand). Let's do it outside.

The `_old`-table is only needed because there's no atomic *table replacement* in mysql. I would therefore suggest to treat it as an "internal" thing and not expose it to much:
- Don't consider it in `*InProgress()`
- Introduce a `dbdb::replaceTable($current, $replacement)

fauvel:
@njam ok, please review!

njam:
- CMService/MaxMind.php: Make it service manager aware, instead of using singleton access?

Rest lgtm

fauvel:
@njam last review please!

njam:
lgtm

PR #2318, opened 2016-09-06
'Undefined index' in CM_Paging_ModelAbstract

fvovan:
There are still some problems  with model paging 

ErrorException: E_NOTICE: Undefined index: a:2:{s:2:"id";s:7:"5927208";s:4:"type";s:1:"1";} in /home/{appDir}/vendor/cargomedia/cm/library/CM/Paging/ModelAbstract.php on line 41

https://github.com/cargomedia/cm/pull/2295
https://github.com/cargomedia/cm/pull/2279

### DIFF
--- a/library/CM/Paging/ModelAbstract.php
+++ b/library/CM/Paging/ModelAbstract.php
@@ -38,10 +38,6 @@ abstract class CM_Paging_ModelAbstract extends CM_Paging_Abstract {
             $this->_populateModelList($this->_getItemsRaw());
         }
         $index = serialize($itemRaw);
-        $model = isset($this->_modelList[$index]) ? $this->_modelList[$index] : null;
-        if (null === $model) {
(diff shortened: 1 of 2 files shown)


fvovan:
@njam please review

njam:
Now $model will be undefined?
Also I think we *never* want to throw now, correct?
So rather `$model = isset($foo['bar']) ? $foo['bar'] : null;`?

fvovan:
We need to throw here.
This exception is awaited in child paging.

njam:
You mean `SK_Paging_Entity_Pinboard`? Doesn't this work equally well with `null`?:
```php
    protected function _processItem($item) {
        $entity = null;
        try {
            $entity = parent::_processItem($item);
        } catch (CM_Exception_Nonexistent $e) {
            $entity = null;
        }
```

njam:
As discussed returning `null` is fine here.

fvovan:
@njam please review

njam:
lgtm

PR #2322, opened 2016-09-06
Send mail via SMTP with concurrency support

zazabe:
`Swift_SmtpTransport` seems to have some concurrency issues...

Actually, the connection is [open with `stream_socket_client`](https://github.com/swiftmailer/swiftmailer/blob/5.x/lib/classes/Swift/Transport/StreamBuffer.php#L249) and those params:
```php
// https://github.com/swiftmailer/swiftmailer/blob/5.x/lib/classes/Swift/Transport/EsmtpTransport.php#L37
[
    'protocol'               => 'tcp',
    'host'                   => 'localhost',
    'port'                   => 25,
    'timeout'                => 30,
    'blocking'               => 1,
    'tls'                    => false,
    'type'                   => Swift_Transport_IoBuffer::TYPE_SOCKET,
    'stream_context_options' => [],
]
```

So the connection is blocking by default, but I could imagine that multiple connections from different threads to the same host, even in blocking mode, could produce some issues...

Similar issues have been reported:
```
- https://github.com/swiftmailer/swiftmailer/issues/490
- https://github.com/swiftmailer/swiftmailer/issues/788
```

### DIFF
--- a/library/CM/Mail/Mailer.php
+++ b/library/CM/Mail/Mailer.php
@@ -17,13 +17,11 @@ class CM_Mail_Mailer extends Swift_Mailer implements CM_Service_ManagerAwareInte
         }
 
         $numSent = 0;
-        $failedRecipients = null;
         $context = new CM_Log_Context();
         try {
             $numSent = parent::send($message, $failedRecipients);
         } catch (Exception $e) {
             $context->setException($e);
-            throw $e;
         }
 
         $this->getTransport()->stop();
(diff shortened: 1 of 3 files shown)


zazabe:
https://github.com/zazabe/cm/blob/8a027f4f3e9ce8975bbf2d545c17480ab9a3c5a0/library/CM/Mail/Mailer.php#L29 seems to actually fix the issue (I can't reproduce it anymore in local) and should be fine as long as gearman doesn't run multiple mail tasks at the same time...

wdyt @njam @kris-lab ?

njam:
> should be fine as long as gearman doesn't run multiple mail tasks at the same time

I *will* run 30 job workers. See Procfile:
```
jobs: ./bin/cm --forks=30 job-distribution start-worker
```

So you'll also have it locally.
If it solves it, sgtm.
Maybe add a comment above that line and reference this ticket?

zazabe:
@njam please review

njam:
lgtm

PR #2324, opened 2016-09-07
Disable xdebug on Travis to speed up builds

dlondero:


dlondero:
@njam wdyt?

njam:
Looks like it saves ~0.8 min, great!

Could you check it also in SK?

dlondero:
In the meantime is this good to go?

njam:
yes this lgtm

Issue #2329, opened 2016-09-12
Make sure IDs are strings/enums in Loggly


njam:
Currently IDs (user-id, showSession-id, viewSession-id) are showing up as *numbers* in Loggly, which makes it impossible to aggregate on them. It works fine for strings.

Suggestion:
- Always log PHP objects as Hashes. E.g. `user: {id: 123, username: "foo"}`
- In Loggly handler add special case for "id", and make it a string

Tested: it works fine if we send the (numerical) ID as a string to Loggly!

fvovan:
Maybe it would be better to convert all numeric values to strings then?

njam:
No, we actually have some numerical values like "FPS", which we want to average, sum up etc.

zazabe:
btw, i think it's a bit related to https://github.com/cargomedia/cm/pull/2330#issuecomment-248312432

njam:
@tomaszdurka could you look into this?

zazabe:
done by #2379

PR #2330, opened 2016-09-13
Auto encode DateTime values in logger metaInfo

zazabe:
Automatically encode certain object types in "metaInfo" as strings in the Logger.
Any `DateTime` should be encoded as `format('c')`.

### DIFF
--- a/library/CM/Log/Context.php
+++ b/library/CM/Log/Context.php
@@ -99,15 +99,8 @@ class CM_Log_Context {
     /**
      * @param array $extra
      * @return $this
-     * @throws CM_Exception_Invalid
      */
     public function setExtra(array $extra) {
-        $flattenExtra = Functional\flatten($extra);
-        if (Functional\some($flattenExtra, function ($el) {
-            return is_object($el);
-        })) {
-            throw new CM_Exception_Invalid('Object can not be passed to "Extra"');
-        }
         $this->_extra = $extra;
         return $this;
     }
--- a/library/CM/Log/ContextFormatter/Cargomedia.php
+++ b/library/CM/Log/ContextFormatter/Cargomedia.php
@@ -12,19 +12,6 @@ class CM_Log_ContextFormatter_Cargomedia implements CM_Log_ContextFormatter_Inte
         $this->_appName = (string) $appName;
     }
 
-    public function formatRecordContext(CM_Log_Record $record) {
-        $levelsMapping = array_flip(CM_Log_Logger::getLevels());
-        $context = $record->getContext();
-
-        $result = [
-            'message'   => (string) $record->getMessage(),
-            'level'     => strtolower($levelsMapping[$record->getLevel()]),
-            'timestamp' => $record->getCreatedAt()->format(DateTime::ISO8601),
-        ];
-        $result = array_merge($result, $this->formatContext($context));
-        return $result;
-    }
-
     public function formatContext(CM_Log_Context $context) {
         $result = [];
         if ($computerInfo = $context->getComputerInfo()) {
--- a/library/CM/Log/ContextFormatter/Cargomedia.php
+++ b/library/CM/Log/ContextFormatter/Cargomedia.php
@@ -81,7 +81,7 @@ class CM_Log_ContextFormatter_Cargomedia implements CM_Log_ContextFormatter_Inte
 
     public function formatAppContext(CM_Log_Context $context) {
         $result = [];
-        $appAttributes = $context->getExtra();
+        $appAttributes = $this->_encodeExtra($context->getExtra());
         if ($user = $context->getUser()) {
             $appAttributes['user'] = $user->getId();
         }
@@ -91,6 +91,14 @@ class CM_Log_ContextFormatter_Cargomedia implements CM_Log_ContextFormatter_Inte
         }
         $result[$this->_appName] = $appAttributes;
         return $result;
+    }
+
+    /**
+     * @param array $extra
+     * @return array
+     */
+    protected function _encodeExtra(array $extra) {
+        return $extra;
     }
 
     /**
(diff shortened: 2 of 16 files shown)


zazabe:
@tomaszdurka please review

zazabe:
@tomaszdurka can you check it? https://github.com/cargomedia/sk/pull/3777 is waiting for it

library/CM/Log/ContextFormatter/Cargomedia.php
```diff
@@ -97,6 +97,21 @@ public function formatAppContext(CM_Log_Context $context) {
      * @param array $data
      * @return array
      */
+    protected function _encode(array $data) {
+        foreach ($data as $key => $value) {
+            if (is_array($value)) {
+                $data[$key] = $this->_encode($value);
+            } elseif ($value instanceof DateTime) {
```
tomaszdurka: I understand this is one of the special cases. We might have more of them in the future. I would vote for using some kind of interface for this. Maybe simply `JSONSerializable`?

We could also then allow all objects with that interface.

@zazabe wdyt?

library/CM/Log/Context.php
```diff
@@ -101,7 +101,7 @@ public function setException($exception) {
     public function setExtra(array $extra) {
         $flattenExtra = Functional\flatten($extra);
         if (Functional\some($flattenExtra, function ($el) {
-            return is_object($el);
+            return !($el instanceof DateTime) && is_object($el);
```
tomaszdurka: Hacky!

Disallow all not implementing JSONSerializable?

tomaszdurka: I guess the only problem is that DateTime might implement this interface differently or not at all... :/

What about Params::encode or something like that?

zazabe: `json_encode()` is working on a `DateTime`, but it's not implementing `JsonSerializable`...

library/CM/Log/ContextFormatter/Cargomedia.php
zazabe: The interface will not work here i guess, because of your next comment... maybe we can change `_extra`  by a `ArrayObject` instance,  with some extra logic to validate data recursively and serialize them.

zazabe:
@tomaszdurka please review

library/CM/Log/ContextFormatter/Cargomedia.php
tomaszdurka: I think we had object representation of the Extra once, but we removed that and kept simple array.

zazabe:
I have the feeling that some refactoring is needed to solve properly this issue...
Why not reverting this PR to 3a4a853e11c9a8994025e1a5b903db226311a705 (the hacky solution) and then open a new issue to solve this properly?

wdyt @fvovan @tomaszdurka @njam ?

Some thoughts about the logger in general:
- always use CM_Log_Handler_Factory to create handlers (mongoDb?, NewRelic?)
- all handlers in Log/Context/Handler (CMService_NewRelic_Log_Handler?)
- all handlers have a record formatter, and then record formatter have a context formatter
- CM_Log_ContextFormatter_Interface ?? those 3 methods are strange, `formatRecordContext(CM_Log_Record $record)`? so the context formatter can format a record... not sure to get it...

njam:
I'll leave this to you @tomaszdurka and @fvovan as you are the experts here

fvovan:
>always use CM_Log_Handler_Factory to create handlers (mongoDb?, NewRelic?)

as for me it's not really important thing. if we notice we need to make some additional actions before the constructor call, we will always can move initialization to factory.

>all handlers in Log/Context/Handler (CMService_NewRelic_Log_Handler?)

agree.

>all handlers have a record formatter, and then record formatter have a context formatter

I think it should be not a part of contract. `CM_Log_Handler_HandlerInterface` is pretty ok now. May be encoding of certains classes could be done as kind of "record preprocessing" in `CM_Log_Handler_Abstract::handleRecord`. I would definitely not create separate class for `Extra`

>`CM_Log_ContextFormatter_Interface` ?? those 3 methods are strange, formatRecordContext(CM_Log_Record $record)? so the context formatter can format a record... not sure to get it.

hmm. Indeed. Context formatter should not do it.

tomaszdurka:
As discussed on hipchat, we've agreed to:

- keep context extra as simple array
- do not control its values when storing them initially
- encode/filter out specific types of values (DateTime, object) in specific handlers only (Mongodb, Loggly)

zazabe:
so:
- remove the `setExtra()` check (what about potential huge objects? see https://github.com/cargomedia/cm/pull/2236#issuecomment-229679859)
- encode in specific handlers 
  - fluentd: format DateTime / drop other objects? 
  - mongodb: use MongoDate for Datetime? / allow jsonSerializable objects only?

tomaszdurka:
@zazabe Sounds good.

I would worry about huge objects when dumping them in handler. If we pass it to context in php it's trivial reference.

zazabe:
- introduce `CM_Log_ContextFormatter_MongoDb`
  - use `MongoDate` for `DateTime`
  - format objects in `[<classname>]` / cm models  `[<classname>:<id>]`
  - use `jsonSerialize()` on  `JsonSerializable` objects + recursive formatting 
- introduce `CM_Log_ContextFormatter_Fluentd`
  - use `format('c')` for `DateTime`
  - format objects in `[<classname>]` / cm models  `[<classname>:<id>]` 
  - `JsonSerializable` objects are also formatted as `[<classname>]`

maybe  `[<classname>:<id>]` formatting is enough to solve https://github.com/cargomedia/cm/issues/2329...

wdyt @fvovan @tomaszdurka ?

tomaszdurka:
should I take over here?

fvovan:
it's up to you. actually i haven't managed to make any changes yet. but was going to start in the evening.

tomaszdurka:
Closing in favor of #2368

PR #2260, opened 2016-09-14
Fix invalid cache for subscriber-paging in onSubscribe-hook

alexispeter:
CM_Paging_User_StreamChannelSubscriber and other similar pagings have an invalid state during onSubscibe()/onPublish() hooks, because they are called before the automatic cache invalidation mechanism kicks in.

### DIFF
--- a/library/CM/Model/Stream/Publish.php
+++ b/library/CM/Model/Stream/Publish.php
@@ -23,6 +23,10 @@ class CM_Model_Stream_Publish extends CM_Model_Stream_Abstract {
         return CM_Db_Db::select('cm_stream_publish', '*', array('id' => $this->getId()))->fetch();
     }
 
+    protected function _onCreate() {
+        $this->getStreamChannel()->onPublish($this);
+    }
+
     protected function _onDelete() {
         CM_Db_Db::delete('cm_stream_publish', array('id' => $this->getId()));
     }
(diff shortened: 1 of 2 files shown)


alexispeter:
please review @tomaszdurka

library/CM/Model/Stream/Publish.php
```diff
@@ -72,7 +76,6 @@ protected static function _createStatic(array $data) {
             'channelId'    => $streamChannel->getId(),
         ));
         $streamPublish = new self($id);
-        $streamChannel->onPublish($streamPublish);
```
tomaszdurka: Im wondering. Wasn't there a reason to actually call this method onDelete? I presume it could remove outdated (including this deleted) streams from channel. Wdyt?

tomaszdurka: Wows, ignore this one. I misread the code in diff mode.

So actual change moves the onPublish call after `_onChange` and `_changeContainingCacheables`. Can you explain a bit what is intention here. Im not so into the specifics.

alexispeter: This change is about the following four pagings.
* CM_Paging_StreamPublish_StreamChannel
* CM_Paging_User_StreamChannelPublisher
* CM_Paging_StreamSubscribe_StreamChannel
* CM_Paging_User_StreamChannelSubscriber
The content of those pagings changes when a subscriber or publisher is created (db-insert in `_createStatic()`), but the cache of these pagings is only invalidated AFTER `_createStatic()`, when `_changeContainingCacheables()` is called.
That means that during any onSubscribe()/onPublish()-hooks, those pagings do not contain the viewer/subscriber/publisher that was just added to the streamchannel which can lead to bugs in streamchannel-implementations.

Issue #2327, opened 2016-09-09
Improve cross-layout pageload experience


njam:
Problem: currently when loading a page via Ajax (`CM_Http_Response_View_Abstract::loadPage()`) that has a different layout then the current page we only detect that on the client (looking at `layoutClass`), and then request a reload of the full document. So we have 2 requests, and load the whole page content twice.

Possible solutions:
- A: **Detect the layout change on the server in `loadPage()`**. And only return a flag to the client. We'll still have 2 requests, but transmit the content only once. Easy to implement.
- B: **Keep a page->layout map on the client**. The client receives or builds a map over time of what page uses what layout. That map can change with every release. When loading a page we look up what the layout will be and act appropriately. Will require storing/requesting that map, but easy to implement
- C: **Replace the whole layout dynamically**. Instead of only replacing *pages* on the client allow to replace full *layouts*. There's no good reason why it should be different. Most difficult, but most efficient.

@zazabe @vogdb @tomaszdurka wdyt?

vogdb:
I would do A now if it's urgent and do C later.

njam:
Agree.
@fauvel wanna implement A as a first step?

fauvel:
👍

fauvel:
I'm implementing A in #2333

njam:
As discussed with @fauvel @vogdb @tomaszdurka - new idea:
- Introduce "Document" view, which contains the HTML template. Only 1 template per namespace/site.
- Make "Layout" only responsible for *visible* template (`#body-container`)
- In `loadPage` determine if the next page uses the same or a different layout and return either the "pageHtml" or "layoutHtml".
 - Remove Embed Page response?
 - Move some code out of the Page response into a new "processor"?

Let's think about it a bit more and decide in 1-2 days.

PR #2333, opened 2016-09-15
Detect page layout changes on the server side

fauvel:


fauvel:
@tomaszdurka please review

fauvel:
@njam it is very complicated to figure out cleaner alternatives to this approach, as the code path is very intricate here. @tomaszdurka's just given up for today :) What's your opinion? If you agree with my proposal I would add a unit test and deploy as is.

njam:
As discussed let's close this for the moment, as we discuss a "broader" solution in https://github.com/cargomedia/cm/issues/2327.
It's not worth testing and thinking through this solution in much details, because the context might change with https://github.com/cargomedia/cm/issues/2327.

PR #2335, opened 2016-09-15
Pass offset instead of timeZoneName to IntlDateFormatter

dlondero:
Addressing  https://github.com/cargomedia/sk/issues/3861


dlondero:
@njam please review

library/CM/Model/Location.php
```diff
@@ -125,6 +125,9 @@ public function getTimeZone() {
             return null;
         }
         $timezoneNameList = DateTimeZone::listIdentifiers();
+        $timezoneNameList = \Functional\select($timezoneNameList, function ($timeZone) {
```
njam: How about making this more explicit (and not depend on an implicit bool cast)?
```
reject($list, fn($tz) { return null === foo($tz); })
```

dlondero: Agree.

dlondero:
@njam please check last commit

njam:
lgtm

PR #2344, opened 2016-09-16
Follow-up #2315

dlondero:


dlondero:
@njam please review

njam:
I would not use the comma, I think in some languages this might be wrong (arabic? chinese?). Space is probably safer?

dlondero:
@njam agree, please check

dlondero:
Ok now it's good to review.

njam:
lgtm

PR #2346, opened 2016-09-16
Upgrade to PHPUnit 5.5

dlondero:
Now that we have PHP 5.6 we can finally upgrade to PHPUnit 5.5.

### DIFF
--- a/tests/helpers/CMTest/library/CMTest/TestCase.php
+++ b/tests/helpers/CMTest/library/CMTest/TestCase.php
@@ -39,35 +39,6 @@ abstract class CMTest_TestCase extends PHPUnit_Framework_TestCase implements CM_
     }
 
     /**
-     * @param string      $originalClassName
-     * @param array       $methods
-     * @param array       $arguments
-     * @param string      $mockClassName
-     * @param bool        $callOriginalConstructor
-     * @param bool        $callOriginalClone
-     * @param bool        $callAutoload
-     * @param bool        $cloneArguments
-     * @param bool        $callOriginalMethods
-     * @param object|null $proxyTarget
-     * @return PHPUnit_Framework_MockObject_MockObject
-     */
-    public function getMock($originalClassName, $methods = [], array $arguments = [], $mockClassName = '', $callOriginalConstructor = true, $callOriginalClone = true, $callAutoload = true, $cloneArguments = false, $callOriginalMethods = false, $proxyTarget = null) {
-        $mockBuilder = $this->getMockBuilder($originalClassName)
-            ->setMethods($methods)
-            ->setConstructorArgs($arguments)
-            ->setMockClassName($mockClassName)
-            ->setProxyTarget($proxyTarget);
-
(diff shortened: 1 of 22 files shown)


dlondero:
@njam please review

tests/helpers/CMTest/library/CMTest/TestCase.php
```diff
@@ -39,10 +39,39 @@ public function callProtectedMethod($object, $methodName, array $arguments = nul
...
+     * @param string      $mockClassName
+     * @param bool        $callOriginalConstructor
+     * @param bool        $callOriginalClone
+     * @param bool        $callAutoload
+     * @param bool        $cloneArguments
+     * @param bool        $callOriginalMethods
+     * @param object|null $proxyTarget
+     * @return PHPUnit_Framework_MockObject_MockObject
+     */
+    public function getMock($originalClassName, $methods = [], array $arguments = [], $mockClassName = '', $callOriginalConstructor = true, $callOriginalClone = true, $callAutoload = true, $cloneArguments = false, $callOriginalMethods = false, $proxyTarget = null) {
```
njam: What is this about? Backwards compatibility?

dlondero: Kinda. The real `getMock` is deprecated and we should use `createMock` or the `mockBuilder` instead. With both is not possible to pass all params at once so changes would be in quite a few places.

njam: I see. Probably to avoid having so many arguments?
How about we add a `@deprecated` annotation to our implementation then, so developers don't use it?

dlondero: Either adding `@deprecated` or going through usages and replace them already so we get rid of `getMock`. Which one?

njam: Adding the annotation should be fine imo

tomaszdurka: Mhmm, so we adding deprecated method to replace phpunit's deprecated method? :)

njam: The difference is that PHPUnit writes a warning to STDERR - not so nice.

dlondero: I guess @tomaszdurka would prefer the second option for this PR: replace usages of `getMock` w/o having this new deprecated method.

njam: That's also good.
I thought they will vanish "over time" anyway.
Up to you!

tomaszdurka: Ah ofc. Marking method as deprecated with docblock is something much different then triggering `E_USER_DEPRECATED`...

dlondero:
@njam please review. Using `mockBuilder` so we don't need to do it in the future when `getMock` will be removed.

njam:
lgtm

PR #2348, opened 2016-09-18
add page path to manifest's "start_url"

njam:


njam:
Needed so that the manifest start URL is always covered by the service worker scope.

PR #2332, opened 2016-09-20
Migrate PersistentStorage in CM

zazabe:
Use `PersistentStorage` for `CM_App.storage`

PR #2352, opened 2016-09-21
PersistentStorage: read/write on every access

zazabe:
Followup of https://github.com/cargomedia/cm/pull/2332

Problem: LocalStorage can be accessed by multiple tabs, so we cannot cache `_data` locally.
Idea: Always read/write from local storage. If not available instead install a "Object" adapter.

cc @vogdb


zazabe:
@vogdb please review

zazabe:
Ooops, sorry, should be fixed and improved ;) please review @vogdb

Issue #2320, opened 2016-09-06
Ensure mongodb client will fallback to different server/router


njam:
We pass a list of MongoDB routers to the mongodb client:

``` php
    $config->services['MongoDb'] = [
        'class'     => 'CM_MongoDb_Client',
        'arguments' => [
            'config' => [
                'db'      => 'example',
                'server'  => 'mongodb://localhost:27017,www3.example.com:27017,www4.example.com:27017,search2.example.com:27017',
                'options' => ['connect' => true, 'username' => 'example', 'password' => 'example', 'db' => 'example'],
            ],
        ],
    ];
```

Just today we experienced that one of these routers was down, and as a result many web processes' mongodb queries timed out after 60 seconds.

Let's investigate how the PHP client behaves if 1 router is down.
How can we minimize the impact, so that the client will fall back to a different router as soon as possible.

@alexispeter @fvovan @zazabe maybe for you?

fvovan:
I didn't try to reproduce anything but just read the docs yet.
 First of all I see that `replicaSet` option is not used, and according documentation it means 

> Automatic failover will not work correctly if replicaSet is not specified.

So it looks like it just checks server status for the moment of initial connection and doesn't care about reconnections after its crashing.

But even if this param is set and  primary node is down then some delay on write operations will occur.

>During this time (» 20-60 seconds), the connection will not be able to perform any write operations and attempts to do so will result in an exception. Connections to secondaries will still be able to perform reads.

The  way to decrease this interval is (as usual :) ) to upgrade mongoDb to 3.2.  It introduces specific settings for election.

> For instance, it may take 10-30 seconds for the members of a replica set to declare a primary inaccessible (see electionTimeoutMillis). One of the remaining secondaries holds an election to elect itself as a new primary. The election itself may take another 10-30 seconds.
Changed in version 3.2: Starting in MongoDB 3.2, with the replication election enhancements, MongoDB reduces replica set failover time. See replication election enhancements for details.

And the last note is `MongoClient` is deprecated now. It might be we would like to replace it later?

@njam wdyt?

njam:
I see.
Can we upgrade to the new mongodb PHP driver http://php.net/manual/en/set.mongodb.php while we're still on Mongodb 2.6? If yes, then let's do it, to not spend time on an outdated version.

alexispeter:
We can.
 https://docs.mongodb.com/ecosystem/drivers/php/#drivers

PR #2331, opened 2016-09-23
Upgrade mongodb client/driver

alexispeter:
Old: http://php.net/manual/en/book.mongo.php
New: http://php.net/manual/en/set.mongodb.php

Compatibility:
https://docs.mongodb.com/ecosystem/drivers/php/#mongodb-compatibility

Let's also check https://github.com/cargomedia/cm/issues/2320


alexispeter:
Detailed Upgrade Guide: https://github.com/mongodb/mongo-php-library/blob/v1.0/docs/upgrade-guide.md

Changes to the interface of CM_MongoDb_Client
#### deleteIndex
Now uses the index' name instead of it's specification.
```php
// legacy
$client->deleteIndex(['foo' => -1]);

// new
$client->deleteIndex('foo_-1');
```
#### find
* Now always returns a `MongoDB\Driver\Cursor` instead of using different cursors for aggregated and non-aggregated operations.
* The cursor is traversable but does not provide an iterator-interface anymore. 
* Secondary search-attributes like `skip`, `limit` are no longer called on the cursor, but have to be passed as query-options, the new fourth argument.
```php
// legacy
$cursor = $client->find($collection, $criteria, $projection);
$cursor->sort(['foo' => -1]);
// consume cursor

// new
$cursor = $client->find($collection, $criteria, $projection, null, ['sort' => ['foo' => -1]]);
// consume cursor
```

(WIP)

alexispeter:
please review @tomaszdurka

njam:
@tomaszdurka please review

tests/library/CM/Model/StorageAdapter/MongoDbTest.php
```diff
@@ -43,8 +43,9 @@ public function testLoad() {
     public function testLoad_nonExistent() {
         $type = 99;
         $adapter = $this->_getAdapter();
+        $mongoDb = CM_Service_Manager::getInstance()->getMongoDb();
 
-        $this->assertFalse($adapter->load($type, ['id' => (string) new MongoId()]));
+        $this->assertFalse($adapter->load($type, ['id' => $mongoDb->getObjectId()]));
```
tomaszdurka: I would rather vote for having static methods for this ? Constructor is static method in the end, so maybe if we want to overwrite it we should use static method signature? On the other hand it would make sense if we would declare it on the Mongo class.
Not sure.

tests/library/CM/Redis/ClientTest.php
```diff
@@ -2,6 +2,10 @@
 
 class CM_Redis_ClientTest extends CMTest_TestCase {
 
+    public static function tearDownAfterClass() {
+        // To avoid issues that arise when forking a process with an established mongodb-connection. see https://jira.mongodb.org/browse/PHPC-433
```
tomaszdurka: What is the default behaviour of teardown here?

alexispeter: Teardown will reload `CM_MongoDb_SetupScript`, which will cause the mongo driver to explode because it was forked.

tomaszdurka: But shouldn't we fork only with service manager as container for mongo instances and if so reset it before? Can't recall how it exactly works.

alexispeter:
Some methods were renamed because they were split into multiple methods and it was decided to use the names of the driver's methods.

To try and keep the old name, we could do the following changes.
##### findAndModify
* Split into `findAndModify()` update-functionality, `findAndReplace()` replace-functionality, `findAndDelete()` delete-functionality
##### update
* Split into `update()` multi-update and `replace()`, don't provide `updateOne()`-functionality
##### remove
* Keep name for multi-delete, don't provide `deleteOne()`-functionality

About internal classes. I don't think we can get rid of `MongoDB\BSON\ObjectID` or `MongoDB\BSON\UTCDateTime`, but maybe we could introduce some sort wrapper for the cursors returned by `find()`.

wdyt @tomaszdurka

tests/library/CM/Redis/ClientTest.php
alexispeter: It's like this. The mongoClient internally sets up a Manager-object as a connection is needed. As soon as this object has been created you cannot fork or at least you cannot fork and then perform mongo operations without getting a million errors.

When running a test, CM_MongoDb_SetupScript::_isLoaded() is run as part of the setup-routine, which initiates a connection. After running a test CM_MongoDb_SetupScript::reload() is run, which will fail if the process has been forked in the meantime.

tomaszdurka:
I thought about something similar, but then when thought why they did it this way and not with some changeable iterators? So we would need to implement this logic.

You mean something like this?
```php

$statement = $mongo->find('collection');
$statement->setSort('name');
$iterator = $statement->execute();
foreach ($iterator as $item) {
}
// or
$iterator->getAll();
```

alexispeter:
The reason they did it like this and also the reason for all the changes in the interface is so the php client would be more in line with other clients, the legacy client was not an official driver.

What we need is an object that allows us to 
1. Traverse the cursor using mongodb's chunking functionality.
2. Return the whole resultset as array.

But I'm not sure how to satisfy point 1. The `MongoDB\Cursor` class is a built-in class that implements Traversable without using an iterator, which is not possible in userland php.

Or we just do without this functionality which is probably not required outside of update-scripts. It's often better to use manual chunking with mongodb. In this case I would change `MongoClient::find()` to `MongoDB\Cursor::toArray()`.

library/CM/MongoDb/Client.php
```diff
@@ -266,13 +313,11 @@ public function rename($collectionSource, $collectionTarget, $dropTarget = null)
                 'collectionTarget' => $collectionTarget,
             ]);
         }
-        $result = $this->_getClient()->selectDB('admin')->command([
+        $this->_getClient()->selectDatabase('admin')->command([
             'renameCollection' => $this->_getDatabaseName() . '.' . $collectionSource,
             'to'               => $this->_getDatabaseName() . '.' . $collectionTarget,
             'dropTarget'       => $dropTarget,
         ]);
-        $this->_checkResultForErrors($result);
```
alexispeter: All possible failures throw exceptions.

library/CM/Model/StorageAdapter/MongoDb.php
```diff
@@ -71,8 +72,8 @@ public function delete($type, array $id) {
         $type = (int) $type;
         $id = (string) $id['id'];
         $collectionName = $this->_getCollectionName($type);
-        $mongoId = new MongoId($id);
-        $this->_getMongoDb()->remove($collectionName, ['_id' => $mongoId]);
+        $mongoDb = $this->_getMongoDb();
+        $mongoDb->deleteOne($collectionName, ['_id' => $mongoDb->getObjectId($id)]);
```
alexispeter: Agreed to keep them as is, to try and conform to official drivers.

alexispeter:
please review @tomaszdurka

tomaszdurka:
lgtm

PR #2337, opened 2016-09-27
Fix CommandManager "locking/synchronizing" mechanism

njam:
`CM_Cli_CommandManager` has a mechanism to ensure a CLI command that defines the `@synchronized` annotation only runs once across the installation of the application.

See `monitorSynchronizedCommands`, `_lockCommand` and `_checkLock`.

The mechanism relies on a unique host identifier, by calling `hostid`, unfortunately that identifier is not unique across our hosts.

Let's investigate why `hostid` is not unique and/or find another mechanism.

Another improvement: in the foreach of `monitorSynchronizedCommands` let's only modify rows of the *current* host.

### DIFF
--- a/library/CM/Cli/CommandManager.php
+++ b/library/CM/Cli/CommandManager.php
@@ -334,7 +334,19 @@ class CM_Cli_CommandManager {
      * @return string
      */
     protected function _getMachineId() {
-        $file = new CM_File('/etc/machine-id'); // https://www.freedesktop.org/software/systemd/man/machine-id.html
+        // Global machine-id from systemd https://www.freedesktop.org/software/systemd/man/machine-id.html
+        $file = new CM_File('/etc/machine-id');
+
+        if (!$file->exists()) {
+            // Local machine-id as a backup
+            $serviceManager = CM_Service_Manager::getInstance();
+            $file = new CM_File('machine-id', $serviceManager->getFilesystems()->getData());
+            if (!$file->exists()) {
+                $uuid = Ramsey\Uuid\Uuid::uuid4()->toString();
+                $file->write($uuid);
+            }
+        }
+
         return trim($file->read());
     }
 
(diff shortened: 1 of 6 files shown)


njam:
@tomaszdurka please review

njam:
@tomaszdurka please review

library/CM/Cli/CommandManager.php
```diff
@@ -334,7 +334,19 @@ protected function _outputError($message) {
...
-        $file = new CM_File('/etc/machine-id'); // https://www.freedesktop.org/software/systemd/man/machine-id.html
+        // Global machine-id from systemd https://www.freedesktop.org/software/systemd/man/machine-id.html
+        $file = new CM_File('/etc/machine-id');
+
+        if (!$file->exists()) {
+            // Local machine-id as a backup
+            $serviceManager = CM_Service_Manager::getInstance();
+            $file = new CM_File('machine-id', $serviceManager->getFilesystems()->getData());
+            if (!$file->exists()) {
+                $uuid = Ramsey\Uuid\Uuid::uuid4()->toString();
```
tomaszdurka: Well well if we create it like this, maybe we should not rely on the global id at all.

resources/config/deploy.php
```diff
@@ -1,4 +1,4 @@
 <?php
 return function (CM_Config_Node $config) {
-    $config->deployVersion = 1;
+    $config->deployVersion = 1474973160;
```
tomaszdurka: Intended?

library/CM/Cli/CommandManager.php
njam: The global id will be more reliable so I would still use it. In case of app re-installs etc?

njam:
@tomaszdurka please review

resources/config/deploy.php
tomaszdurka: Is this intended? :)

njam: It's an outdated diff. The reviews and the attached code can be a bit confusing :/ Check the PR's diff.

njam:
@tomaszdurka please review

PR #2357, opened 2016-09-27
Exceptions not displayed in CLI

alexispeter:
For example: running `bin/cm app setup --reload` would throw an error/exception:
> Argument 1 passed to Denkmal_Model_Venue::setFacebookPage() must be an instance of Denkmal_Model_FacebookPage, null given, called in /home/vagrant/denkmal/library/Denkmal/library/Denkmal/Model/Venue.php on line 292 and defined

But no error is displayed on the standard output.
Exit code is 1 though.

@fvovan @tomaszdurka can you look into this?

### DIFF
--- a/library/CM/Process.php
+++ b/library/CM/Process.php
@@ -228,7 +228,7 @@ class CM_Process {
                 $forkHandler->runAndSendWorkload();
                 $forkHandler->closeIpcStream();
             } catch (Exception $e) {
-                CM_Service_Manager::getInstance()->getLogger()->addMessage('Forking workload failed', CM_Log_Logger::exceptionToLevel($e), (new CM_Log_Context())->setException($e));
+                CM_Bootloader::getInstance()->getExceptionHandler()->handleException($e);
             }
             exit;
         }
--- a/library/CM/Process/ForkHandler.php
+++ b/library/CM/Process/ForkHandler.php
@@ -76,7 +76,7 @@ class CM_Process_ForkHandler {
             $return = $workload($result);
             $result->setResult($return);
         } catch (Exception $e) {
-            CM_Service_Manager::getInstance()->getLogger()->addMessage('Forked workload failed', CM_Log_Logger::exceptionToLevel($e), (new CM_Log_Context())->setException($e));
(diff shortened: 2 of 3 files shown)


alexispeter:
please review @fvovan

alexispeter:
I'm wondering if it really is a good idea to skip the exception handler. I think it's ok on the application-level, but it can get messy if done on the framework-level. If the only reason for it is to provide a clearer message, then we should find a better way.

fvovan:
wdyt, should this be fixed as well (maybe it's needed for displaying of system problems like "failed to close socket" etc) ?
https://github.com/cargomedia/cm/pull/2308/files#diff-ce44370f9548cf834f2bb08c0ecf64e6R231


>I'm wondering if it really is a good idea to skip the exception handler. I think it's ok on the application-level, but it can get messy if done on the framework-level. If the only reason for it is to provide a clearer message, then we should find a better way.

Most of these cases are really were the places where we called `handleException` _only_ to log it.
Another point was relying more on `ServiceManager` but not on `Bootloader`

alexispeter:
I'm just saying that whether you want to just log an exception or also print it, depends more on the context in which it is run as on the semantics of the code. Sure there are cases where just logging is fine, especially if you're catching specific exceptions. But anything that catches all `CM_Exceptions`s or even every `Exception`whatsoever should be thought through twice. 
By the way, this is not the first time we have had to do this fix, see https://github.com/cargomedia/cm/pull/1449.

fvovan:
>I'm just saying that whether you want to just log an exception or also print it, depends more on the context in which it is run as on the semantics of the code. Sure there are cases where just logging is fine, especially if you're catching specific exceptions. But anything that catches all CM_Exceptionss or even every Exception whatsoever should be thought through twice. 

I agree, and all remaining places handling general exceptions  that were changed in https://github.com/cargomedia/cm/pull/2308 seem to be okay with it, because it's mostly background processing where there is no much sense in printing them to STDERR.

alexispeter:
please review @fvovan

fvovan:
lgtm

PR #2359, opened 2016-09-28
make StreamChannel::_decryptKey() usable in a static context

alexispeter:
So it can be used to decrypt encoded keys when creating a streamchannel and other stuff.


alexispeter:
please review @dlondero

alexispeter:
please review @dlondero

alexispeter:
please review @dlondero

PR #2355, opened 2016-09-28
Introduce "Document" view

njam:
Part of https://github.com/cargomedia/cm/issues/2327

Responsible for rendering the base HTML document.
Will receive a *Layout* (which has a *Page*), so meta titles, description etc can be determined.

- Introduce "Document" view, which contains the HTML template. Only 1 template per namespace/site.
- Make "Layout" only responsible for *visible* template (`#body-container`)

TODO:
- ~~Check for where JS assumes Layout is the *root* view~~


njam:

@fauvel could you test it thoroughly in combination with https://github.com/cargomedia/sk/pull/3941? Please verify also tracking!

library/CM/Http/Response/Page/Embed.php
```diff
@@ -23,7 +23,9 @@ public function getTitle() {
     protected function _renderPage(CM_Page_Abstract $page) {
         $renderAdapterLayout = new CM_RenderAdapter_Layout($this->getRender(), $page);
         $this->_title = $renderAdapterLayout->fetchTitle();
-        return $renderAdapterLayout->fetchPage();
+        
+        $renderAdapterPage = new CM_RenderAdapter_Page($this->getRender(), $page);
+        return $renderAdapterPage->fetch();
```
tomaszdurka: Wasn't this done through layout adapter for a reason?

njam: The only reason was to call both functions on the layout adapter. The implementation did exactly the same.
I'm planning to fetch the title from the page later too, I think it will be easier if all of this is on the page.

njam:
I moved the rendering of the document title from the Layout to the Page now.
So there are two functions:
- `CM_RenderAdapter_Page::fetchTitle()`: renders `title.tpl`, e.g. "My article"
- `CM_RenderAdapter_Page::fetchTitleWithBranding()`: renders `title-with-branding.tpl`, e.g. "My article - my website"

Having all of it on the Page reduces complexity of the Layout render adapter.

@tomaszdurka please review

library/CM/Site/Abstract.php
```diff
@@ -75,6 +75,13 @@ public function getModules() {
     }
 
     /**
+     * @return CM_View_Document
+     */
+    public function getDocument() {
+        return new CM_View_Document();
```
tomaszdurka: Should this return instance or rather classname? We do not consider constructor deps for it.

njam:
@tomaszdurka please review

PR #2358, opened 2016-09-29
"Data" analog for $viewResponse->addCssClass()

fvovan:
May be store a more general "list of attributes".

cc @christopheschwyzer, @njam


fvovan:
@njam is it what we need?
 I applied these attributes to `CM_RenderAdapter_Component` and `CM_RenderAdapter_FormField` but probably other places are also needed

library/CM/Frontend/ViewResponse.php
```diff
@@ -84,6 +88,40 @@ public function getCssClasses() {
...
+     * @return string[]
+     */
+    public function getDataHtml() {
+        return $this->_dataHtml;
+    }
+
+    /**
+     * @return string
+     */
+    public function getDataHtmlFormatted() {
```
njam: I would not put this rendering logic into this data holding class. Let's only expose the values from here.
For rendering it might be best to reuse the HTML tag rendering code in `smarty_function_tag`?

fvovan:
>What about storing 1 array of attributes? Both css and data-something are just attributes.

we can, not sure if it will make the code more readable.  they are not real `attributes` (because for example some css classes are added on the fly).

>I would not put this rendering logic into this data holding class. Let's only expose the values from here. For rendering it might be best to reuse the HTML tag rendering code in smarty_function_tag?

I didn't want to do it as well, but then we have to duplicate a bit of code in `RenderAdapter/Component.php` and `RenderAdapter/FormField.php`.
Not sure how to reuse `smarty_function_tag ` there. They are smarty agnostic.

njam:
Agree, it's not really more readable, because the `class` attribute is a joined string of multiple strings, while he `data-something` attributes are just set. Let's keep it separate, sorry.

Regarding `smarty_function_tag` - How about extracting the main logic into a new class? Like `CM_Frontend_HtmlTagRenderer`? This could be useful for other use cases too!

fvovan:
>How about extracting the main logic into a new class? Like CM_Frontend_HtmlTagRenderer?

Agree, but probably it's better to render only inner attributes, but not tag with its contents.
So it's rather `CM_Frontend_HtmlTagAttributesRenderer`
please review.

njam:
As discussed: let's try to extract the rendering of a full HTML tag (`smarty_function_tag`), and use it to render the `<div>`.

fvovan:
@njam please review.
I also thought about passing `class` as a separate argument (like html data) and perform its `impode` inside `$tagRenderer` but not sure about it.

library/CM/Frontend/HtmlTagRenderer.php
```diff
@@ -0,0 +1,47 @@
+<?php
+
+class CM_Frontend_HtmlTagRenderer {
+
+    /**
+     * @param string      $elementName
+     * @param string|null $content
+     * @param array|null  $attributes
+     * @param array|null  $dataHtml
```
njam: Call it `$dataAttributes`?

library/CM/Frontend/HtmlTagRenderer.php
```diff
@@ -0,0 +1,47 @@
...
+            'track', 'wbr'];
+
+        $html = '<' . $elementName;
+        foreach ($attributes as $attributeName => $attributeValue) {
+            if (isset($attributeValue)) {
+                $html .= ' ' . CM_Util::htmlspecialchars($attributeName) . '="' . CM_Util::htmlspecialchars($attributeValue) . '"';
+            }
+        }
+        foreach ($dataHtml as $dataKey => $dataValue) {
+            $html .= ' data-' . CM_Util::htmlspecialchars($dataKey) . '="' . CM_Util::htmlspecialchars($dataValue) . '"';
```
njam: Merge data attributes with other attributes and then loop once?

library/CM/Frontend/ViewResponse.php
```diff
@@ -17,13 +17,17 @@ class CM_Frontend_ViewResponse extends CM_DataResponse {
     /** @var CM_Frontend_JavascriptContainer_View */
     protected $_js;
 
+    /** @var array */
+    protected $_dataHtml;
```
njam: Call it dataAttributes?

fvovan:
@njam please review

library/CM/SmartyPlugins/function.tag.php
```diff
@@ -13,23 +13,16 @@ function smarty_function_tag(array $params, Smarty_Internal_Template $template)
...
-
-    // http://www.w3.org/TR/html-markup/syntax.html#void-element
-    $namesVoid = array('area', 'base', 'br', 'col', 'command', 'embed', 'hr', 'img', 'input', 'keygen', 'link', 'meta', 'param', 'source', 'track',
-        'wbr');
-
-    $html = '<' . $name;
-    foreach ($attributes as $attributeName => $attributeValue) {
-        if (isset($attributeValue)) {
-            $html .= ' ' . $attributeName . '="' . CM_Util::htmlspecialchars($attributeValue) . '"';
+    $dataHtml = [];
```
njam: Update name

tests/library/CM/Frontend/ViewResponseTest.php
```diff
@@ -36,4 +36,26 @@ public function testGetCssClasses() {
         $viewResponse->addCssClass('zoo');
         $this->assertSame($classNames, $viewResponse->getCssClasses());
     }
+
+    public function testAddGetSetDataHtml() {
```
njam: Update name

fvovan:
>Update name

oh. bad me. sorry..
please review

PR #2361, opened 2016-09-30
fix decoding of recursively encoded ArrayConvertibles

alexispeter:


alexispeter:
please review @tomaszdurka

library/CM/Params.php
```diff
@@ -650,6 +650,9 @@ public static function decode($value, $json = null) {
             if (!is_subclass_of($className, 'CM_ArrayConvertible')) {
                 throw new CM_Exception_InvalidParam('Class for decoding is not CM_ArrayConvertible', null, ['class' => $className]);
             }
+            if (!empty($value)) {
+                $value = self::decode($value);
+            }
```
tomaszdurka: Looks good, was a bit confusing that this piece is to actually only run `array_map decode` part.

Wdyt about? Is it more or rather less readible? You decide.

```php
    public static function decode($value, $json = null) {
        if ($json) {
            $value = CM_Util::jsonDecode($value);
        }
        
        if (is_array($value)) {
            if (isset($value['_class'])) {
                $className = (string) $value['_class'];
                if (!class_exists($className)) {
                    throw new CM_Exception_InvalidParam('Class for decoding does not exist', null, ['class' => $className]);
                }
                if (!is_subclass_of($className, 'CM_ArrayConvertible')) {
                    throw new CM_Exception_InvalidParam('Class for decoding is not CM_ArrayConvertible', null, ['class' => $className]);
                }
                unset($value['_class']);
            }
            $value = array_map('self::decode', $value);
        }
        
        if (isset($className)) {
            /** @var CM_ArrayConvertible $className */
            $value = $className::fromArray($value);
            if (!$value) {
                return false;
            }
        }
        
        return $value;
    }
```

tomaszdurka: As discussed - let's keep as is.

PR #2363, opened 2016-10-01
Breaking: Inject layout dynamically in ajax_loadPage()

njam:
Part of #2327

Based on #2355 

TODO:
- Check usages of `cm.findView()` and `cm.getLayout()`

### DIFF
--- a/library/CM/App.js
+++ b/library/CM/App.js
@@ -90,6 +90,17 @@ var CM_App = CM_Class_Abstract.extend({
   },
 
   /**
+   * @returns {CM_View_Document}
+   */
+  getDocument: function() {
+    var document = this.findView('CM_View_Document');
+    if (!document) {
+      throw new CM_Exception('Cannot find document');
+    }
+    return document;
+  },
+
+  /**
    * @param {Number} min
    * @param {Number} max
    * @return {Number}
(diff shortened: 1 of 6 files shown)


zazabe:
reminder: 
- move `loadPage()`/`_loadPageThrottled` to the document
- whenever the layout has changed, replace it and then inject the page (`<div class="page-placeholder" />`)
- check usages of `cm.findView()` / `cm.getLayout()`, they may use `loadPage()` on the layout cmp
- remove the logic to reload the page when the layout has changed, [`layoutClass`, see todo](https://github.com/cargomedia/cm/pull/2363/files#diff-bcefa35f2b86b41f511e45e3780dd24eR125)

zazabe:
It works in general, few questions still:
- [`scrollToHash`](https://github.com/cargomedia/cm/pull/2363/commits/08d003341c4b08c4b1db3271c9d77011d5d8e58f#diff-1be377e12edf0831d5bba657e9621597R142) feature: how it is supposed to work?     
for example, `cm.getDocument().loadPage('/about/help#codeOfHonor')` will load `/about/help` only, because `#codeOfHonor` is dropped by [`CM_Http_Response_View_Abstract::_createGetRequestWithUrl()`](https://github.com/cargomedia/cm/blob/master/library/CM/Http/Response/View/Abstract.php#L260)
- more questions here: https://github.com/cargomedia/sk/pull/3991

zazabe:
@vogdb please review

vogdb:
> - scrollToHash

We can drop it.

vogdb:
lvgtm. One minor note. What happens to the old layout when it is replaced? Is it destroyed somehow?

zazabe:
> What happens to the old layout when it is replaced? Is it destroyed somehow?

[`CM_View_Abstract::replaceWithHtml()`](https://github.com/cargomedia/cm/blob/master/library/CM/View/Abstract.js#L238) should take care of that, isn't it?

PR #2365, opened 2016-10-03
Enable recursive decoding of an encoded object's ArrayConvertible-properties

alexispeter:
Followup to https://github.com/cargomedia/cm/pull/2361


alexispeter:
please review @tomaszdurka

PR #2368, opened 2016-10-04
Introduce Log encoders

tomaszdurka:


tomaszdurka:
lgtm

So I should make this as followup for https://github.com/cargomedia/puppet-packages/pull/1498.

tomaszdurka:
@fvovan please review

tomaszdurka:
Well. I am not sure. I thought about making those encoders more generic.
Especially the mongodb one which could be used in general for transforming php data into mongo compatible . I agree they don't need common interface though. It brings more confusion then anything else. Requiring interface on the mongdb-handler is wrong  - as it should only accept mongdb-specific encoder.

So I would still keep encoders when they encapsulate some common functionality (MongoDB) and inline others. Also let's remove interface, you agree?

fvovan:
>So I would still keep encoders when they encapsulate some common functionality (MongoDB) and inline others. Also let's remove interface, you agree?

Yes, sounds good.

tomaszdurka:
@fvovan please review

fvovan:
If we change mongo handler's constructor signature merge will be breaking.
Do we need to prepare another PR beforehand?

tomaszdurka:
@fvovan do you approve this pr? :)

fvovan:
Yes, if it's merged after sk changes :)

zazabe:
about https://github.com/cargomedia/cm/pull/2368#issuecomment-252006362, it could be done via https://github.com/cargomedia/sk/pull/3777, wdyt @fvovan @tomaszdurka ?

PR #2369, opened 2016-10-10
Increase timeout of Phantomjs test runner

vogdb:


vogdb:
@zazabe please review

Issue #2309, opened 2016-08-27
performance of mongodb queries to retrieve logs in admin


njam:
It happens again. Probably also related to write queries happening at the same time?

Related:
https://github.com/cargomedia/cm/pull/2160
https://github.com/cargomedia/cm/pull/2228

```
mongos> db.cm_log.count()
4969165
mongos> db.cm_log.count({level:400})
20805
mongos> db.cm_log.count({level:300})
4193617
mongos> db.cm_log.getIndexes()
[
    {
        "v" : 1,
        "key" : {
            "_id" : 1
        },
        "name" : "_id_",
        "ns" : "example.cm_log"
    },
    {
        "v" : 1,
        "key" : {
            "level" : NumberLong(1),
            "createdAt" : NumberLong(1)
        },
        "name" : "level_1_createdAt_1",
        "ns" : "example.cm_log",
        "background" : true
    },
    {
        "v" : 1,
        "key" : {
            "context.extra.type" : NumberLong(1),
            "level" : NumberLong(1),
            "createdAt" : NumberLong(1)
        },
        "name" : "context_extra_type_1_level_1_createdAt_1",
        "ns" : "example.cm_log",
        "background" : true
    },
    {
        "v" : 1,
        "key" : {
            "message" : NumberLong(1)
        },
        "name" : "message_1",
        "ns" : "example.cm_log",
        "background" : true
    }
]
```

njam:
It never happens for "/log/stats" but only for "/log?level=X".

@fvovan can you investigate?

njam:
cc @kris-lab @alexispeter

kris-lab:
Example long running query:
````
		{
			"opid" : 166492112,
			"active" : true,
			"secs_running" : 1182,
			"microsecs_running" : NumberLong(1182201490),
			"op" : "query",
			"ns" : "example.cm_log",
			"query" : {
				"count" : "cm_log",
				"query" : {
					"message" : {
						"$exists" : true
					},
					"level" : {
						"$in" : [
							NumberLong(300)
						]
					}
				}
			},
			"planSummary" : "IXSCAN { level: 1, createdAt: 1 }",
			"client" : "10.55.40.160:59820",
			"desc" : "conn217970",
			"threadId" : "0x7f3eb6b16700",
			"connectionId" : 217970,
			"locks" : {
				"^fuckbook" : "R"
			},
			"waitingForLock" : true,
			"numYields" : 44658,
			"lockStats" : {
[...]

PR #2373, opened 2016-10-12
Create an level index on cm_log.

fvovan:

### DIFF
--- a/resources/db/mongo/collections.json
+++ b/resources/db/mongo/collections.json
@@ -34,6 +34,14 @@
       "options": {
         "name": "message_1"
       }
+    },
+    {
+      "key": {
+        "level": 1
+      },
+      "options": {
+        "name": "level_1"
+      }
     }
   ]
 }
--- a/resources/db/update/72.php
+++ b/resources/db/update/72.php
@@ -1,0 +1,9 @@
+<?php
+
+return;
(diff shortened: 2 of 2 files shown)


fvovan:
I read documentation again, and  still don't completely understand how it actually works.
My assumption is that either for dev or production query
`db.cm_log.find({level  : { $in : [300] } }).sort({"_id" : -1}).skip(0).limit(50) ;`
fetches every document in the collection sorted by "_id", but then filter 50 docs  matching `{level  : { $in : [300] } }` criteria. 
it can explain why `{level  : { $in : [30011] } }` takes a while, but still no any visible reasons why `{level  : { $in : [300] } }` "real" query works in a shell but fails when executed with php.

The only detail that is different  is 'batchResult' setting for `MongoCursor`, but I can't imagine how it could affect the performance. May be query optimiser denies the plan I described  ?

Basically i'm still not sure that `{level: 1}` index will resolve a problem, but it definitely needs to be tried.

If it does not help, I have a suggestion to change sorting criteria to `{"createdAt" : -1}` so both query and sorting will be "covered" with single index and this way it could be more consistent in terms of executing with PHP driver.

@njam please review
cc @kris-lab

njam:
lgtm

PR #2366, opened 2016-10-12
Smarty function "lessVariable" uses file storage

alexispeter:
In `CM_Asset_Css::_compile()` we use *file* storage because the assumption is that this code is only executed a few times and then cached by the CDN. But it turns out that `smarty_function_lessVariable` also calls this code, which is executed with every rendering.

We could probably cache the result of `Render::getLessVariable()` in APC?

@tomaszdurka @zazabe wdyt?


alexispeter:
please review @njam

njam:
@tomaszdurka please review

PR #2374, opened 2016-10-12
Only trigger slider's "change" event when value changes

njam:
@zazabe @christopheschwyzer please review


christopheschwyzer:
lgtm

PR #2376, opened 2016-10-14
Change the ordering criteria in CM_Paging_Log.

fvovan:
according to https://docs.mongodb.com/v2.6/core/index-intersection/#index-intersection-and-sort 
Lets sort by `createdAt` but not by `_id`  ( because we already have `{level:1, createdAt:1}` index)

### DIFF
--- a/library/CM/Paging/Log.php
+++ b/library/CM/Paging/Log.php
@@ -73,7 +73,7 @@ class CM_Paging_Log extends CM_Paging_Abstract implements CM_Typed {
             ];
             $source = new CM_PagingSource_MongoDb(self::COLLECTION_NAME, null, null, $aggregate);
         } else {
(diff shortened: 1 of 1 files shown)


fvovan:

fvovan:
@njam please review.
cc @kris-lab

njam:
@kris-lab @alexispeter wdyt?

alexispeter:
Does this still apply, even though you're sorting in the other direction than the index is pointing? If not, should we change the index?

fvovan:
>Does this still apply, even though you're sorting in the other direction than the index is pointing? If not, should we change the index?

It's not clear from the docs, but when i run `explain` it shows that even compound index component can be reversed if single field sorting  is required.

```
db.cm_log.find({level: {$in : [300]} }).sort({createdAt: -1}).skip(0).limit(50).explain()
{
	"cursor" : "BtreeCursor level_1_createdAt_1 reverse",
```

alexispeter:
lgtm then

PR #2377, opened 2016-10-14
Logger: record "httpRequest.query" as a string vs array

zazabe:
It's quite useless to have `httpRequest.query` as an array:
- not searchable on loggly
- produce some failures when fluentd is [used in local, in combination with  ElasticSearch](https://github.com/cargomedia/docker-fluentd)      
```
elasticsearch_1  | MapperParsingException[failed to parse]; nested: IllegalArgumentException[mapper [httpRequest.query.value] of different type, current_type [string], merged_type [long]];
```

What about recording it as a string instead? maybe as a JSON? wdyt @fvovan  @tomaszdurka? 

cc @njam

### DIFF
--- a/library/CM/Log/ContextFormatter/Cargomedia.php
+++ b/library/CM/Log/ContextFormatter/Cargomedia.php
@@ -29,7 +29,7 @@ class CM_Log_ContextFormatter_Cargomedia implements CM_Log_ContextFormatter_Inte
             ];
             $query = $request->findQuery();
             unset($query['viewInfoList']);
-            $formattedRequest['query'] = $this->_encodeAsArray($query);
+            $formattedRequest['query'] = json_encode($query);
             if (array_key_exists('http_referer', $serverArray)) {
                 $formattedRequest['referer'] = (string) $serverArray['http_referer'];
             }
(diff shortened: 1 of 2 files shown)


zazabe:
like this, wdyt @fvovan ?

fvovan:
https://github.com/cargomedia/puppet-cargomedia/issues/1341

fvovan:
Personally i don't mind but have concerns that it could bring some other limitations like too long string or something else. I don't remember have we considered Json as encoding here? 
@tomaszdurka ?

njam:
We're sending long strings in the "log" field already, it works fine.
Personally I would prefer a more human readable format.
How about pretty-printed json?

We should use `CM_Util::jsonEncode()` which does proper error handling.

zazabe:
@fvovan please review

library/CM/Log/ContextFormatter/Cargomedia.php
```diff
@@ -29,7 +29,7 @@ public function formatContext(CM_Log_Context $context) {
             ];
             $query = $request->findQuery();
             unset($query['viewInfoList']);
-            $formattedRequest['query'] = $this->_encodeAsArray($query);
+            $formattedRequest['query'] = CM_Util::jsonEncode($query, true);
```
tomaszdurka: I missed this issue in my mailbox, getting to it now.

I would vote to not json-encode in cases like this. This should be rather done by encoders for specific outputs (mongo, fluent). And if we generally accept hashes in those encoders but this is special case, maybe wrapping it in a special object and treating in special way in just some encoders would suffice.

Could be overkill though.

zazabe: Could be better, but the encoding was done in `CM_Log_ContextFormatter_Cargomedia` so I thought we could simply do the JSON encoding there...
btw, is it a problem for [`CM_Janus_HttpApiClient`](https://github.com/cargomedia/cm/blob/master/library/CM/Janus/HttpApiClient.php#L51)?

PR #2370, opened 2016-10-14
Record user's UA whenever he's online

zazabe:
related to https://github.com/cargomedia/sk-support/issues/44#issuecomment-252866401

### DIFF
--- a/library/CM/Http/Request/Abstract.php
+++ b/library/CM/Http/Request/Abstract.php
@@ -399,6 +399,7 @@ abstract class CM_Http_Request_Abstract {
      */
     public function setSession(CM_Session $session = null) {
         if (null !== $session) {
+            $session->setRequest($this);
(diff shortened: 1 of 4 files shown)


zazabe:
@njam please review

zazabe:
hum, it will not work i think, because `CM_Session::findById()` retrieve a `CM_Session` without any `CM_Http_Request_Abstract` when the `sessionId` cookie is set...
Let me check it...

zazabe:
Should be fine like that, please review @njam

njam:
`addFromRequest` is quite a heavy operation (mysql write query), so we should probably not run it for every request. How about moving it into the `if (!$user->getOnline())`?

zazabe:
@njam please review

njam:
lgtm

PR #2378, opened 2016-10-14
Location form field: trigger an event when distance field is enabled/disabled

njam:


christopheschwyzer:
lgtm

PR #2379, opened 2016-10-14
Streamline how object/model/id are logged through fluentd/mongoDB

zazabe:
- record `object` as `{class: <instanceName>}`
- record `CM_Model_Abstact` instance as `{id: '123', class: <instanceName>}`
- cast to `string` any record with an `id` key (only for FluentD, see https://github.com/cargomedia/cm/issues/2329)

### DIFF
--- a/tests/library/CM/Log/Handler/FluentdTest.php
+++ b/tests/library/CM/Log/Handler/FluentdTest.php
@@ -53,7 +53,7 @@ class CM_Log_Handler_FluentdTest extends CMTest_TestCase {
         $contextFormatter = $this->mockInterface('CM_Log_ContextFormatter_Interface')->newInstanceWithoutConstructor();
         $contextFormatter->mockMethod('formatContext')->set(['key' => 'value']);
         /** @var CM_Log_ContextFormatter_Interface $contextFormatter */
-        
+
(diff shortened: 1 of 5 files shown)


zazabe:
@fvovan please review

cc @tomaszdurka @njam

fvovan:
lgtm but probably it would be better to rename 
`Fake_Model_Fluentd` to something like `CM_Model_Mock{something}` like we usually do.

tests/library/CM/Log/Handler/FluentdTest.php
```diff
@@ -81,4 +81,40 @@ public function testSanitizeRecord() {
...
+                'bax' => [
+                    'class' => 'CM_Model_Mock_Fluentd',
+                    'id'    => '42',
+                ],
+            ],
+        ], CMTest_TH::callProtectedMethod($mock, '_encodeRecord', [$record]));
+    }
+}
+
+class CM_Model_Mock_Fluentd extends CM_Model_Abstract {
```
tomaszdurka: Why use class definition when you can mock the class in the specific test?

zazabe: Hum, I tried, but then `get_class()` was retrieving a different classname  every time...

PR #2380, opened 2016-10-17
Fluentd Logger adapter: send time with sub-second resolution

fvovan:
For https://github.com/cargomedia/puppet-cargomedia/issues/1575

Instead of:
> 2016-10-15T15:37:28+0000

Should be:
> 2016-10-15T15:37:28.123+0000

### DIFF
--- a/library/CM/Log/Record.php
+++ b/library/CM/Log/Record.php
@@ -30,8 +30,7 @@ class CM_Log_Record {
         $this->_level = $level;
         $this->_message = $message;
         $this->_context = $context;
-        $timeOfDay = gettimeofday();
-        $this->_createdAt = DateTime::createFromFormat('U.u', $timeOfDay['sec'] . '.' . $timeOfDay['usec']);
+        $this->_createdAt = DateTime::createFromFormat('U.u', (string) microtime(true));
     }
 
     /**
--- a/library/CM/Log/Record.php
+++ b/library/CM/Log/Record.php
@@ -30,8 +30,8 @@ class CM_Log_Record {
         $this->_level = $level;
         $this->_message = $message;
         $this->_context = $context;
-        $this->_createdAt = new DateTime();
-        $this->_createdAt->setTimestamp(time());
+        $timeOfDay = gettimeofday();
+        $this->_createdAt = DateTime::createFromFormat('U.u', $timeOfDay['sec'] . '.' . $timeOfDay['usec']);
     }
 
     /**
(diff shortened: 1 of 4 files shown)


fvovan:
@njam please review.

library/CM/Log/Record.php
```diff
@@ -30,8 +30,8 @@ public function __construct($level, $message, CM_Log_Context $context) {
         $this->_level = $level;
         $this->_message = $message;
         $this->_context = $context;
-        $this->_createdAt = new DateTime();
-        $this->_createdAt->setTimestamp(time());
+        $timeOfDay = gettimeofday();
```
njam: Maybe use `microtime(true)` instead - could become a oneliner?

fvovan: `microtime(true)` has less precision.
4 vs 6 fraction digits

njam: I think that's fine - milliseconds are enough, not more is visible in Loggly.

fvovan:
Unfortunately is not possible (at least easily) to hack `microtime` with php_runkit in our case.
If we do it then tests using `PHPUnit_Framework_MockObject_Generator` hangs forever.
I suppose it happens because of its internal `microtime` usage (in `PHPUnit_Framework_MockObject_Generator`)

may be leave `gettimeofday` then?
@njam wdyt?

njam:
In general I think we should get away from using runkit (see also https://github.com/cargomedia/cm/pull/2000), or at least not introduce new usages.
Why do we need to mock it here? Is there another way of testing it?

fvovan:
>Why do we need to mock it here? Is there another way of testing it?

It's paging log tests (cleanup and aggregation checkings).
We create `createdAt` field using `microtime` or `gettimeofday`, and it's impossible to use time forwarding without substitution because values of hacked `time` and  original `microtime` are not in conjunction.

njam:
What alternatives way do you see for testing it? Let's discuss together!

fvovan:
Hmm. Anyway we need to have some difference between log record's creation time and current time. So if we can't change current time we have to either be able to define time in `CM_Log_Record ` constructor  or to "hack" created records on mongoDB level.

I would prefer the latter one, because we don't need to change API's and we use mongo's capabilities to test mongo's pagings. It could be a bit cumbersome, but there are not a lot of such tests.

@njam wdyt?

njam:
As discussed let's use `gettimeofday(true)`

fvovan:
@njam please review

njam:
lgtm

PR #2384, opened 2016-10-18
Add {contex.extra.type:1, createdAt: 1} index, remove {level:1} index

fvovan:
Hopefully last stage. Sole compound index seems to work well.

We  could force  `{contex.extra.type:1, level: 1, createdAt: 1}` existing  index using if add extra  `{level : {$in : [100, 200, 300, 400, 500]}}` clause (which means "any")  to  `{type: 307}` query.
But I would prefer to be explicit now.

### DIFF
--- a/resources/db/mongo/collections.json
+++ b/resources/db/mongo/collections.json
@@ -37,10 +37,11 @@
     },
     {
       "key": {
-        "level": 1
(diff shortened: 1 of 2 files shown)


fvovan:
@alexispeter please review

alexispeter:
lgtm

PR #2386, opened 2016-10-19
Wrong return-type documented for city-maxmindId

alexispeter:
mysql-schema says int.


alexispeter:
please review @tomaszdurka

tomaszdurka:
lgtm

tomaszdurka:
should we also update the CM_Model_Location#getMaxMind() to int/string?

alexispeter:
Do you mean the parameter for CM_Model_Location::createCity()? There are not other occurences in CM_Model_Location.

PR #2387, opened 2016-10-20
Update jQuery to 2.2.4

vogdb:
To prevent memory leaks


vogdb:
Checked usual critical points. Works fine. @zazabe please review

PR #2389, opened 2016-10-20
Upgrade Functional|php

fvovan:


njam:
please check the tests

fvovan:
>please check the tests

Yep, it's a bit tricky error.
the thing is `MongoClient` adds `_id` if we don't specify it. But it adds it using kinda lowlevel magic i.e without references.
>If the parameter does not have an _id key or property, a new MongoId instance will be created and assigned to it. This special behavior does not mean that the parameter is passed by reference. 

So there we  used `Functional\map()` to created "cloned" version of passed array.
But since v1.2.0 `Functional` uses direct callback invocations instead of `call_user_func`.
And thus the trick doesn't work.

So if we really want to keep passed argument untouched we can  either copy old map implementation to `CM_MongoDb_Client::batchInsert()`  or do the "encode/decode" trick with json or with serialized value. 

@njam wdyt?

njam:
I remember vaguely! :) 
@alexispeter I think you wrote this originally? What do you think?

alexispeter:
Passing the elements of `objectList` by reference should solve the problem.
```php
$dataList = \Functional\map($objectList, function (array &$object) {
    return $object;
});
```

fvovan:
wow. thanks @alexispeter!
I would rather imagine the opposite behaviour, and tbh i still don't understand why it works considering `call_user_func` takes its arguments by value.
please review

alexispeter:
I imagine there is some pointer-magic in C involved, the documentation just mentions passing by reference if you don't want the _id-field. Whenever I try thinking about it, I have the same thoughts as you mentioned, followed by splitting headache :P

lgtm

PR #2390, opened 2016-10-21
Force index usage

fvovan:
Needed for `all=1` listing.
It completes the query to match "left prefix" rule for `{level:1, created:1}` index. (the index itself can not be reversed because we do `count` queries without sorting).

### DIFF
--- a/library/CM/Paging/Log.php
+++ b/library/CM/Paging/Log.php
@@ -119,6 +119,8 @@ class CM_Paging_Log extends CM_Paging_Abstract implements CM_Typed {
 
         if (null !== $this->_filterLevelList) {
             $criteria['level'] = ['$in' => $this->_filterLevelList];
+        } else {
+            $criteria['level'] = ['$in' => array_values(CM_Log_Logger::getLevels())];
         }
 
         if (null !== $this->_ageMax) {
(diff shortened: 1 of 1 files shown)


fvovan:
@tomaszdurka please review

library/CM/Paging/Log.php
```diff
@@ -119,6 +119,8 @@ protected function _getCriteria() {
 
         if (null !== $this->_filterLevelList) {
             $criteria['level'] = ['$in' => $this->_filterLevelList];
+        } else {
+            $criteria['level'] = ['$in' => array_values(CM_Log_Logger::getLevels())];
```
tomaszdurka: Let's rather make the whole param _filterLevelList not-nullable?
So in constructor we can set it to all levels by default.

```php
public function __construct(array $filterLevelList = null, $filterType = null, $aggregate = null, $ageMax = null) 
    if (null === $filterLevelList) {
        $filterLevelList = array_values(CM_Log_Logger::getLevels());
    } else {
        foreach ($filterLevelList as $level) {
            $level = (int) $level;
            if (!CM_Log_Logger::hasLevel($level)) {
                throw new CM_Exception_Invalid('Log level does not exist.', null, ['level' => $level]);
            }
        }
    }
```

fvovan:
@tomaszdurka please review.
it think getter is not needed here, `_filterLevelList` is used only once

fvovan:
it didn't help. will be reverted in https://github.com/cargomedia/cm/pull/2399/files

PR #2393, opened 2016-10-25
Fix fractional dateTime creation

fvovan:
`DateTime::createFromFormat('U.u', (string) gettimeofday(true))`  is failed when called in a moment of time without fractional parts.


fvovan:
please review @njam.
I think explicit version would be better here.

njam:
lgtm

Should we provide a `CM_Util::createDateTimeWithMillis()`? I guess it will be useful in other places.

fvovan:
agree it could be useful.
please confirm @njam

njam:
lgtm

PR #2392, opened 2016-10-25
Fix file upload adjustments

vogdb:
follow-up https://github.com/cargomedia/cm/pull/2383

trigger "uploadComplete" in _done_ only

PR #2397, opened 2016-10-25
Fix file upload adjustments

vogdb:
Followup of #2392

### DIFF
--- a/library/CM/FormField/File.js
+++ b/library/CM/FormField/File.js
@@ -64,7 +64,6 @@ var CM_FormField_File = CM_FormField_Abstract.extend({
 
       },
       done: function(e, data) {
-        inProgressCount--;
         if (data.result.success) {
           while (field.getOption("cardinality") && field.getOption("cardinality") < field.$('.previews .preview').length) {
             field.$('.previews .preview').first().remove();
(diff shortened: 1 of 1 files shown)


vogdb:

vogdb:
@njam please review

vogdb:
@zazabe please review

vogdb:
anyone?

christopheschwyzer:
Code looks ok to me. 
But I'm wondering - If cardinality=1 we probably should just remove the _multiple_ attribute from the form input field - this way we would prevent uploading multiple files altogether. 
cc @njam

njam:
*reviewing*

library/CM/FormField/File.js
```diff
@@ -66,13 +67,20 @@ var CM_FormField_File = CM_FormField_Abstract.extend({
...
         if (data.result.success) {
           data.$preview.html(data.result.success.preview + '<input type="hidden" name="' + field.getName() + '[]" value="' + data.result.success.id + '"/>');
-          if (inProgressCount === 0) {
-            field.trigger("uploadComplete", data.files);
-          }
         } else if (data.result.error) {
           data.$preview.remove();
           field.error(data.result.error.msg);
         }
+        if (inProgressCount === 0) {
```
njam: Why only check it when the last upload finishes? It means that I can upload 10 files to a field with the cardinality of 2, and the 8 too many will only be removed when all are successfully uploaded.
How about just remove them as they come in?

vogdb:
@njam please look

vogdb:
@njam please review once more, sorry

njam:
ok lgtm

PR #2313, opened 2016-10-25
Helper to create a promise cancellable by some events

zazabe:
Could be nice to have a helper on `CM_View_Abstract` to create a promise that will be automatically cancelled if some specific events are triggered.


zazabe:
@vogdb please review

PR #2398, opened 2016-10-26
Use autotrack for google analytics

zazabe:


zazabe:
There's one issue with this solution, page loaded by ajax will be tracked twice, because both autotack and CM are calling `ga("send", "pageview", ...)`...
We could send "pageview" with CM on page load only, but then some custom tracking may not work anymore I guess... wdyt @njam?

zazabe:
Note that `autotrack` will not send "pageview" when the page is loaded, only on `pushState/replaceState`...

zazabe:
I think the best is to not use `autotrack` and to send "pageview" in `cm.route.router` directly

njam:
sgtm!

zazabe:
as discussed with @njam, let's drop the `pushState` tracking, to avoid having too many tracked url on GA, but let's keep the GA plugin support on `CMService_GoogleAnalytics_Client`...

zazabe:
@njam please review

njam:
lgtm

PR #2400, opened 2016-10-28
CM_Form_Abstract.submit error handling

vogdb:
@njam @tomaszdurka your notes?
Currently in `CM_Form_Abstract.submit` we catch and display `CM_Exception_FormFieldValidation`. Does it make any sense to propagate this error any further if it's already handled inside? I see that we don't catch such error anywhere.

### DIFF
--- a/library/CM/Form/Abstract.js
+++ b/library/CM/Form/Abstract.js
@@ -45,9 +45,7 @@ var CM_Form_Abstract = CM_View_Abstract.extend({
       $btn.on(event, {action: name}, function(event) {
         event.preventDefault();
         event.stopPropagation();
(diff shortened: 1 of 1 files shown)


vogdb:
@njam please review
btw where do we use this `response.exec`? I have not found any usages of it.
```js
        if (response.exec) {
          handler.evaluation = new Function(response.exec);
          handler.evaluation();
        }
```

njam:
For example `CM_Http_Response_View_Abstract::reloadComponent()` will send it.

@zazabe could you review this?

vogdb:
@zazabe please don't forget about me....

PR #2405, opened 2016-11-02
Drop {type, level, createdAt} index

fvovan:

Personally I doubt that it won't affect the performance. On dev environment it shows "QueryOptimizerCursor" (which is usually slow on production servers) instead of  "Complex Plan" (which should be in case of intersected indexes according documentation), but maybe it's worth to try.

@njam wdyt?

### DIFF
--- a/resources/db/mongo/collections.json
+++ b/resources/db/mongo/collections.json
@@ -19,16 +19,6 @@
     },
     {
       "key": {
-        "context.extra.type": 1,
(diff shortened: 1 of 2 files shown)


njam:
I would suggest to leave it as-is if you're unsure. wdyt?

fvovan:
agree. can be checked when(if) we really experience problems with RAM capacity or insertion performance

PR #2401, opened 2016-11-04
improve CM_FormField_File

vogdb:
- update `jquery.fileupload`
- do not send redundant files to upload. Now we check for them only after they have been uploaded.


vogdb:
@zazabe please review

vogdb:
@zazabe please review.
@njam FYI.

PR #2408, opened 2016-11-08
Ability to change floatbox' options

vogdb:


vogdb:
@zazabe or @christopheschwyzer please review

njam:
@zazabe can you review this? https://github.com/cargomedia/cm/pull/2408

client-vendor/after-body/jquery.floatbox/jquery.floatbox.js
```diff
@@ -33,6 +33,13 @@
     $parent: null,
     $layer: null,
     $floatbox: null,
+    setOptions: function(value){
+      if (_.isObject(value)) {
+        //`value` as object to rewrite options
+        _.extend(this.options, value);
```
zazabe: maybe throw if `value` is not an object?

client-vendor/after-body/jquery.floatbox/jquery.floatbox.js
```diff
@@ -155,4 +162,12 @@
       }
     });
   };
+  $.fn.floatbox = function(method, value) {
```
zazabe: From what I understand, it will be mainly use to call `setOptions()`, why not introducing a more specific `$.fn.setOptions`? instead?

vogdb: `close` method is also a target here

vogdb:
@zazabe please review

client-vendor/after-body/jquery.floatbox/jquery.floatbox.js
zazabe: ok, let's keep it like that!
anyhow `$.fn.setOptions` is not correct, should be `$.fn.floatboxSetOptions` I guess, which looks like 💩  ...

PR #2409, opened 2016-11-08
Model: don't call "_onChange" when creating a model

alexispeter:
Carefully check the implementation, and move/duplicate some things to `_onCreate`.


njam:
@alexispeter what's the status here?

alexispeter:
There is no status at the moment.

alexispeter:
cm-part is done. I wanted to check how this change would affect sk and adjust if necessary before I merge anything. I haven't found the time to do this yet.

alexispeter:
please review @njam

Model-creation issues both _onCreate() and _onChange() when using a simple model with `replace`-mechanics. I'm not too happy about that, but we have no way of distinguishing real creation from a simple updating create.

njam:

library/CM/Model/Abstract.php
```diff
@@ -94,9 +94,11 @@ public function commit($useReplace = null) {
                 $this->_loadAssets(true);
                 $cache->save($type, $this->getIdRaw(), $this->_getData());
             }
-            $this->_onChange();
             $this->_changeContainingCacheables();
             $this->_onCreate();
+            if ($useReplace) {
+                $this->_onChange();
+            }
```
tomaszdurka: That is a bit tricky as we don't really know if this is create or update, but I guess we have no other choice.

tomaszdurka:
Before merging (or after) it would be nice to check it as well with dependant repos (sk). I don't think we rely on this change tirggered from creates anywhere, but worth checking.

PR #2412, opened 2016-11-11
Introduce CM_MongoDb_Client::rename()

zazabe:


zazabe:
@fvovan please review

zazabe:
FYI:

> WARNING
If the renameCollection operation does not complete, the target collection and indexes will not be usable and will require manual intervention to clean up.
WARNING
This command obtains a global write lock and will block other operations until it has completed.

https://docs.mongodb.com/manual/reference/command/renameCollection/#behavior

zazabe:
cc @alexispeter @njam

library/CM/MongoDb/Client.php
```diff
@@ -243,6 +243,36 @@ public function count($collection, array $criteria = null, array $aggregation =
...
+        if (!$dropTarget && $this->existsCollection($collectionTarget)) {
+            throw new CM_MongoDb_Exception('Target collection already exists', null, [
+                'collectionSource' => $collectionSource,
+                'collectionTarget' => $collectionTarget
+            ]);
+        }
+        $result = $this->_getClient()->selectDB('admin')->command([
+            'renameCollection' => $this->_getDatabaseName() . '.' . $collectionSource,
+            'to'               => $this->_getDatabaseName() . '.' . $collectionTarget,
+            'dropTarget'       => $dropTarget ? 'true' : 'false'
```
fvovan: It's not necessary to pass stringified values there. just boolean `$dropTarget` will be okay.

library/CM/MongoDb/Client.php
```diff
@@ -243,6 +243,36 @@ public function count($collection, array $criteria = null, array $aggregation =
...
     /**
+     * @param string    $collectionSource
+     * @param string    $collectionTarget
+     * @param bool|null $dropTarget
+     * @return array
+     * @throws CM_MongoDb_Exception
+     */
+    public function rename($collectionSource, $collectionTarget, $dropTarget = null) {
+        $dropTarget = (bool) $dropTarget;
+        if (!$this->existsCollection($collectionSource)) {
```
fvovan: It would be good to explicetely cast `$collectionSource` and `$collectionTarget` to strings

zazabe:
@fvovan please check again

> Additionally we need to verify that mongo user has appropriate permissions on admin database on production

@njam wdyt?

fvovan:
lgtm

zazabe:
@njam should we drop this? see `mongo-rename` conversation...

njam:
Would be good to have in general.

@alexispeter does the behaviour change with the [new mongodb client](https://github.com/cargomedia/cm/pull/2331)? If it's similar then maybe keep it?

alexispeter:
The new client doesn't have a renameCollection function either, so you would also have to run a command directly on the admin db.

njam:
I understand it work in general, but not on production, probably due to permissions.
I would still keep it.
wdyt?

alexispeter:
Can't hurt.

zazabe:
ok, let's merge it then!

PR #2413, opened 2016-11-11
Display CM_Exception::metaInfo in phpunit results

zazabe:
nice to have, handy in dev since we use more generic message in exceptions now

### DIFF
--- a/tests/helpers/CMTest/library/CMTest/ExceptionWrapper.php
+++ b/tests/helpers/CMTest/library/CMTest/ExceptionWrapper.php
@@ -2,19 +2,10 @@
 
 class CMTest_ExceptionWrapper extends PHPUnit_Framework_ExceptionWrapper {
 
-    public function __construct($e) {
+    public function __construct(CM_Exception $e) {
         parent::__construct($e);
-
-        if ($e instanceof CM_Exception) {
-            $message = $e->getMessage();
-            $variableInspector = new CM_Debug_VariableInspector();
-            $metaInfo = Functional\map($e->getMetaInfo(), function ($value) use ($variableInspector) {
-                return $variableInspector->getDebugInfo($value);
-            });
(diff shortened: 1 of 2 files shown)


njam:
Maybe use `CM_ExceptionHandling_Formatter_Plain::getMetaInfo()`?

zazabe:
could works, but in this case, exception must be converted into `CM_ExceptionHandling_SerializableException` first, is it worth it?

njam:
True. Is there a way we can print the *full* exception?
Then we could just call `CM_ExceptionHandling_Formatter_Plain::formatException(Exception $exception)`, and it would be fully consistent with our CLI output.

zazabe:
We don't want to format the stacktrace, but anyhow, this PR is more a hack...

[long story]
I tried different approaches:
- phpunit allows to define your own `PHPUnit_TextUI_ResultPrinter`, but unfortunately, IDEs like PhpStorm are already creating one (ie. `/home/vagrant/.phpstorm_helpers/phpunit.php`) and we can't easily override it i think.
- we can also [add a listener](https://phpunit.de/manual/current/en/appendixes.configuration.html#appendixes.configuration.test-listeners) in `phpunit.xml` for  [`PHPUnit_Framework_TestListener::addError`](https://github.com/sebastianbergmann/phpunit/blob/5.5/src/Framework/TestListener.php#L25), but it will not be nicely printed at the end of the test because this method is called when the error is thrown...

Finally, i thought that the easiest way is to override the exception message with more details and then let phpunit display it... hacky but it works...

njam:
I see, let's keep it like you did it then, that meta data is definitely useful.
I would still use `CM_ExceptionHandling_Formatter_Plain::getMetaInfo()` because it's the canonical way to print exception meta info, wdyt?

zazabe:
@njam please review

njam:
lgtm

PR #2410, opened 2016-11-15
Fix lazyload memory leak

vogdb:
Currently lazyload leaks a lot with ajax-navigation. It has to be a well-known but still not fixed problem. tuupola/jquery_lazyload#269
In general we need to update the library to the latest master and custom-patch it for now to have `detroy` method.
cc @zazabe @njam


vogdb:
@zazabe please review. Did I manage the placement of the new library correct? Should I move it elsewhere like `after-body-main`?

vogdb:
@tomaszdurka please review

njam:
lgtm

I would suggest to wait a few days, to see if we get a response to those PRs, wdyt?

vogdb:
ok. I expected them to be merged/edited quicker cause my previous requests had callbacks less than 24hours. However now it is the third day of waiting :(

vogdb:
@njam FYI

PR #2414, opened 2016-11-16
CM_Jobdistribution_Cli: restart process after X job runs

tomaszdurka:
Goal: restart job worker processes once in a while to avoid filling up memory. Like apache (or php-fpm) we could just exit a job worker after X job runs. The CLI manager would then restart the worker.

Default: 1000

screen shot 2016-11-14 at 10 38 17

### DIFF
--- a/library/CM/Jobdistribution/Cli.php
+++ b/library/CM/Jobdistribution/Cli.php
@@ -6,7 +6,7 @@ class CM_Jobdistribution_Cli extends CM_Cli_Runnable_Abstract {
      * @keepalive
      */
     public function startWorker() {
-        $worker = new CM_Jobdistribution_JobWorker();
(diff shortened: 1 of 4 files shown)


tomaszdurka:
@alexispeter or @njam please review

alexispeter:
lgtm

njam:
lgtm

njam:
This will result in warnings "Respawning dead child." correct?
@tomaszdurka what do you think? Should we avoid it somehow?

njam:
Let's use `$workloadResult->isSuccess())` to decide whether to log a *warning* or an *info*.

njam:
I can't push to your branch.
Can you merge in this commit and release it? https://github.com/njam/CM/commit/7e476976c2bc87595b2c5a697439c603ba1714dc

tomaszdurka:
lgtm

PR #2419, opened 2016-11-19
Add deprecation annotation for mysql DELAYED insert

njam:
@ppp0 please review


ppp0:
lgtm!

Please keep in mind that we run `mysql5.5` and (hopefully!) still considering moving to mariadb. This will not be possible anymore when upgrading to mysql5.6.

njam:
good to know!

PR #2420, opened 2016-11-22
Breaking: Introduce "cm app set-config"

zazabe:


zazabe:
@njam please review

njam:
Nice!

Maybe concatenate a string and `write()` it in an atomic operation, instead of using `appendLine()`?

zazabe:
ok, please check again @njam

njam:
lgtm

PR #2421, opened 2016-11-22
Change response processing error message

njam:
THis error has nothing to do with HTML, rather with HTTP.

@fvovan please review
cc @tomaszdurka


fvovan:
I guess it should have been `HTTP response error`
But that one is good as well

PR #2425, opened 2016-11-23
Run synchronous jobs always with prio HIGH

njam:
Reasoning: since we run the job synchronously we usually want the response ASAP.


alexispeter:
lgtm

PR #2423, opened 2016-11-23
"undefined is not an object" in `promise.finally` of promiseThrottler

vogdb:


vogdb:
@njam please review. Unfortunately can't reproduce, so have to monkey-patch here.

njam:
@zazabe could you review?

zazabe:
I can imagine it happens because [some throttled promises](https://github.com/cargomedia/sk/blob/master/library/SK/library/SK/Form/SearchContent.js#L32) use [`cm.router.route`](https://github.com/cargomedia/cm/blob/master/library/CM/App.js#L1435), which may return `null` in some conditions (bug introduced by me :/ )...

Anyhow, what about wrapping `fn.apply(this, arguments)` by a `Promise.try()` to always handle its return as a promise?

vogdb:
> I can imagine it happens because [some throttled promises](https://github.com/cargomedia/sk/blob/master/library/SK/library/SK/Form/SearchContent.js#L32) use [`cm.router.route`](https://github.com/cargomedia/cm/blob/master/library/CM/App.js#L1435), which may return `null` in some conditions (bug introduced by me :/ )...

Thank you so much for this. I've missed this. Thought we always return promises in out usages. Will try to fix it first and see

vogdb:
@zazabe please look

zazabe:
looks good, but it would be safer to also check if the `promiseThrottler` function returns a `Promise`, no? 
wrapping with `Promise.try()` could be tricky because it will postpone its execution, but maybe the same code as before would be enough...

vogdb:
> but maybe the same code as before would be enough...

@zazabe you mean @1a541b4 ? Should I add `instanceof Promise`?

client-vendor/after-body/promise-utils/promise-throttler.js
```diff
@@ -48,10 +48,13 @@
     }
 
     return function() {
-      if (!promise || !promise.isPending()) {
+      if (!promise) {
         promise = fn.apply(this, arguments);
-        promise.finally(resetClosurePromise.bind(null, promise));
-        return promise;
+        //we check here for `promise` again cause it can be nullified `resetClosurePromise`
```
zazabe: it's more about what `fn` returns I guess

vogdb:
@zazabe please review

library/CM/App.js
```diff
@@ -1433,7 +1433,7 @@ var CM_App = CM_Class_Abstract.extend({
       var urlBase = cm.options.urlBase;
       if (0 === url.indexOf(urlSite)) {
         var path = url.substr(urlBase.length);
-        cm.getDocument().loadPage(path);
+        return cm.getDocument().loadPage(path);
```
zazabe: hope it's the only misusage of `promiseThrottler`...

zazabe:
Oops, `qunit` tests are broken, could you open a followup @vogdb?

PR #2427, opened 2016-11-23
null is not an object (evaluating 'containerContext.images.filter')

vogdb:
`var batch=containerContext.images.filter(inview);`


vogdb:
@njam please review

njam:
ok!

lgtm

PR #2430, opened 2016-11-24
Followup of #2423

vogdb:
Followup of #2423


vogdb:
@zazabe please review

PR #2432, opened 2016-11-25
Make CM_Model_User:activityStamp optional

zazabe:


zazabe:
@tomaszdurka please review

library/CM/Model/User.php
```diff
@@ -216,7 +218,8 @@ public function getEnvironment() {
     }
 
     public function updateLatestActivityThrottled() {
-        if ($this->getLatestActivity() < time() - self::ACTIVITY_EXPIRATION) {
+        $activityStamp = $this->getLatestActivity();
+        if (null === $activityStamp || $activityStamp < time() - self::ACTIVITY_EXPIRATION) {
```
tomaszdurka: Shouldn't this be `null !== $activityStamp && $activityStamp < time() - self::ACTIVITY_EXPIRATION`?

library/CM/Model/User.php
```diff
@@ -266,7 +269,7 @@ public static function offlineOld() {
 			SELECT `o`.`userId`
 			FROM `cm_user_online` `o`
 			LEFT JOIN `cm_user` `u` USING(`userId`)
-			WHERE `u`.`activityStamp` < ? OR `u`.`userId` IS NULL',
+			WHERE `u`.`activityStamp` < ? OR `u`.`activityStamp` IS NULL OR `u`.`userId` IS NULL',
```
tomaszdurka: Shouldn't we rather ignore those which doesn't have activityStamp, cause they should have no entry for `cm_user_online`?

library/CM/Model/User.php
zazabe: I think it's correct like that, because we want to set it if `$activityStamp == null` when this method is called.

library/CM/Model/User.php
zazabe: maybe useless, but having `u.activityStamp IS NULL` at least clarify that a user without  `activityStamp` is considered as offline.

tomaszdurka: ok

library/CM/Model/User.php
```diff
@@ -79,6 +80,7 @@ public function setOnline($state = true, $visible = true) {
             CM_Db_Db::delete('cm_user_online', array('userId' => $this->getId()));
             $this->_set(array('online' => null, 'visible' => null));
         }
+        $this->updateLatestActivityThrottled();
```
tomaszdurka: Do we still need `updateLatestActivityThrottled` usage in CM_Session?

zazabe: I think i've put this here because i thought changing the online state is one kind of activity update.
The `CM_Session` usage makes sense also, because we want to update it, no matter if the user is already online or not.

now I'm wondering, does it make sense to update it when someone goes offline? maybe only when it's a user action...
So at the end I'm not sure about this line, we could also remove it, it's not directly related to this PR anyhow...

zazabe: @tomaszdurka wdyt? [ 👍 / 👎 ]

zazabe: Feel strange to remove it and change `CM_Model_UserTest::testOfflineOld()`, seems wrong... how a user can be online with `activityStamp = null`?

zazabe: ...ok, could make sense for a user created through an API and with the online state set by the API... let me remove it then!

tomaszdurka:
Ah, ok then, thought it was indented. Build fails.

zazabe:
as discussed with @tomaszdurka, let's not tied online and activityStamp together

PR #2418, opened 2016-11-25
Don't delay insert on "cm_user_online"

fvovan:
We sometimes experience issues with the way we insert rows into the `cm_user_online` table, which breaks replication (due to duplicate primary key).

```
161119 11:57:57 [ERROR] Slave SQL: Could not execute Write_rows event on table cm_user_online; Duplicate entry '16630921' for key 'PRIMARY', Error_code: 1062; handler error HA_ERR_FOUND_DUPP_KEY; the event's master log mysql-bin.000367, end_log_pos 947677533, Error_code: 1062
161119 11:57:57 [Warning] Slave: Duplicate entry '16630921' for key 'PRIMARY' Error_code: 1062
```

We use the `REPLACE DELAYED` statement to add rows to the table.
The `DELAYED` insert/replace statements have been [removed completely in MySQL 5.7](http://mysqlserverteam.com/removal-and-deprecation-in-mysql-5-7/) due to many problems.

Let's find a different way to insert these rows, to avoid having this problem again in the future.
We might want to consider migrating this table to InnoDB, but not sure.
We need to observe the load on the table (New Relic can help), and then probably benchmark a few different ways.

Note also that this table is JOINed when loading a user model. But that could also be done in 2 queries, maybe performs better.

@ppp0 @kris-lab fyi
@fvovan @fauvel @dlondero any takers?

### DIFF
--- a/library/CM/Config/Generator.php
+++ b/library/CM/Config/Generator.php
@@ -56,9 +56,7 @@ class CM_Config_Generator extends CM_Class_Abstract {
         $this->generateClassTypes();
         $config = new CM_Config_Node();
         foreach ($this->getNamespaceTypes() as $namespaceClass => $typeList) {
-            $config->$namespaceClass->types = \Functional\map($typeList, function ($className) {
-                return "$className::class";
-            });
+            $config->$namespaceClass->types = $typeList;
         }
         $classTypes = $this->getClassTypes();
         foreach ($classTypes as $type => $class) {
--- a/library/CM/Config/Generator.php
+++ b/library/CM/Config/Generator.php
@@ -56,7 +56,9 @@ class CM_Config_Generator extends CM_Class_Abstract {
         $this->generateClassTypes();
         $config = new CM_Config_Node();
         foreach ($this->getNamespaceTypes() as $namespaceClass => $typeList) {
-            $config->$namespaceClass->types = $typeList;
+            $config->$namespaceClass->types = \Functional\map($typeList, function ($className) {
+                return "$className::class";
+            });
         }
         $classTypes = $this->getClassTypes();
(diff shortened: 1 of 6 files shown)


fvovan:
@njam wdyt?

ppp0:
@fvovan can this operation be reverted (ie InnoDB -> myISAM) should we measure a drop in performance?

fauvel:
How big is this table in production? How long would the migration script take to run?

fvovan:
>How big is this table in production? 

~ 4500

>How long would the migration script take to run?

on dev without decent load it's instant (30ms)

fauvel:
So let's not have this "return" in the migration script?

fvovan:
actually I put `return` in order to be able to run the migration a bit later after deployment.
So if release contains several PRs and  performance is okay right after the deployment, but it becomes poor with the migration, we can be 100% sure about the reasons.
But maybe i'm too paranoid :)

fauvel:
Ah ok. Still maybe it's not so good to remove the delayed insertion in the code before migrating to InnoDB manually, that's why I thought we should rather do both at the same time.

ppp0:
I believe the migration could be done while the app is offline. It shouldn't take too long. To be really sure,  it should be feasible to test the duration on a dev box with the full production size of this table. I can give you a dump (`count(*)` as of now is ca 4200 records)

fvovan:
I already tested engine migration even on twice bigger table.
Yes. I agree with Sebastien, there is no point to run it later since "delayed" will be removed. Will update PR a bit later

fvovan:
@fauvel  please review

fvovan:
thanks! let's release it tomorrow morning then to have enough time if something goes wrong.

PR #2431, opened 2016-11-25
CM_Form_Abstract autoSave param

vogdb:
Via https://github.com/cargomedia/sk/issues/4201

### DIFF
--- a/layout/default/FormField/Abstract/default.less
+++ b/layout/default/FormField/Abstract/default.less
@@ -5,3 +5,11 @@
     .animation(highlightErrorField, 1000ms, ease, 0s, 1);
   }
 }
+
+&[data-formfield-success=true] {
+  .textinput, select + .button, input[type=checkbox] + label::before, input[type=radio] + label::before {
+    outline: 2px solid transparent;
+    outline-offset: -1px;
+    .animation(highlightSuccessField, 1000ms, ease, 0s, 1);
+  }
+}
--- a/layout/default/FormField/Abstract/default.less
+++ b/layout/default/FormField/Abstract/default.less
@@ -2,7 +2,7 @@
   .textinput, select + .button, input[type=checkbox] + label::before, input[type=radio] + label::before {
     outline: 2px solid transparent;
     outline-offset: -1px;
-    .animation(highlightErrorField, 1000ms, ease, 0s, 1);
+    animation: highlightErrorField 1000ms ease 0s 1;
   }
 }
 
@@ -10,6 +10,6 @@
   .textinput, select + .button, input[type=checkbox] + label::before, input[type=radio] + label::before {
     outline: 2px solid transparent;
     outline-offset: -1px;
-    .animation(highlightSuccessField, 1000ms, ease, 0s, 1);
+    animation: highlightSuccessField 1000ms ease 0s 1;
   }
 }
--- a/layout/default/FormField/Abstract/default.less
+++ b/layout/default/FormField/Abstract/default.less
@@ -1,15 +1,18 @@
+& {
+  position: relative;
+}
+
 &[data-formfield-error=true] {
-  .textinput, select + .button, input[type=checkbox] + label::before, input[type=radio] + label::before {
-    outline: 2px solid transparent;
-    outline-offset: -1px;
-    animation: highlightErrorField 1000ms ease 0s 1;
+  .formFieldFeedback-error {
+    visibility: visible;
+    animation: formError 1000ms ease 0s 1;
+    animation-fill-mode: forwards;
   }
 }
 
 &[data-formfield-success=true] {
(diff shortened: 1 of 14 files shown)


njam:
> shall we pass `autoSave` through back-end? Like:

I would say yes. Let's call it just "autosave" (lower case "s"), because the word describes a concept on its own.

> Shall `autoSave` exclude submit buttons? Theoretically we can have auto-save for one action and button-save for another. Do we need such complication?

I would not cover this case. But I can imagine that it will somewhat okay-ish by default. But it's not a required use case.

> Should we add `data-formfield-success` only when the form was submitted on auto-save or anytime the form was submitted despite the method? 

This attribute should be added to a field *only when the form was autosaved*. The reasoning is that only then we now the specific field that we responsible for the "success".

vogdb:
So, the final logic is:
- we add `data-formfield-error` to the field anytime the field has an error.
- We add `data-formfield-success` only when the field is autosaved successfully.
- Html form representation has `data-autosave="true"`
- `autosave` property transfers form action from back-end to front-end

@njam please look

njam:
@christopheschwyzer wanna review?

layout/default/FormField/Abstract/default.less
```diff
@@ -1,4 +1,4 @@
-&.hasError {
+&.data[formfield-error=true] {
```
christopheschwyzer: Selector should be: `&[data-formfield-error=true]`

library/CM/Form/Abstract.js
```diff
@@ -49,10 +52,22 @@ var CM_Form_Abstract = CM_View_Abstract.extend({
...
-    this.$el.on('submit', function() {
-      handler.$el.find('input[type="submit"], button[type="submit"]').first().click();
-      return false;
-    });
+    if (this.autosave) {
+      this.on('change', function(field) {
+        if (field) {
+          handler.submit(handler.autosave).then(function() {
+            if (field.getValue()) {
+              field.success();
```
christopheschwyzer: I think we shouldn't trigger success but rather do a validation first? E.g. when you enter "foo" into an email field it shouldn't be highlighted as success.

vogdb:
> I think we shouldn't trigger success but rather do a validation first? E.g. when you enter "foo" into an email field it shouldn't be highlighted as success.

@christopheschwyzer @njam I want to remind that this works only when the submitted form is valid at any time. This happens cause we submit form with `handleErrors: false`. It means that in case of failure we receive all errors as one big text message. It will be very hacky to extract the exact error message of the target field. So in case if two fields fail during validation, we will show the error message for both of them in the autosaved field. If we use `handleErrors: true` then the form will handle and show errors on its own which is good but this will always make `form.submit` to resolve successfully even in case of validation errors. So, we can't detect 100% whether the autosaved field was submitted successfully.

@christopheschwyzer please look.

njam:
It sounds like it would be better to use `handleErrors: true` to have proper error handling.
How about changing the `submit()` function to *even* reject the promise in that case? We would need to check the usages. But it doesn't look like it would be a big problem.

vogdb:
> It sounds like it would be better to use `handleErrors: true` to have proper error handling...

We need to duplicate a `catch` then in all places where we submit manually from js. We have 15 such places. In all of them we would need
```js
form.submit('action').catch(
  CM_Exception_FormFieldValidation,
  function() {
    if(options.handleErrors) {
      // do nothing cause it is handled
    }
  }
)
```

Does not look so nice :) We also can have alternatives. We can split `submit` on two methods: `submit and handle validation` and `submit only`. Of course the first is based on the second. In those 15 places we will use the first. In autosave we will use the second.
Another solution would be to have a method on Form like `getCurrentErrors` that returns all errors the form has now. So anytime we catch we can retrieve the map with errors.

vogdb:
According to the discussion we shall split `submit` on two methods. One for submit (including triggering of events), another for handling errors of submit.

vogdb:
I am sorry. I've rebased the PR to make a cleaner history. Now it can be easily reviewed a commit by commit and does not contain any back and forth redundancy.
key changes:
 - `trigger('complete')` is removed. CM_Form_Abstract does not trigger such event anymore cause it is not used at all and this event didn't fit into the logic of splitting :(
 - `required` errors that previously didn't trigger `error` event, now trigger it.

@njam FYI. @christopheschwyzer please review.

christopheschwyzer:
I added improved form feedback. 
I tested a bit and it seems fine - although I don't understand 100% what's going on. I understand that _autosave_ only works for form actions without any required field. Can we somehow make this fact obvious?

vogdb:
> I understand that _autosave_ only works for form actions without any required field. Can we somehow make this fact obvious?

Nope. Actually now it can work with any form :) due to the implemented split of `submit` 🎉

njam:
Looks great! Agree to release on Monday though.

PR #2417, opened 2016-11-28
Make app generate-config-internal use ::class constants instead of strings

fvovan:


fvovan:
@tomaszdurka please review.
It looks like too straightforward solution but i'm not sure if it's worth to make it more flexible..

tomaszdurka:
I would not make it optional. Let's force class constants.

fvovan:
The actual reason of this optional choice is https://github.com/cargomedia/cm/blob/master/tests/helpers/CMTest/library/CMTest/TestSuite.php#L26

So we either need to evaluate this strings like we did in `exportAsString()` or just return strings as before.

fvovan:
@tomaszdurka wdyt?

tomaszdurka:
Still not sure why we need that? :)

fvovan:
`getConfigClassTypes()` has to return strings (in order to be processed with `CM_Config_Node::exportAstring()`)  and it returns let's say `"CM_Class_Abstract::class"`   but  not `CM_Class_Abstract::class`.
So when dealing with class constants I have to handle and resolve these "class ending" strings in `CM_Config_Node::exportAstring()` (please see the changes).
But when we generate internals on the fly for `CM_Tests`, they are not processed with CM_Config_Node, so without conversion `"CM_Class_Abstract::class"` value will be written to config node.

tomaszdurka:
Oh dears. What about always have string "Foo::class" and either export/write it or eval it?

fvovan:
>What about always have string "Foo::class" and either export/write it or eval it?

Not sure if it's more clear now :)
please review @tomaszdurka

tomaszdurka:
Me neither! :)))
Lets say it's good!

PR #2434, opened 2016-11-29
Revert InnoDB for cm_user_online

fvovan:


fvovan:
revert of https://github.com/cargomedia/cm/pull/2418

fvovan:
@njam please review. I suggest just to revert it, but not sure how we avoid background jobs errors

njam:
@fvovan closeable since this was not the cause of the problem? :)

fvovan:
agree

PR #2440, opened 2016-12-02
Install specific version of colorjizz

njam:


njam:
@vogdb please review

See https://github.com/mikeemoo/ColorJizz-PHP/commits/master

PR #2446, opened 2016-12-05
Streamline floatbox

vogdb:


vogdb:
@zazabe please review

PR #2452, opened 2016-12-06
Handle Required errors as separate Exception

vogdb:


vogdb:
@njam please look

library/CM/Exception/FormFieldRequired.js
```diff
@@ -0,0 +1,16 @@
+(function(global) {
+
+  /**
+   * @class CM_Exception_FormFieldRequired
+   * @extends CM_Exception_FormFieldValidation
+   */
+  function CM_Exception_FormFieldRequired() {
+    global.CM_Exception_FormFieldValidation.apply(this, arguments);
```
njam: To keep it simple, how about *not* extending `CM_Exception_FormFieldValidation` here, then this could become just:
```
window.CM_Exception_FormFieldRequired  = CM_Exception.extend('CM_Exception_FormFieldRequired')
```

library/CM/Form/Abstract.js
```diff
@@ -206,7 +206,7 @@ var CM_Form_Abstract = CM_View_Abstract.extend({
 
     this.resetErrors();
     if (_.size(errorList)) {
```
njam: So these are *only* errors about required fields? Then let's rename this variable to `errorListRequired`? And also the function where we get the errors from?

library/CM/Exception/FormFieldRequired.js
vogdb: I would gladly do it but here I'm just mirroring the previous logic we had. Those errors should be handled by default for `{handleErrors: true}` case. The only handled errors by default are  `CM_Exception_FormFieldValidation` errors. That is why I'm inheriting from this class. If I inherit from `CM_Exception` directly, those errors won't handled for `{handleErrors: true}` and bring `unhandledRejection`.

njam: Makes sense, let's do the inheritance.
And we cannot use `CM_Exception_FormFieldValidation.extend('CM_Exception_FormFieldRequired')`?

vogdb: Well, if we add a method `extend` to `CM_exception.prototype` and adjust it a bit. Currently this method is static and always extends from `CM_Exception`.

vogdb: hm, extending of `CM_Exception.prototype` won't work. My brain fails me again... We rather need to add a second argument of parent class to `CM_Exception.extend`.

vogdb: I think such idea deserves a separate PR :)

vogdb:
@njam please review

library/CM/Form/Abstract.js
```diff
@@ -256,6 +256,13 @@ var CM_Form_Abstract = CM_View_Abstract.extend({
     if (displayErrors) {
       this._displayValidationError(error);
     }
+    var isRequired = error instanceof CM_Exception_FormFieldRequired;
```
njam: I found this function very difficult to read.
It seems it has one part which is for "required" errors, and one which is for "validation" errors. Maybe it would be easier to split these two use cases into two functions after all? And catch the two exceptions separately, and call the appropriate function?

vogdb:
@njam the split on separate methods leads to code duplication cause I need to catch both of exceptions in usual submit and autosave. I've introduced `_handleSubmitError` in order to exclude code duplication.
please review.

njam:
As discussed - let's simplify it!
- _submitOnly: remove options
- Remove `_stopErrorPropagation` mechanism (conversation should display error below field)
- CM_FormField_Password: remove setting the field "visible" on submit, thus removing the event binding on `error`. Needs to be tested.
- Combine `_handleValidationError` and `_handleRequiredError`?
- Move triggering `error` to `submit()`?

vogdb:
> - CM_FormField_Password: remove setting the field "visible" on submit, thus removing the event binding on `error`. Needs to be tested.

Chrome PW behaves differently for the case when password is visible (type='text'). It offers to save password only once. If you ignore its offer then you won't see it ever again on any repeated login. On the other hand when password is hidden (type='password'), it offers to save password on any repeated login. The save of visible password works fine.

Firefox PW does not offer to save password at all when it's visible :( We have to keep this code.

@njam please review.

library/CM/Form/Abstract.js
```diff
@@ -176,53 +173,44 @@ var CM_Form_Abstract = CM_View_Abstract.extend({
...
-          this.disable();
-        }
         this.trigger('submit', [data]);
         return cm.ajax('form', {viewInfoList: this.getViewInfoList(), actionName: action.name, data: data})
       })
       .then(function(response) {
         if (response.errors) {
           var error = new CM_Exception_FormFieldValidation();
           error.setErrorList(response.errors);
+          this.trigger('error error.' + actionName, error.message, error.name, error.isPublic);
```
njam: Are these parameters we pass needed (message, name, isPublic)? If not, let's drop them!

library/CM/Form/Abstract.js
```diff
@@ -176,53 +173,44 @@ var CM_Form_Abstract = CM_View_Abstract.extend({
...
       })
       .catch(CM_Exception_FormFieldValidation, function(error) {
-        this._handleValidationError(error, actionName, options.handleErrors);
+        this._displayValidationError(error);
       });
   },
 
   /**
    * @param {String} [actionName]
-   * @param {Boolean} [disableUI]
```
njam: `disableUI` was defaulting to `true`, but you removed it.
I can also see we use it for "autosave", and for some form-preview in Denkmal. Let's keep it as a parameter on `_submitOnly`? - Denkmal will use this function then.

vogdb: good catch, now this param is `false` by default. Should I change it to `true`?

vogdb:
@njam please review

njam:
lgtm

`disableUI` is always passed, so we can just make it mandatory.

PR #2444, opened 2016-12-07
reload of popOut Component.

vogdb:
Firstly, when we do `component.popOut`, we do smth like this:
```js
this.$el.floatbox(options);
var self = this;
this.$el.one('floatbox-close', function() {
    self.remove();
});
```
if component gets reloaded, the new loaded replacement does not get such listener `.one('floatbox-close')` and when floatbox is closed the replacement component is kept under `cm.window._$hidden`.

Secondly, we use `replaceWithHtml` to load pages/layouts:
```js
replaceWithHtml: function($html) {
  this.remove(true);
  this.$el.replaceWith($html);
},
```

This method uses `this.remove(true)` which does not remove `component.$el` and presumes that all children's `$el` are inside parent's `$el` but that is not true for cases when children is in floatbox. So, in general we allow for component to be outside of its parent but `this.remove(true)` requires all children to be inside of their parents. How should we proceed in general? If I fix the bug of `'floatbox-close'` the problem might easily come up in the future with some custom usecase.

### DIFF
--- a/client-vendor/after-body/jquery.floatbox/jquery.floatbox.js
+++ b/client-vendor/after-body/jquery.floatbox/jquery.floatbox.js
@@ -105,6 +105,7 @@
       if (this.$parent.length) {
         this.$parent.append($element);
       }
+      this.$floatbox.trigger('floatbox-close');
       $element.removeData('floatbox');
       this.$layer.remove();
       $viewport.children('.floatbox-layer:last').addClass('active');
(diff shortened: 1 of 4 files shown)


vogdb:
@njam I was wrong about only single usage of `popOut` in `popOutComponent`. As you can see there are usages as the one in `CM_Component_Example`. There are 2 similar usages in the project *that should be called here*. Not sure about this argument `removeOnClose` anymore. The previous condition `if (cm.window.isHidden(self.el)) {` does not seem so incorrect anymore.
wdyt?

njam:
I thought we would default `removeOnClose` to `false`.
Then pass it from `popOutComponent()`, which I thought will be the only usage where we pass `true` - is it correct?

vogdb:
You are right. It's my error of research. Please review.

njam:
lgtm

PR #2449, opened 2016-12-07
Streamline CM_View_Abstract.replaceWithHtml

vogdb:
See the reason here #2444

The solution is to swap lines:
```js
  replaceWithHtml: function($html) {
    this.$el.replaceWith($html);
    this.remove();
  },
```
and test it carefully


vogdb:
tested, works fine.
@njam please review

njam:
Is there still a use case of `skipDomRemoval`? Otherwise we can remove it?

vogdb:
sgtm, @njam please review

njam:
lgtm

PR #2451, opened 2016-12-07
Remove error formfield icon

christopheschwyzer:
It's not needed, because we additionally have an error message.
And it's in conflict with other form field UI

And Make formfield success icon transparent


christopheschwyzer:
@vogdb please review

PR #2455, opened 2016-12-08
Move requiredFields from form-action to form

njam:


njam:
@vogdb and @tomaszdurka would you mind reviewing?
I'm moving the definition of *required fields* from the action to the form. This will require some adjustments in SK.
The advantage is that we know which fields are required when we render the form.

library/CM/Form/Abstract.js
```diff
@@ -286,30 +259,38 @@ var CM_Form_Abstract = CM_View_Abstract.extend({
...
+   * @returns {{}}
    */
-  _getAction: function(actionName) {
-    var action = this._actions[actionName];
-    if (!action) {
-      throw new CM_Exception('Form `' + this.getClass() + '` has no action `' + actionName + '`.');
-    }
-    action.name = actionName;
-    return action;
+  _getData: function() {
```
vogdb: What is the difference of this method with its public counterpart `getData`?

library/CM/FormAction/Abstract.php
```diff
@@ -51,16 +51,6 @@ public function getForm() {
     }
```
vogdb: Why do we still keep `CM_FormAction_Abstract->_getRequiredFields`?

tomaszdurka: Also do we use `getFieldList` still?

tomaszdurka: It's probably used (need to look into the code in IDE) - saw all the usages in diff removed, thought it might be not.

library/CM/Form/Abstract.js
njam: Ha! What will we do without you?

njam:
Thanks for the reviews! Those were oversights and have been adjusted.
I take it that you agree with the general direction, and will start adjusting usages.

njam:

library/CM/Form/Abstract.js
```diff
@@ -8,17 +8,19 @@ var CM_Form_Abstract = CM_View_Abstract.extend({
...
+  actionNames: null,
 
   /** @type Object **/
-  _actions: {},
+  _fields: {},
 
   initialize: function() {
     CM_View_Abstract.prototype.initialize.call(this);
 
     this._fields = {};
```
vogdb: why do we need to assign `_fields` to `{}` again in `initialize`?

njam: Otherwise it's on the prototype *I think* and will be shared across instances.

njam:
@tomaszdurka it's ready. We need to decide whether to go with either:
- A: `_getRequiredFields()` on the form (current solution)
- B: Pass a param `required` to the form field's constructor

wdyt?

njam:
We discussed to keep it on the form (A).

PR #2435, opened 2016-12-08
Remove Process forking IPC

alexispeter:
Currently when forking a child process, that process will communicate its success/failure status back to the parent process through an IPC stream.

Since we never send any "workload result" from a child to a parent process we could replace this IPC stream by just relying on the exit code of the child (0 = success, other = failure). This will simplify a lot of code.

@tomaszdurka agree?

### DIFF
--- a/library/CM/Process/WorkloadResult.php
+++ b/library/CM/Process/WorkloadResult.php
@@ -2,41 +2,29 @@
 
 class CM_Process_WorkloadResult {
 
-    /** @var mixed */
-    private $_result;
-
-    /** @var CM_ExceptionHandling_SerializableException|null */
-    private $_exception;
+    /** @var int|null */
+    private $_returnCode;
 
     /**
-     * @return mixed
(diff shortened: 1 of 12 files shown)


njam:
@alexispeter how is this coming along? would be great to release it soon to get rid of those IPC errors!

alexispeter:
This is ready for review, just wanted to double-check the diff.

library/CM/Process/WorkloadResult.php
```diff
@@ -2,48 +2,36 @@
 
 class CM_Process_WorkloadResult {
```
njam: How about renaming this to just `Result` now? I see this more as a wrapper to store the result of a process now, which is not necessarily related to the "workload" concept.

In the future this could also be used to spawn "shell commands" (which we currently do with `CM_Util::exec()`.

njam:
lgtm

PR #2456, opened 2016-12-08
refactor site mocking

fvovan:
The idea is to register the type before `getMockForAbstractClass()` call. So type will be available in `CM_Class_Abstact::__constructor` . It's needed for https://github.com/cargomedia/cm/pull/2428


fvovan:
hmm. that does not help a lot, we also use `$this->getMockForAbstractClass('CM_Site_Abstract')`
will try injecting settings directly to  constructor.

PR #2458, opened 2016-12-09
CM_FormField_Suggest trigger `change` when value is dropped

vogdb:


vogdb:
This feature shows itself on autosave forms with CM_FormField_Suggest. `change` event does not happen when a correct value is changed to en empty one.
@zazabe please review.

PR #2459, opened 2016-12-09
Throttle autosave of CM_Form_Abstract

vogdb:


vogdb:
@zazabe please review. Not sure still. It might be better to use here a usual `_.throttle`, wdyt?

zazabe:
I think it's good to stick with this solution (cancel previous request + validate + save data during typing).
Using `_.throttle` will produce less requests, but we sill need to add some logic to cancel the leading request... 
Also, `_.throttle` will introduce some delay in the execution of the second call I guess.

or we could do `_throttle(promiseThrottler(...), delay)`?

vogdb:
double throttle? XD No :) Lets stick to this one then.

PR #2460, opened 2016-12-12
FB-113 Site settings in DB - stage 1

fvovan:
We need to split it into 2 parts in order migrate existing configuration properly.
This PR will contain schema changes, model classes and other functionality to allow migration, example application data and settings editor.
The second one will actually _use_ these settings


fvovan:
I'm not quite sure I chose best naming conventions here.  `CM_Site_SiteSettings` looks similar to existing `CM_Site_SiteFactory` but may be it's not the best option.
The same is for `configuration` field of table. Besides it turned out we can't avoid using autoincrement ID value for models, so `siteId` was created. Also we discussed `classType` column creating, but now I doubt how it could be used considering `CM_Site_Abstract->getId()` is equal by definition to `CM_Site_Abstract->getType()`.

@njam , @tomaszdurka could you have a look please while i'm preparing the second part?

library/CM/Site/Abstract.php
```diff
@@ -264,6 +240,50 @@ public function equals(CM_Comparable $other = null) {
...
+            $siteList[] = new $className();
+        }
+        return $siteList;
+    }
+
+    /**
+     * @param $type
+     * @return string|null
+     */
+    public static function findClassName($type) {
```
njam: Seems to be unused

library/CM/Site/SiteSettings.php
```diff
@@ -0,0 +1,126 @@
...
+     * @param CM_Params $configuration
+     */
+    public function setConfiguration(CM_Params $configuration) {
+        $this->_set('configuration', CM_Params::jsonEncode($configuration->getParamsEncoded()));
+    }
+
+    /**
+     * @return int
+     */
+    public function getConfigurationSize() {
```
njam: Unused?

library/CM/Site/SiteSettings.php
```diff
@@ -0,0 +1,126 @@
...
+     * @param string $name
+     */
+    public function setName($name) {
+        return $this->_set('name', $name);
+    }
+
+    /**
+     * @return string|null
+     */
+    public function findSiteClassName() {
```
njam: Unused?

library/CM/Site/SiteSettings.php
```diff
@@ -0,0 +1,126 @@
...
+    }
+
+    /**
+     * @return CM_Params
+     */
+    public function getConfiguration() {
+        if (!$this->_has('configuration')) {
+            return CM_Params::factory([]);
+        }
+        $paramsEncoded = CM_Params::jsonDecode($this->_get('configuration'));
```
njam: Use CM_Util::jsonDecode()  instead

library/CM/Site/SiteSettings.php
```diff
@@ -0,0 +1,126 @@
...
+        return CM_Params::factory($paramsEncoded, true);
+    }
+
+    /**
+     * @param CM_Params $configuration
+     */
+    public function setConfiguration(CM_Params $configuration) {
+        $this->_set('configuration', CM_Params::jsonEncode($configuration->getParamsEncoded()));
```
njam: Use CM_Util::jsonEncode() instead

library/CM/Site/SiteSettings.php
```diff
@@ -0,0 +1,126 @@
...
+    /**
+     * @return string|null
+     */
+    public function findSiteClassName() {
+        return CM_Site_Abstract::findClassName($this->getSiteId());
+    }
+
+    protected function _getSchema() {
+        return new CM_Model_Schema_Definition([
+            'siteId'        => ['type' => 'int', 'optional' => true],
```
njam: Why is it optional?

library/CM/Site/SiteSettings.php
```diff
@@ -0,0 +1,126 @@
...
+            'name'          => ['type' => 'string', 'optional' => true],
```
njam: Make the name mandatory?

library/CM/Site/SiteSettings.php
```diff
@@ -0,0 +1,126 @@
...
+        ]);
+        $siteSettings->commit();
+        return $siteSettings;
+    }
+
+    /**
+     * @param int $siteId
+     * @return CM_Site_SiteSettings|null
+     */
+    public static function findBySiteId($siteId) {
```
njam: How about keeping the same ID values for "site" and "site-settings"?
Then this would not be needed?

library/CM/Util.php
```diff
@@ -592,6 +592,27 @@ public static function jsonDecode($value) {
 
     /**
      * @param string $value
+     * @return bool
+     */
+    public static function jsonIsValid($value) {
```
njam: Unused?

fvovan:
this was simplified.
@njam please review

library/CM/Site/Abstract.php
```diff
@@ -264,6 +240,50 @@ public function equals(CM_Comparable $other = null) {
...
+        $this->_themes = array('default');
+        return $this;
+    }
+
+    /**
+     * @return CM_Site_Abstract[]
+     */
+    public static function getAll() {
+        $siteList = array();
+        foreach (CM_Config::get()->CM_Site_Abstract->types as $className) {
```
njam: iirc we wanted this to use the database table, so that we could potentially have database-only sites in the future?

library/CM/Site/Abstract.php
njam: This function is only used in `findSiteClassName()`, which could be using just `get_class()` and is itself unused. Can this be removed?

library/CM/Site/SiteSettings.php
```diff
@@ -0,0 +1,129 @@
...
+     */
+    public function setSiteType($siteType) {
+        return $this->_set('siteType', $siteType);
+    }
+
+    /**
+     * @return CM_Params
+     */
+    public function getConfiguration() {
+        if (!$this->_has('configuration')) {
```
njam: Is this necessary? The field is mandatory, so it should always be present?

library/CM/Site/SiteSettings.php
```diff
@@ -0,0 +1,129 @@
...
+     */
+    public function setConfiguration(CM_Params $configuration) {
+        $this->_set('configuration', CM_Util::jsonEncode($configuration->getParamsEncoded()));
+    }
+
+    /**
+     * @param string $key
+     * @param string $value
+     */
+    public function upsertConfigurationValue($key, $value) {
```
njam: What is the use case for this?

library/CM/Site/SiteSettings.php
```diff
@@ -0,0 +1,129 @@
...
+    /**
+     * @return string|null
+     */
+    public function findSiteClassName() {
+        return CM_Site_Abstract::findClassName($this->getSiteType());
+    }
+
+    protected function _getSchema() {
+        return new CM_Model_Schema_Definition([
+            'siteType'      => ['type' => 'int', 'optional' => true],
```
njam: Why is it optional?

library/CM/Site/SiteSettings.php
```diff
@@ -0,0 +1,129 @@
...
+    public static function create($siteType = null, $name, CM_Params $configuration = null) {
+        if (null !== $siteType) {
+            $siteType = (int) $siteType;
+        }
+        $name = (string) $name;
+        if (null === $configuration) {
+            $configuration = CM_Params::factory([]);
+        }
+        $siteSettings = new self();
+        $siteSettings->_set([
```
njam: Let's use the setters instead, this way we can avoid duplicating some code (` jsonEncode()`).

resources/db/structure.sql
```diff
@@ -282,6 +282,18 @@ CREATE TABLE `cm_session` (
...
 ) ENGINE=MyISAM DEFAULT CHARSET=utf8;
 
+DROP TABLE IF EXISTS `cm_site_setting`;
+
+
+CREATE TABLE `cm_site_settings` (
+  `id` int(10) unsigned NOT NULL AUTO_INCREMENT,
+  `siteType` int(10) unsigned DEFAULT NULL,
+  `name` varchar(32) NOT NULL,
+  `configuration` varchar(2000) NOT NULL,
```
njam: I think we should make it larger. We might potentially store a lot of strings in here. And since this table is small and this field is not indexed it doesn't matter for performance. 65535?

resources/db/structure.sql
```diff
@@ -282,6 +282,18 @@ CREATE TABLE `cm_session` (
...
+  PRIMARY KEY (`id`),
+  UNIQUE KEY `siteType` (`siteType`)
+) ENGINE=MyISAM DEFAULT CHARSET=utf8;
```
njam: Should we use InnoDB for this new table where performance considerations don't matter?

tests/library/CM/Paging/SiteSettings/AllTest.php
```diff
@@ -0,0 +1,19 @@
+<?php
+class CM_Paging_SiteSettings_AllTest extends CMTest_TestCase {
```
njam: newline

tests/library/CM/Paging/SiteSettings/AllTest.php
```diff
@@ -0,0 +1,19 @@
...
+    }
+
+    public function testAll() {
+        $paging = new CM_Paging_SiteSettings_All();
+        $this->assertSame(0, $paging->getCount());
+
+        $settingsAdded = CM_Site_SiteSettings::create(1, 'Baz', CM_Params::factory(['bar' => 'foo']));
+        $paging = new CM_Paging_SiteSettings_All();
+        $this->assertSame(1, $paging->getCount());
+        $settings = $paging->getItem(0);
```
njam: A more concise way of asserting pagings is:
```php
$this->assertEquals([$item1, $item2], $paging);
```

njam:
I think it would be good to implement getters and setters on the site class(es) for the different fields, e.g. `getSupportEmail()`, and have a strict schema for them.
Doing that *in the `SiteSettings` model* would be great, as we could re-use the existing schema mechanism, but it would mean that we'd need to have separate `SiteSettings` models in SK and in CM.

Alternatively we could just implement getters and setters with casting on the `Site` model, which would access the data on the `SiteSettings` with generic functions accessing the data bag.
@tomaszdurka @fvovan wdyt?

library/CM/Site/Abstract.php
fvovan: yes, you're right, but now we need to keep it in order to perform settings migration.

library/CM/Site/SiteSettings.php
fvovan: it's connected to the fact that sometimes we create settings before even have any site class. Mostly it happens in tests. Yep, code intended only for tests is obviously a code smell. And i'm not quite sure how best to proceed here. In fact the only purpose of `cm_site_setting.siteId` is mapping for `CM_Site_Abstract::getAll()`. That's why maybe it would be a good idea to have "not persisted setting" class. So theoretically we'll be able to create sites with settings which are not in the DB.

resources/db/structure.sql
fvovan: I set it to 21800. (maximum value for utf-8 is 21844 but for some reason it breaks tests)

library/CM/Site/SiteSettings.php
fvovan: Ahh. good catch

njam:
As discussed: maybe it's better to make the *Site* extend *Model* so we can reuse schema and persistence functionality.
Vladimir will prepare a new branch with how the final logic could look in CM. Then we can compare it and decide which way to go.

fvovan:
was done in https://upsource.cargomedia.ch/cm/review/CM-CR-29

PR #2461, opened 2016-12-12
Fix textinput form elements

christopheschwyzer:
Make text input fields `display: block`. 
screen shot 2016-12-12 at 14 01 00


christopheschwyzer:
@njam only second commit contains changes :)

njam:
If you say it should be block, then it should be block.
lgtm.

christopheschwyzer:
🙏 amen

PR #2462, opened 2016-12-12
Adjust Select2 fix

christopheschwyzer:
Follow-up https://github.com/cargomedia/cm/pull/2445


christopheschwyzer:
When I think about it. Not even sure this makes sense. Maybe original fix was better after all. @njam please reopen should you disagree.

njam:
Ah, I thought we would move *all* select2 related styles to a separate stylesheet.
This seems reasonable because we also want to style select2 on the page if it is *not* inside a Suggest formfield.
wdyt?

christopheschwyzer:
I think this is the case already. In `Suggest/default.less` there are only styles directly related to the _Suggest_ form field. Or am I missing something?

njam:
As discussed: let's move it to a global `select2.less` and rename the relevant classes from `.suggestItem-something` to for example `.select2-item-something`.

christopheschwyzer:
@njam please review

njam:
lgtm

PR #2467, opened 2016-12-13
Introduce SliderRange formfield

njam:


christopheschwyzer:
@njam 
- `Slider.js` should inherit from `SliderAbstract.js`?! Currently files are almost identical. 
- Key input only works for singular slider.

njam:
SliderAbstract - yes, my bad! fixed!

About key input: I restricted key event handling to the case when there's only 1 handle. It would add additional complexity to do it for the other cases, and I thought it's not so important. It's a pity it's not supported by NoUiSlider itself.
I would skip it for now, wdyt?

christopheschwyzer:
ok. lgtm.

NicolasSchmutz:
The behaviour works good. I could only test it in the simulator for iOS. Two thing:
- Firefox: The progress bar behind the button is moving faster if you change the slider with the keyboard.
- iOS simulator: If one slider-cyrcle goes under the other you sometimes  lose focus, then you start scrolling the page instead.

In general the operation is not super convenient on mobile:
- My finder is bigger than the slider-cyrcle so you are never sure if you selected it or not unless you select, move your finger down and then slide. How about the slider-cyrcle gets bigger once selected on mobile?
- Also I don't see the numbers while sliding as a right-handed person. How about displaying the number before (or on top of) the slider? I guess we should go in favour of right-hander: http://www.uxmatters.com/mt/archives/2013/02/how-do-users-really-hold-mobile-devices.php

njam:
I pushed a version with tooltips (based on [noUiSlider's option](https://refreshless.com/nouislider/slider-options/#section-tooltips)). I can imagine it's better on mobile. But it requires more height.
wdyt?
image

NicolasSchmutz:
better IMO, so you also realise when you move the slider.

njam:
I think so too! Did you test it? Is it okay?

@christopheschwyzer maybe you have some additional thoughts?

NicolasSchmutz:
>I think so too! Did you test it? Is it okay?

No, can test it on Monday.

christopheschwyzer:
tooltips sgtm

NicolasSchmutz:
Much better!
- maybe the box around the number could be red as well when selected?

Still an issue:
- Firefox: The progress bar behind the button is moving faster if you change the slider with the keyboard.

- - -

>iOS simulator: If one slider-cyrcle goes under the other you sometimes  lose focus, then you start scrolling the page instead.

works now!

njam:
> Firefox: The progress bar behind the button is moving faster if you change the slider with the keyboard.

I'm not sure exactly what you mean. What I notice is that if I press down the left or right arrow key for longer (so that key repeat starts), then sometimes the "handle" moves to the edge of the slider quicker than the "background". Is this what you mean?

> maybe the box around the number could be red as well when selected?

Done.


christopheschwyzer:
lgtm

PR #2466, opened 2016-12-13
Fix CM_View_Abstract.replaceWithHtml

vogdb:
There is a problem that events are not triggered on `$el` that was replaced. Consider this example:
```html
<head></head>
<body>
  <div class="container">
    <ul class="list">
      <li></li>
      <li></li>
      <li></li>
    </ul>
  </div>
  <script src="./client-vendor/after-body/10-jquery/jquery.js"></script>
  <script src="./client-vendor/after-body/20-underscore/underscore.js"></script>
  <script>
    $(document).ready(function() {
      $('li').on('hello', function() {
        console.log('hello');
      });
      var $ul = $('ul');
      $ul.replaceWith('<p>new content</p>');
      $ul.trigger('hello');
    });
  </script>
</body>
```

`hello` event is not triggered. Hence `this.remove();` from `replaceWithHtml` does not trigger any destruction events. Working on it now.

### DIFF
--- a/library/CM/View/Abstract.js
+++ b/library/CM/View/Abstract.js
@@ -226,14 +226,8 @@ var CM_View_Abstract = Backbone.View.extend({
    * @param {jQuery} $html
    */
   replaceWithHtml: function($html) {
-    var $parent = this.$el.parent();
-    var $next = this.$el.next();
-    cm.window.appendHidden(this.$el);
-    if ($next.length) {
-      $next.before($html);
-    } else {
-      $parent.append($html);
-    }
+    var parent = this.el.parentNode;
+    parent.replaceChild($html[0], this.el);
     this.remove();
   },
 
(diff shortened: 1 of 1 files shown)


njam:
> Hence this.remove(); from replaceWithHtml does not trigger any destruction events. Working on it now.

What kind of destruction events? We have the `destruct` event on `CM_View_Abstract`, but it's triggered on the view instance, so it should work even if there's no DOM Element anymore?

The changes you made would move the old Element to an invisible area, then insert the new Element, then remove the old View. If we do that we could also just remove the old View (and its Element), and then add the new Element.
The idea of `replaceWithHtml()` was to do it in a almost atomic operation, to avoid layout reflows.

vogdb:
> What kind of destruction events? We have the `destruct` event on `CM_View_Abstract`, but it's triggered on the view instance, so it should work even if there's no DOM Element anymore?

Sorry for misleading name. I am talking about jQuery events here. If jQuery element is removed with `replaceWithHtml` then it or any of its children elements loose all their event listeners. Please look at the example in the topic comment with `hello` event. It's very descriptive. Feel free to call me when you have a moment and discuss.

vogdb:
@njam please review. `Node.replaceChild` works. Nice finding!

njam:
You could use `this.el`, otherwise lgtm

PR #2469, opened 2016-12-15
Store autosave fields before submit CM_Form_Abstract

vogdb:


njam:
@zazabe could you review?

zazabe:
yes

library/CM/Form/Abstract.js
```diff
@@ -14,15 +14,24 @@ var CM_Form_Abstract = CM_View_Abstract.extend({
...
+      })
       .then(function() {
-        field.success();
+        this._autosaveFields.forEach(function(field) {
+          field.success();
+        });
+        this._autosaveFields = [];
       })
       .catch(CM_Exception_FormFieldValidation, function(error) {
         this._displayValidationError(error);
```
zazabe: what about clearing `this.autosaveFields` here too, to avoid keeping invalid field in this list?

library/CM/Form/Abstract.js
```diff
@@ -63,7 +72,8 @@ var CM_Form_Abstract = CM_View_Abstract.extend({
     if (this.autosave) {
       this.on('change', function(field) {
         if (field) {
-          handler._autosaveSubmitThrottled(field);
+          this._autosaveFields.push(field);
```
zazabe: maybe use  either `handler` or `this` for both statements?

vogdb:
@zazabe please review

PR #2472, opened 2016-12-18
Fix passing of options to Audio constructor from createAudio()

njam:


njam:
There was a bug that resulted in `options` not being passed along correctly.
I'm wondering if we should drop this function altogether though. It's only two lines, and probably more readable calling them directly. @zazabe wdyt?

zazabe:
you mean something like that?
```js
var audio = new cm.lib.Media.Audio();
audio.getElement().loop = true;
```
Agree, will work also. 
I see only one advantage with `media.setOptions()`, having the documentation.

njam:
I meant instead of:
```js
var audio = cm.dom.createAudio('foo.mp3');
```
to do:
```js
var audio = new cm.lib.Media.Audio();
audio.setSource('foo.mp3');
```

And remove `createAudio()`. wdyt?

zazabe:
ah! agree! ;)

njam:
On second thought it's also good to have a single function to create an `<audio>` element for playback, which atomically fails or succeeds. I would suggest to keep `createAudio()` and fix how options are passed. It's done in https://github.com/cargomedia/cm/pull/2471

PR #2471, opened 2016-12-19
Media.js: don't set "crossOrigin" by default

njam:
Currently we set `crossOrigin` to `anyonymous` when creating new audio or video elements in media.js:
```
    this._options = _.defaults(options || {}, {
      loop: false,
      autoplay: false,
      crossOrigin: 'anonymous'
    });
```

This is useful only when we want to use the created media in a `<canvas>`:
> CORS-enabled resources can be reused in the <canvas> element without being tainted. 
https://developer.mozilla.org/en/docs/Web/HTML/Element/video#Attributes

In all other cases it's better not to set the attribute, otherwise the elements gets tainted (when CORS doesn't match), and can't be played.

I would suggest to *not* set `crossOrigin` by default, and pass it in those cases where we use the element in a `<canvas>`. @zazabe wdyt?

### DIFF
--- a/client-vendor/source/cm/media.js
+++ b/client-vendor/source/cm/media.js
@@ -12,14 +12,12 @@ var Media = Event.extend({
    * @param {Object} [options]
    * @param {Boolean} [options.loop=false]
    * @param {Boolean} [options.autoplay=false]
-   * @param {String} [options.crossOrigin='anonymous']
    */
   constructor: function(element, options) {
     this._element = element;
     this._options = _.defaults(options || {}, {
       loop: false,
-      autoplay: false,
-      crossOrigin: 'anonymous'
+      autoplay: false
     });
 
     this._sources = [];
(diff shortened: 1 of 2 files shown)


njam:
@vogdb please review

I will adjust in SK accordingly.

njam:
or @zazabe please review

zazabe:
It's maybe even better to remove `crossOrigin` from default value, because:
> If invalid, it is handled as if the enumerated keyword anonymous was used
https://developer.mozilla.org/en/docs/Web/HTML/Element/video#attr-crossorigin

And I'm not sure if doing `element.crossOrigin = null;` will not result in an invalid `crossOrigin` value (should be tested), wdyt?

njam:
Setting it to `null` works in Chrome. But you're right, it's safer to *not* set it.

I've also fixed passing `options` along in `createAudio()` (https://github.com/cargomedia/cm/pull/2472).

Please review

njam:
@zazabe could you take over, and adjust SK usages when merged?

zazabe:
yep

PR #2476, opened 2016-12-29
Introduce CM_Iterator_Transform / CM_Iterator_Permissive

zazabe:
Inspired by @tomaszdurka work :)

### DIFF
--- a/library/CM/Iterator/Permissive.php
+++ b/library/CM/Iterator/Permissive.php
@@ -1,0 +1,66 @@
+<?php
+
+class CM_Iterator_Permissive extends IteratorIterator {
+
+    /** @var Closure */
+    private $_transformer;
+
+    /** @var mixed */
+    private $_current;
+
+    /**
+     * @param Traversable|array $traversable
+     * @param Closure           $transformer
+     */
+    public function __construct($traversable, Closure $transformer) {
+        if (is_array($traversable)) {
+            $traversable = new ArrayIterator($traversable);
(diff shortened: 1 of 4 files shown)


zazabe:
@tomaszdurka please review

njam:
Is this to have a *lazy* map function?

Could this also be written with *generators*?
```php
/**
 * @param array|Iterator $collection
 * @param callable       $function
 * @return Iterator
 */
function map($collection, callable $function) {
    foreach ($collection as $element) {
        yield $function($element);
    }
}
```
(via http://stackoverflow.com/questions/6566129/php-lazy-array-mapping/32099190#32099190)

zazabe:
@njam yes, something like that, but in a generic iterator that wrap the iteration logic, looks a bit more OOP I guess, not sure if it's worth it...
Wondering how we could achieve the same thing with `yield`, 
- does it mean we should we have something like `CM_Utils:transform(...)` / `CM_Utils:permissive(...)`?
- or simply using `yield` specifically every time we want to map an array with some classes, with some extra logic if necessary (like for this "permissive" stuff).

njam:
How to proceed here?

zazabe:
Not sure if it's really needed, solved with `yield` in https://github.com/cargomedia/cm/pull/2477/files#diff-cf4749258410d777ac80766d5d8808c7R37, I would suggest to close it..

njam:
👍

Issue #2403, opened 2016-11-01
Allow to include "description" in migration file name


njam:
Goal: allow to include a *description* in the name of a migration script ("34-some-change" instead of "34"), to avoid VCS conflicts.

This was discussed before in https://github.com/cargomedia/cm/issues/588

The new migration script could extend `CM_Provision_Script_Abstract` and use `CM_Provision_Script_OptionBased` for persisting whether a script has ran or not.
-> Or it could use a separate table, which can be used in more powerful ways in the future (keep timestamps etc).

The filename should still include a timestamp for ordering. For example the file could be named `1478003870-some-change.php` and the class `CM_Provision_Migration_1478003870-some-change`.

We should introduce a CLI helper to create a new migration. The *description* could by default be taken from the GIT branch name.

cc @tomaszdurka @dlondero @fauvel

zazabe:
- Let's use a separate table, maybe auto-generated like described [here](https://github.com/cargomedia/cm/issues/588#issuecomment-78231730)
- I would suggest to always generate the migration class, with a reference to the previous migration script.  
  Then, we could add a CI script to test if there's multiple migration classes with a ref to the same migration script.

```
CM_Provision_MigrationScript extends CM_Model_Abstract
  - filename: string
  - classname: string
  - previous: CM_Provision_MigrationScript|null 
  - execStamp: DateTime  
```

wdyt?

njam:
> previous migration ... test if there's multiple migration classes with a ref to the same migration script

My understand was that this is what we want to allow to do. Two developers should be able to commit a migration script to master both succeeding the same previous migration. In most cases that's not a problem, as there were not conflicts in the code.

@tomaszdurka wdyt?

njam:
As discussed:
- Reuse `CM_Provision_Loader` but add a custom persistence layer for migrations (extending `CM_Provision_Script_Abstract`).
- Add new folder `db/migration/` for the new system, so we can introduce it side-by-side.
- CM_App::runUpdateScripts(): Dropping of test database not needed anymore, because `CM_Db_SetupScript` will drop DB in tests.

fauvel:
An old discussion indeed:) Summary:

The idea is to make a trade-off between:
(a) controlling the execution order of the migration scripts merged from different branches, and
(b) avoiding unnecessary merge conflicts when devs create simultaneously independent migration scripts in different branches.

[...]

PR #2477, opened 2016-12-29
Implement more flexible migration scripts

zazabe:
original issue: #2403
previous discussion: #588 

~_Note: `CM_Iterator_*` are coming from https://github.com/cargomedia/cm/pull/2476_~


library/CM/Migration/Loader.php
```diff
@@ -0,0 +1,54 @@
+<?php
+
+class CM_Migration_Loader implements CM_Service_ManagerAwareInterface {
+
+    const MIGRATION_DIR = 'migration';
```
tomaszdurka: Passing directories into constructor would solve potential problem of this constant directory name.

library/CM/Migration/Loader.php
```diff
@@ -0,0 +1,54 @@
...
+            return $this->_getMigrationScript($script);
+        });
+    }
+
+    /**
+     * @param CM_File $script
+     * @return CM_Migration_Script
+     */
+    protected function _getMigrationScript(CM_File $script) {
+        $className = sprintf('CM_Migration_Script_%s', $script->getFileNameWithoutExtension());
```
tomaszdurka: What about non-CM scripts?

I had similar problems when implementing migrations some time ago. I did checked all classes loaded- if they have been declared within the migration directories (and subclass of Migration) then they are valid script classes and should be considered.
image

library/CM/Migration/Cli.php
```diff
@@ -0,0 +1,61 @@
+<?php
+
+class CM_Migration_Cli extends CM_Cli_Runnable_Abstract {
+
+    public function all() {
```
njam: How about combining this into 1 command with an optional argument?
```
bin/cm migration run --name=foo
```

library/CM/Migration/Cli.php
```diff
@@ -0,0 +1,61 @@
...
+
+    /**
+     * @param string $name
+     */
+    public function load($name) {
+        $loader = new CM_Migration_Loader($this->getServiceManager());
+        if ($script = $loader->findScript($name)) {
+            $this->_loadScript($script);
+        } else {
+            $this->_getStreamError()->writeln(sprintf('Migration script "%s" not found', $name));
```
njam: This should make the command exit with a non-zero status code. At the moment I think this can be only achieved by throwing an exception.

library/CM/Migration/Cli.php
```diff
@@ -0,0 +1,61 @@
...
+    protected function _loadScript(CM_Migration_Script $script) {
+        $output = $this->_getStreamOutput();
+        $output->write(sprintf('- %s', $script->getName()));
+        if ($desc = $script->getDescription()) {
+            $output->write(sprintf(': %s', $desc));
+        }
+        try {
+            $script->load();
+        } catch (Exception $e) {
+            $output->write(" \e[31m×\e[0m\n");
```
njam: I'm not a big fan of colors for CLI commands. Quite often I experience problems with it: colors are not rendered or "strange" characters are rendered, for example because of some configuration of the terminal, or because of an SSH tunnel etc.
For example I'm not sure this will work fine in pulsar-rest-api where this code will be executed and the output stored.

If we do it, then I think it should be possible to turn it off, and maybe it should be turned off automatically if there's no TTY.

In this case it might not be worth it. How about we use only the checkmark/cross characters?

library/CM/Migration/Generator.php
```diff
@@ -0,0 +1,92 @@
...
+
+    /** @var CM_File */
+    private $_file;
+
+    /**
+     * @param string      $name
+     * @param string|null $namespace
+     */
+    public function __construct($name, $namespace = null) {
+        $this->_name = $name;
```
njam: It's never used because it's stored in $_fileName, so no need to store it.

library/CM/Migration/Generator.php
```diff
@@ -0,0 +1,92 @@
...
+        $this->_namespace = $namespace;
+        $this->_fileName = sprintf('%s_%s', time(), CM_Util::camelize(trim($name)));
```
njam: Also no need to store it, as we can access it from `$_file->getFileName()`?

library/CM/Migration/Generator.php
```diff
@@ -0,0 +1,92 @@
+<?php
+
+class CM_Migration_Generator {
+
+    const MIGRATION_CLASSNAME = 'CM_Migration_Script';
```
njam: Maybe rename to something like "classNamePrefix"

library/CM/Migration/Generator.php
```diff
@@ -0,0 +1,92 @@
...
+            throw new CM_Exception_Invalid('Class `' . $className . '` already exists');
+        }
+
+        $class = new CodeGenerator\ClassBlock($className, self::MIGRATION_CLASSNAME);
+        $reflection = new ReflectionClass(self::MIGRATION_CLASSNAME);
+        $method = CodeGenerator\MethodBlock::buildFromReflection($reflection->getMethod('up'));
+        $method->setAbstract(false);
+        $method->setDocBlock(join(PHP_EOL, [
+            '/**',
+            ' * Describe the migration script',
```
njam: Also add "TODO:"

library/CM/Migration/Loader.php
```diff
@@ -0,0 +1,54 @@
...
+    public function __construct(CM_Service_Manager $serviceManager) {
+        $this->setServiceManager($serviceManager);
+    }
+
+    /**
+     * @param string $name
+     * @return CM_Migration_Script|null
+     */
+    public function findScript($name) {
+        return \Functional\first($this->getScriptList(), function (CM_Migration_Script $script) use ($name) {
```
njam: Theoretically there can be multiple scripts with the same name, but in different namespaces. Such a scenario is problematic, and maybe we should throw an exception in that case?

library/CM/Migration/Loader.php
njam: Maybe we should detect the classname from the file?
Would be very flexible, but seems not super easy:
http://stackoverflow.com/questions/928928/determining-what-classes-are-defined-in-a-php-class-file

library/CM/Migration/Model.php
```diff
@@ -0,0 +1,85 @@
...
+     */
+    public function getExecStamp() {
+        return $this->_has('execStamp') ? $this->_get('execStamp') : null;
+    }
+
+    /**
+     * @param DateTime $execStamp
+     * @return $this
+     */
+    public function setExecStamp($execStamp) {
```
njam: add type hinting

library/CM/Migration/Model.php
```diff
@@ -0,0 +1,85 @@
...
+     * @return bool
+     */
+    public function hasExecStamp() {
+        return null !== $this->getExecStamp();
+    }
+
+    protected function _getSchema() {
+        return new CM_Model_Schema_Definition(array(
+            'name'      => ['type' => 'string'],
+            'execStamp' => ['type' => 'DateTime', 'optional' => true],
```
njam: I would rename this to "executedAt", because:
- No abbreviations
- No mentioning of variable type in name

I know we use this "somethingStamp" in many places, but I think we should move away from it.

library/CM/Migration/Model.php
```diff
@@ -0,0 +1,85 @@
...
+    }
+
+    /**
+     * @param string $name
+     * @return CM_Migration_Model
+     */
+    static public function findByName($name) {
+        $cache = CM_Cache_Local::getInstance();
+        if (false === ($migrationId = $cache->get($name))) {
+            $migrationId = CM_Db_Db::select(self::getTableName(), 'id', array('name' => $name))->fetchColumn();
```
njam: For better portability you could use the storage adapter's `findByData()`- [example: `Location::findByAttributes()`](https://github.com/cargomedia/cm/blob/master/library/CM/Model/Location.php#L203).

library/CM/Migration/Script.php
```diff
@@ -0,0 +1,65 @@
...
+    }
+
+    abstract public function up();
+
+    /**
+     * @param CM_OutputStream_Interface|null $output
+     */
+    public function load(CM_OutputStream_Interface $output = null) {
+        $this->up();
+        $this->_getRecord()->setExecStamp((new DateTime())->setTimestamp(time()));
```
njam: What is `setTimestamp()`?

library/CM/Migration/Loader.php
zazabe: well, if the classname is based on the filename, we should not have the same script name multiple time.
Thought about filtering by namespace (like with the current `cm db run-update`), but it makes not much sense in this case...

njam: True, didn't notice that.

library/CM/Iterator/Permissive.php
```diff
@@ -0,0 +1,66 @@
+<?php
+
+class CM_Iterator_Permissive extends IteratorIterator {
```
zazabe: it is, but those `CM_Iterator_*` are actually parts of another PR (https://github.com/cargomedia/cm/pull/2476), and  `CM_Iterator_Permissive` is something I may re-use [here](https://github.com/zazabe/sk/blob/0d16e1bfaaaff7b84c8302ebe7e4ad2605cf9389/library/SK/library/SKService/AmateurCommunity/CallbackIterator.php) in https://github.com/cargomedia/sk/pull/4172 ;)

zazabe: I can remove it and wait for something like `CM_Iterator_Transform` when #2476 will be merged...

library/CM/Migration/Loader.php
zazabe: - why should we care about non-CM scripts?  
- why not just basing it on naming conventions like it's done now? 

I don't see the use case, is `CM_Migration_Script` not flexible enough? 
I agree to introduce `CM_Migration_ScriptInterface` and check if `$className` implements it, but i don't think we need a more complex solution...

library/CM/Migration/Script.php
zazabe: not sure why... thought i copy/past it from somewhere... maybe i was afraid about the [leap second 2017](https://community.ntppool.org/t/leap-second-2017-status/59)!

library/CM/Migration/Script.php
```diff
@@ -0,0 +1,65 @@
...
+    }
+
+    /**
+     * @return string|null
+     */
+    public function getDescription() {
```
zazabe: I thought it's nice to optionally get this from the docu, but we are not used to add description to our docu and maybe overriding `getDescription` in the migration script is more conventional...

library/CM/Migration/Script.php
```diff
@@ -0,0 +1,65 @@
...
+        $reflector = new ReflectionClass($this);
+        $doc = $reflector->getMethod('up')->getDocComment();
+        if ($doc && preg_match('/\*[ ]+([\S ]+)/', $doc, $matches)) {
+            return $matches[1];
+        }
+        return null;
+    }
+
+    public function getName() {
+        return str_replace('CM_Migration_Script_', '', parent::getName());
```
zazabe: you meant to use internals?

library/CM/Migration/Script.php
zazabe: ho... I have maybe an explanation... seems `new DateTime()` is not taking in account `CMTest_TH::timeInit()` overwrites...

njam: True. We should use this [libfaketime](https://github.com/wolfcw/libfaketime) instead. But another time.

library/CM/Migration/Loader.php
zazabe: finally, i'll follow your suggestion @tomaszdurka, could be nice to support any filename at the end ;)

zazabe:
Almost everything has been done, few notes:
- no more naming convention between filename and classname    
  I use a [similar technic](https://gist.github.com/tomaszdurka/6277999ae74f602b85c3be1019801e46#file-migrations-php-L116) than @tomaszdurka to get the classname from the last included file, seems to work quite nicely ;)    
The downside is that we can have duplicated filename in different namespaces now...   
It will only affect `cm migration run` with the `--name` argument I think, so maybe we could simply display a warning if duplicated scripts are detected, wdyt @njam?
- `CM_Migration_Model`: seems handy to me
- optionally use the docu as the script description: I can remove it if you insist ;)

library/CM/Migration/Model.php
```diff
@@ -0,0 +1,103 @@
...
+     * @param array $data
+     * @return static|null
+     */
+    public static function findByAttributes(array $data) {
+        $cache = CM_Cache_Local::getInstance();
+        $cacheKey = CM_CacheConst::Migration_ByAttribute;
+        foreach ($data as $fieldName => $fieldValue) {
+            $cacheKey .= '_name:' . $fieldName . '_value:' . $fieldValue;
+        }
+        if (false === ($id = $cache->get($cacheKey))) {
```
njam: Could be simplified to?:
```php
        $cacheKey = $cache->key(__METHOD__, $data);
        $id = $cache->get($cache->key(__METHOD__, $data), function() {
            return // todo: get id
        });
```

library/CM/Migration/Model.php
```diff
@@ -0,0 +1,103 @@
...
+        return static::findByAttributes([
+            'name' => $name,
+        ]);
+    }
+
+    /**
+     * @param array $data
+     * @return static|null
+     */
+    public static function findByAttributes(array $data) {
```
njam: Maybe this could be moved to CM_Model_Abstract, not sure.

library/CM/Migration/Model.php
```diff
@@ -0,0 +1,103 @@
...
+        $cache = CM_Cache_Local::getInstance();
+        $cacheKey = CM_CacheConst::Migration_ByAttribute;
+        foreach ($data as $fieldName => $fieldValue) {
+            $cacheKey .= '_name:' . $fieldName . '_value:' . $fieldValue;
+        }
+        if (false === ($id = $cache->get($cacheKey))) {
+            /** @var CM_Model_StorageAdapter_Database $persistence */
+            $persistence = self::_getStorageAdapter('CM_Model_StorageAdapter_Database');
```
njam: Use `self:: getPersistenceClass()` to get the persistence class name.

library/CM/Migration/Generator.php
```diff
@@ -0,0 +1,90 @@
...
+    /** @var string */
+    private $_parentClassName;
+
+    /** @var CM_File_Filesystem */
+    private $_filesystem;
+
+    /**
+     * CM_Migration_Generator constructor.
+     * @param CM_File_Filesystem $filesystem
+     * @param null               $parentClassName
```
njam: Docu

library/CM/Migration/Loader.php
```diff
@@ -0,0 +1,101 @@
...
+
+    /**
+     * @param CM_File $file
+     * @return CM_Migration_Script
+     * @throws CM_Exception_Invalid
+     */
+    protected function _prepareScript(CM_File $file) {
+        $className = $this->_requireScript($file->getPathOnLocalFilesystem());
+        $reflector = new ReflectionClass($className);
+        if (!$reflector->isSubclassOf(CM_Migration_Script::class)) {
```
njam: Could we also use `instanceof` or `is_subclass_of()` to not use reflection?

zazabe:
@tomaszdurka please review

library/CM/Migration/Loader.php
```diff
@@ -0,0 +1,100 @@
...
+            require_once($filePath);
+            $classesAfter = get_declared_classes();
+            $diff = array_diff($classesAfter, $classesBefore);
+            if (count($diff) !== 1) {
+                throw new CM_Exception_Invalid('Migration script must declare only one class', null, [
+                    'declaredClasses' => $diff,
+                    'filePath'        => $filePath,
+                ]);
+            }
+            $this->_loadedClasses[$filePath] = \Functional\first($diff);
```
tomaszdurka: Not sure about if this may not generate problems. E.g. if there is a class declaration which inherits from other, other one might be loaded as first one. That's why in my linked implementation I check for the file path to directory match. 

@njam suggested idea with reading file contents and it's token. This sounds good for me. We could read all the class tokens and check if they inherit Migration script, if yes let's add them. This way we have possibility (completely unneeded) to have more than one migration script in single file.

README.md
```diff
@@ -234,35 +234,37 @@ Commands:
  app deploy
  app fill-caches
  app generate-config-internal
- app set-deploy-version [--deploy-version=<value>]
+ app set-config <filename> <config-json> [--merge]
```
zazabe: i copy/pasted the `bin/cm` output, this readme has not been updated for a long time, so no, not all are part of this changeset ;)

library/CM/Migration/Cli.php
```diff
@@ -0,0 +1,85 @@
...
+        }
+    }
+
+    /**
+     * @param string|null $namespace
+     * @param string|null $name
+     */
+    public function add($namespace = null, $name = null) {
+        if (null === $name) {
+            $name = CM_Util::exec('git rev-parse --abbrev-ref HEAD');
```
zazabe: stdin with `$this->_getStreamInput()->read()`? 
but it means we must pipe the name like `echo "foo-bar" | ./bin/cm migration add`, isn't it? also it's blocking if nothing is piped...

...or you're taking about something else!

tomaszdurka: Im talking about this, but I think this is perfectly fine to be blocking. It's action done by user.

There is default value for read.
```php
if (null === $name) {
  $defaultName = CM_Util::exec('git rev-parse --abbrev-ref HEAD');
  $name = $this->_getStreamInput()->read('Name the script', $defaultName);
}
```

README.md
tomaszdurka: lgtm

library/CM/Migration/Cli.php
tomaszdurka: As you can see in `CommandManager` the default input is 
```php
$this->_setStreamInput(new CM_InputStream_Readline());
```
We only set it different when non-interactive parameter is passed:
```php
if ($nonInteractive) {
  $this->_setStreamInput(new CM_InputStream_Null());
}
```

Probably it would be better to pass something else then `Null` one, but that's the mechanics.

zazabe:
Major changes:
- script must implements `CM_Migration_UpgradableInterface` only, no `CM_Migration_Script` dependency anymore
- rename `CM_Migration_Script` > `CM_Migration_Runner`   
  holds the migration instance to execute and keeps tracking of previously executed migration
- generated migration classes are prefixed by `Migration_`, not `CM_`, to avoid confusion
- about @tomaszdurka comment  https://github.com/cargomedia/cm/pull/2477#discussion_r94756950    
   one and only one class implementing the `CM_Migration_UpgradableInterface` is picked between all declared classes during the `require`, i think it's safe and flexible enough, wdyt?


@tomaszdurka please review again!

library/CM/Migration/Cli.php
zazabe: I didn't know that we have this interactive prompt as the stdin! works well, thanks ;)

library/CM/Migration/Loader.php
```diff
@@ -0,0 +1,104 @@
...
+        if (!isset($this->_loadedClasses[$filePath])) {
+            $classesBefore = get_declared_classes();
+            require_once($filePath);
+            $classesAfter = get_declared_classes();
+            $classesDiff = array_values(array_diff($classesAfter, $classesBefore));
+            $classesUpgradable = \Functional\filter($classesDiff, function ($className) {
+                return in_array(CM_Migration_UpgradableInterface::class, class_implements($className));
+            });
+            if (count($classesUpgradable) !== 1) {
+                throw new CM_Exception_Invalid('Migration script must declare one and only one class implementing CM_Migration_UpgradableInterface', null, [
```
tomaszdurka: Again-again, I think it might be problem if we introduce common parent migration script classes.

CM/Migration/SpecificFeatureSetScript.php
```php
class CM_Migration_SpecificFeatureSetScript implements CM_Migration_UpgradableInterface {
  public function specialFeature() {}
}
```

resources/migrations/121232132131_FooScript.php
```php
class FooScript extends CM_Migration_SpecificFeatureSetScript {}
```

This will result in 2 upgradable classes and exception with a bit invalid message.

zazabe: Right, what about checking the file path also, as you did?

zazabe: what about bb9d347?

zazabe:
@tomaszdurka as discussed, let's use `token_get_all`!

library/CM/Migration/Loader.php
```diff
@@ -0,0 +1,106 @@
...
+     */
+    protected function _requireScript($filePath) {
+        if (!isset($this->_loadedClasses[$filePath])) {
+            $classesBefore = get_declared_classes();
+            require_once($filePath);
+            $classesAfter = get_declared_classes();
+            $classesDiff = array_values(array_diff($classesAfter, $classesBefore));
+            $classesUpgradable = \Functional\filter($classesDiff, function ($className) use ($filePath) {
+                $reflectionClass = new \ReflectionClass($className);
+                return false !== strpos($reflectionClass->getFileName(), $filePath)
```
tomaszdurka: Note: I've used strpos and not `=` here also to make "eval" work. Somehow filename of eval-created method contains the filename from which is been evaled and some trailing additional text like "eval line 28".

zazabe: This code has changed already 😛

library/CM/Migration/Cli.php
```diff
@@ -0,0 +1,86 @@
...
+
+    /**
+     * @param string|null $name
+     */
+    public function run($name = null) {
+        $paths = $this->_getMigrationPaths();
+        $loader = new CM_Migration_Loader($this->getServiceManager(), $paths);
+        if (null === $name) {
+            foreach ($loader->getRunnerList() as $runner) {
+                if ($runner->shouldBeLoaded()) {
```
zazabe: > We are using CM_Provision_Script_Abstract for `load` and `shouldBeLoaded`, but then we are adding `up` on top of it.

Not really, `up` is only implemented by the migration script, not by `CM_Migration_Runner`.
The `CM_Migration_Runner` is compliant with the `CM_Provision_Script_Abstract` i think.

I don't see where is the advantage of creating a completely new structure...
About "use load/unload and extend/implements CM_Provision_Script_Abstract", I don't get it, you mean using `CM_Provision_Script_Abstract` for migration scripts themselves, like before?

tomaszdurka:
@zazabe please let me know if/when I should review.

tests/library/CM/Migration/GeneratorTest.php
```diff
@@ -0,0 +1,61 @@
...
+        $path = $file->getPathOnLocalFilesystem();
+
+        $this->assertSame(sprintf('%s_FooBarBaz.php', $time), $file->getFileName());
+        $this->assertSame(join(PHP_EOL, [
+            '<?php',
+            '',
+            sprintf('class Migration_%s_FooBarBaz implements \CM_Migration_UpgradableInterface {', $time),
+            '',
+            '    /**',
+            '     * TODO: Describe the migration script',
```
zazabe: hum... what about keeping it here for now? we could have diff message for `down` in the future or moving it into the class doc...

zazabe:
as discussed, i removed the `CM_Provision_Script_Abstract` dependency from `CM_Migration_Runner`, so, please @tomaszdurka, kindly review this PR :)

library/CM/Migration/Loader.php
```diff
@@ -0,0 +1,126 @@
...
+     * @return CM_File[]
+     */
+    protected function _getFiles() {
+        return $this->_files;
+    }
+
+    /**
+     * @return CM_File[]
+     */
+    protected function _prepareFiles($paths) {
```
tomaszdurka: Let's call it a bit more explicit.

`findFilesWithinDirectories`?

library/CM/Migration/Loader.php
```diff
@@ -0,0 +1,126 @@
...
+        }
+        return $files;
+    }
+
+    /**
+     * @param CM_File $file
+     * @return CM_Migration_Runner
+     * @throws CM_Exception_Invalid
+     */
+    protected function _prepareRunner(CM_File $file) {
```
tomaszdurka: instantiateRunner

zazabe:
@tomaszdurka please check

zazabe:
@tomaszdurka mergeable?

tomaszdurka:
I think it's fine more-less :) still having bit different view on the internals. 
Additionally I think we should have smaller initial pull-request with only script mechanism, but without generators and cli.

zazabe:
"I think it's fine more-less" == approved @tomaszdurka ?

tomaszdurka:
Heh, yes it's approved. I am just saying (like pain in the a...) that I would probably go for a bit different approach, but I don't have monopoly on good solutions - unfortunately! This one is good too! :)

@njam maybe you wanna add something before merge?

njam:
Acceptance testing:
- Weird argument name "space": `migration add [--namespace=<value>] [--name=<space>]`
- It's a pity we can only use `echo()` to print any status updates. How about we make a `CM_OutputStream_Interface` instance available? Like in `CM_Provision_Script_Abstract`.
- If a script outputs anything (like progress), then that output interferes with the status output. How about printing the status output on two lines? `Running 1484828516_Foo: here is the description...\n` and `Done.\n`.
- Generating a script that already exists overrides the previous one. Maybe rather throw an error?
- Default migration template: I would remove the "todo" docu for `up()`. Most of the time it will not be needed because the code is self explanatory. wdyt?
- Running `migration run` gives the error `Call to undefined method CM_Migration_Runner::shouldBeLoaded() in /home/vagrant/cm/library/CM/Migration/Cli.php on line 13`.  It's sub-optimal that the IDE doesn't know the iteration type of `$loader->getRunnerList()` and can't warn you when calling a function on each item. Is there a better way to document this?

zazabe:
@njam all done, please look again

njam:
- Created template migration has an extra line before the `up()` function
- Tests fail

Otherwise works great!

zazabe:
@tomaszdurka approved?

PR #2479, opened 2017-01-03
Don't use Array.includes() due to browser support

njam:


njam:
@fauvel please review

PR #2480, opened 2017-01-03
Add "head-top" block to the default document template

zazabe:


layout/default/View/Document/default.tpl
```diff
@@ -2,6 +2,7 @@
 <html {if $render->getLanguage()}lang="{$render->getLanguage()->getAbbreviation()}"{/if} class="{$viewResponse->getCssClasses()|implode:' '} {block name='html-class'}{/block}" id="{$viewResponse->getAutoId()}" {if isset($webFontLoaderConfig)}data-web-font-loader='{$webFontLoaderConfig}'{/if}>
   <head>
     <meta charset="utf-8">
+    {block name='head-top'}{/block}
```
njam: The convention is `-start` and `-end`, so this should be `head-start`. Feel free to merge when adjusted.

PR #2483, opened 2017-01-05
Use standard "ResponsePage::process()" also for "Embed" response

njam:
Addressing https://github.com/cargomedia/sk/issues/3264

Instead of a custom `_process()` implementation for the *embed* page response (ajax navigation), we'll use the standard one:
```php
    protected function _process() {
        $this->setHeaderDisableCache();
        $this->_site->preprocessPageResponse($this);
        $this->_processContentOrRedirect();
        if ($redirectUrl = $this->getRedirectUrl()) {
            $this->setRedirectHeader($redirectUrl);
        }
    }
```

Those headers that are set on the response will be ignored silently by `loadPage()`.
But the call to `preprocessPageResponse()` is important, as the Site might decide to redirect the user elsewhere.

@fauvel please review

### DIFF
--- a/library/CM/Http/Response/Page/Embed.php
+++ b/library/CM/Http/Response/Page/Embed.php
@@ -26,10 +26,6 @@ class CM_Http_Response_Page_Embed extends CM_Http_Response_Page {
         return $renderAdapterPage->fetch();
     }
 
-    protected function _process() {
-        $this->_processContentOrRedirect();
-    }
-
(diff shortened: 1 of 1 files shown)


njam:
The `return` is there to leave the function early. Usually if a redirect happens we don't want to process further, as then we might override the previous redirect. It's not very nice, but works.

fauvel:
I mean, we should split it up into `$response->redirect($redirectPage); return;`

PR #2487, opened 2017-01-06
Add specific exceptions for invalid entity types and location levels

alexispeter:
cm-changes for https://github.com/cargomedia/sk/issues/4355

### DIFF
--- a/library/CM/Model/Entity/Abstract.php
+++ b/library/CM/Model/Entity/Abstract.php
@@ -50,9 +50,15 @@ abstract class CM_Model_Entity_Abstract extends CM_Model_Abstract {
      * @param int $type
      * @param int $id
      * @return CM_Model_Entity_Abstract
+     * @throws CM_Exception_Invalid
+     * @throws CM_Class_Exception_TypeNotConfiguredException
      */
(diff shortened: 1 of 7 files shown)


library/CM/Model/Entity/Abstract.php
```diff
@@ -50,9 +50,14 @@ public function getUserId() {
      * @param int $type
      * @param int $id
      * @return CM_Model_Entity_Abstract
+     * @throws CM_Entity_InvalidTypeException
      */
     final public static function factory($type, $id) {
         $className = self::_getClassName($type);
+        $calledClass = get_called_class();
+        if (!is_subclass_of($className, $calledClass)) {
```
njam: Could we do `$className instanceof static`?

njam: …and to print `static::class`?

alexispeter: `instanceof` only works with objects as far as I know.

alexispeter: You could replace `get_called_class()` with `static::class` if there was any reason for doing so.

Issue #2482, opened 2017-01-05
Cannot json_encode value of request URL


njam:
JSON encoding of the request's URL fails in the GoogleAnalytics client, when we try to encode the request's URI:
```php
$js .= 'ga("send", "pageview", ' . CM_Params::jsonEncode($pageView) . ');';
```

The error code from `json_last_error` is 5 (`JSON_ERROR_UTF8`).
The string that's being encoded (request URI) looks like this:
- [Loggly](https://cargomedia.loggly.com/search#terms=%22Cannot%20json_encode%20value.%22): `/users/friends1008(13) ?? - ...`
- [New Relic](https://rpm.newrelic.com/accounts/350429/applications/2137721/filterable_errors#/table?top_facet=transactionUiName&primary_facet=error.message&barchart=barchart&filters=%5B%7B%22key%22%3A%22error.class%22%2C%22value%22%3A%22CM_Exception_Invalid%22%2C%22like%22%3Afalse%7D%2C%7B%22key%22%3A%22error.message%22%2C%22value%22%3A%22Noticed%20exception%20%27CM_Exception_Invalid%27%20with%20message%20%27Cannot%20json_encode%20value.%27%20in%20%2Fhome%2Ffuckbook%2Freleases%2F20170104144933%2Fvendor%2Fcargomedia%2Fcm%2Flibrary%2FCM%2FUtil.php%3A559%22%2C%22like%22%3Afalse%7D%5D&_k=9obyk3): `/users/friends1008(13) 信息 - ...`

Probably the truth lies inbetween.

We *are* sanitizing the request URL in `CM_Http_Request_Abstract::__construct()` using `CM_Util::sanitizeUtf()`:
```php
mb_convert_encoding($value, 'UTF-8', 'UTF-8');
```
This seems this doesn't cover all cases.

Unfortunately I am not able to reproduce so far with the example strings from above.

---

Example stack trace:
```
#0 /home/example/releases/20170104144933/vendor/cargomedia/cm/library/CM/Util.php(559): {throw}
#1 /home/example/releases/20170104144933/vendor/cargomedia/cm/library/CM/Params.php(669): CM_Util->jsonEncode('/users/friends1008(13) ??_? ...', NULL)
#2 /home/example/releases/20170104144933/vendor/cargomedia/cm/library/CMService/GoogleAnalytics/Client.php(130): CM_Params->jsonEncode('/users/friends1008(13) ??_? ...')
[...]

PR #2488, opened 2017-01-09
Log non UTF request variables

fvovan:
followup https://github.com/cargomedia/cm/issues/2482

There's a new case :(
```
    [sanitizedFields] => Array
        (
            [code] => 434d5f5574696c2d3e6a736f6e456e636f646528272f702f4e30656d69313030394e30656d6920e75f5fe4b8aae4babae8b55f2e2e2e272c204e554c4c29
            [value] => 272f702f4e30656d69313030394e30656d6920e75f5fe4b8aae4babae8b55fe65f5f202d204675636b626f6f6b27
        )
```

Sanitizing the string *again* makes it work.
I would suggest that we detect this case and log it, *before* sanitizing it.
wdyt?
Could you add the debug code?

### DIFF
--- a/library/CM/Http/Request/Abstract.php
+++ b/library/CM/Http/Request/Abstract.php
@@ -63,7 +63,14 @@ abstract class CM_Http_Request_Abstract {
             }
             $this->_server = array_change_key_case($server);
         }
-        $uri = CM_Util::sanitizeUtf((string) $uri);
+        $uri = (string) $uri;
+        if (!mb_check_encoding($uri, 'UTF-8')) {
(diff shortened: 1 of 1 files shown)


fvovan:
is it okay that we log only uri values, since all wrong input goes from it?
@njam wdyt?

njam:
Agree, that's enough.
But I would be stricter with the condition, to make sure we *only* catch those cases that lead to errors now. Basically if json encoding the *sanitized* value leads to an error. wdyt?

fvovan:
good point!
moreover we'll see how it was transformed, hopefully this won't affect performance much

njam:
lgtm

PR #2491, opened 2017-01-10
Use league/uri to manage URLs

zazabe:
Introduce `CM\Url` namespace, with:
- `UrlInterface`, which implements [`Psr\Http\Message\UriInterface`](http://www.php-fig.org/psr/psr-7/#psrhttpmessageuriinterface)
- `AbsoluteUrl`
- `RelativeUrl`
- `ResourceUrl`

following #2494


zazabe:
pass basic acceptance tests, lot of `TODO` still

zazabe:
I'm wondering if we should not remove the `popPathPart(0)` call in `CM_Http_Request_Abstract::popPathLanguage()`, wdyt @njam?

njam:
We remove (pop) the language part of a URL to make further processing easier.
A URL is initially `/fr/foo`, after popping it's just `/foo`, which is easier to use in `CM_Http_Response_Abstract::createFromRequest()`.

zazabe:
oh! this commit was for #2494 ;)
 I think it could be managed differently with Url objects, but anyway, not in this PR.

zazabe:
@njam please review

library/CM/Url/AbstractUrl.php
```diff
@@ -0,0 +1,27 @@
+<?php
+
+namespace CM\Url;
+
+use Psr\Http\Message\UriInterface;
+use League\Uri\Schemes\Generic\AbstractHierarchicalUri;
+
+abstract class AbstractUrl extends AbstractHierarchicalUri implements UrlInterface {
+
+    public function getRebaseUrl(UrlInterface $baseUrl) {
```
njam: Maybe turn it around and make it `withRelativeComponentsFromAppended`

library/CM/Url/RelativeUrl.php
```diff
@@ -0,0 +1,46 @@
...
+
+    public function withEnvironment(\CM_Frontend_Environment $environment, array $options = null) {
+        $options = array_merge([
+            'sameOrigin' => false,
+        ], (array) $options);
+
+        $site = $environment->getSite();
+        $url = $site->getUrlBase();
+        if (!$options['sameOrigin'] && $site->getUrlCdn()) {
+            $url = $site->getUrlCdn();
```
njam: We don't want to use the CDN for all relative URLs. For example page URLs can be created like this, but are not served from the CDN. Either this should be in a separate URL type, or configurable with an option?

library/CM/Url/ResourceUrl.php
```diff
@@ -0,0 +1,52 @@
...
+            ->append((string) \CM_App::getInstance()->getDeployVersion())
+            ->append((string) $this->path);
+    }
+
+    /**
+     * @param string|null $uri
+     * @param string|null $type
+     * @return ResourceUrl
+     */
+    public static function createFromString($uri = null, $type = null) {
```
njam: Why are both arguments optional? Shouldn't they be mandatory?

tests/library/CM/Url/AbsoluteUrlTest.php
```diff
@@ -0,0 +1,80 @@
+<?php
+
+use CM\Url\AbsoluteUrl;
+
+class CM_Url_AbsoluteUrlTest extends CMTest_TestCase {
```
njam: Use namespace and make classname `AbsoluteUrlTest`

library/CM/Url/AbstractUrl.php
zazabe: maybe we should just drop this, seems strange to concat paths but preserve query/segments... not easily understandable.

we could introduce the concept of path "prefix" instead, like:
```php 
$url = new RelativeUrl::createFromString('/foo?val=1');
$prefixUrl = $url->withPathPrefix('bar');            // /bar/foo?val1
$prefixUrl->withPath('/foz');                        // /bar/foz?val1
$prefixUrl->withPathPrefix('baz')->withPath('/foz'); // /baz/foz?val1
```

njam: sgtm

library/CM/Url/ResourceUrl.php
zazabe: the parent use this signature: 
```php
public static function createFromString($uri = '')
```
which is equivalent to `(string) $uri` here

library/CM/Url/RelativeUrl.php
zazabe: Right, it should be used only for `ResourceUrl`

library/CM/Url/ResourceUrl.php
njam: Empty $type still looks strange, it will be valid but just use '', is it okay?

zazabe: I would simply make it required then, agree, looks strange to call `somethingFromString()` with no string... :)

zazabe: Ooops, incompatible with the parent. ok for `$url = ''` then...

zazabe:
I tried to summarize all types of URLs we use and the type of objects that could reflex them:

```php

- route         <site>/[prefix]/<route-path>/[language]  (ie: ajax, form, etc...)
- page:         <site>/[prefix]/[language]/<page-path>
- worker        <site>/[prefix]/serviceworker[-language]-<deploy-version>.js
- static:   <cdn|site>/static/<file-path>?<deploy-version>
- resource: <cdn|site>/<type>/[language]/<site-id>/<deploy-version>/<file-path>

abstract AbstractUrl
 - isAbsolute()
 - withEnvironment($env, $options = null)
 - withBaseUrl($url)           // authority + prefix if available           
 - withLanguage($language)    
 - abstract _getUriRelativeComponents() // full relative url with all components
 + language                    // optional
 + prefix                      // optional

RouteUrl extends AbstractUrl
 - _getUriRelativeComponents() // [/prefix/][routePath][/language][paramsQuery][fragment]
 - withParams($params) 
 + params

PageUrl extends RouteUrl
 - _getUriRelativeComponents() // [/prefix/][language/][pagePath][paramsQuery][fragment]

ServiceWorkerUrl extends AbstractUrl
 - _getUriRelativeComponents() // [/prefix/][name][-language]-[deployVersion].js
 + name
 + deployVersion
 
abstract AssetUrl extends AbstractUrl
 - withEnvironment($env, $options = ['sameOrigin' => false])
 
StaticFileUrl extends AssetUrl
 - _getUriRelativeComponents() // /static/[filePath][query + <deployVersion>][fragment]

ResourceFileUrl extends AssetUrl
 - _getUriRelativeComponents() // [type][/language]/[siteId][/deployVersion]/[filePath][query][fragment]
 + site
 + type

  
RouteUrl::create($route, $params = null, $baseUrl = null, $language = null);
PageUrl::create($className, $params = null, $baseUrl = null, $language = null);
ServiceWorkerUrl::create($name = 'serviceworker', $baseUrl = null, $language = null, $deployVersion = null);
StaticFileUrl::create($filename, $baseUrl = null, $language = null, $deployVersion = null);
ResourceFileUrl::create($filename, $site, $type, $baseUrl = null, $language = null, $deployVersion = null);
```

wdyt @njam ?

zazabe:
@njam please review

_the 2 last commits are implementing https://github.com/cargomedia/cm/pull/2491#issuecomment-275065815_

library/CM/Url/AbstractUrl.php
```diff
@@ -0,0 +1,147 @@
...
+    /**
+     * @param string                 $url
+     * @param UrlInterface|null      $baseUrl
+     * @param CM_Model_Language|null $language
+     * @return static
+     */
+    protected static function _create($url, UrlInterface $baseUrl = null, CM_Model_Language $language = null) {
+        /** @var AbstractUrl $url */
+        $url = self::getPipeline()->process(
+            parent::createFromString($url)
```
njam: self instead of parent?

zazabe:
few changes:
- introduce `BaseUrl` (useful for `$site->getUrl()`, etc...)
- better `prefix` support

library/CM/Url/AssetUrl.php
```diff
@@ -0,0 +1,71 @@
...
+     * @return AssetUrl
+     */
+    public function withSite(CM_Site_Abstract $site, $sameOrigin = null) {
+        $sameOrigin = (bool) $sameOrigin;
+        $baseUrl = $site->getUrlBase();
+        if (!$sameOrigin && $cdnUrl = $site->getUrlCdn()) {
+            $baseUrl = $cdnUrl;
+        }
+        if (!$baseUrl instanceof UrlInterface) {
+            $baseUrl = $this->_create((string) $baseUrl);
```
njam: call `_create()` on `self::`

library/CM/Url/AbstractUrl.php
```diff
@@ -0,0 +1,147 @@
...
+
+    public function withEnvironment(CM_Frontend_Environment $environment, array $options = null) {
+        $url = clone $this;
+
+        if ($language = $environment->getLanguage()) {
+            $url = $url->withLanguage($language);
+        }
+        $baseUrl = $environment->getSite()->getUrlBase();
+        if (!$baseUrl instanceof UrlInterface) {
+            $baseUrl = $this->_create((string) $baseUrl);
```
njam: call `_create()` on `self::`

library/CM/Url/AssetUrl.php
```diff
@@ -0,0 +1,71 @@
...
+        }
+        return $this->withBaseUrl($baseUrl);
+    }
+
+    /**
+     * @param string                 $filename
```
njam: Param name is confusing, because we're passing a full URL from `withSite()`. Also the parent implementation calls it `$url` - do the same here?

library/CM/Url/AbstractUrl.php
```diff
@@ -0,0 +1,147 @@
...
+
+    public function isAbsolute() {
+        return !('' === $this->getScheme() && '' === $this->getHost());
+    }
+
+    public function getLanguage() {
+        return $this->_language;
+    }
+
+    public function getPrefix() {
```
zazabe: not sure what would be a URL prefix... I would prefer to keep the interface simple, but i can change it if you think it's confusing somehow.

library/CM/Url/AbstractUrl.php
zazabe: changed by `BaseUrl::create()`, in `withBaseUrl()` directly

zazabe:
@njam please check again

zazabe:
how, just notice your comment about CDN, i agree, i'll do it

zazabe:
@njam please check

_I'll open a follow-up with the concrete usages, then we will be able to do some acceptance tests._

library/CM/Url/AbstractUrl.php
njam: sgtm, let's leave it "getPrefix"

njam:
lgtm

Let's adjust usages of `Site::getUrlCdn()` now that it's mandatory. Of course Render can be left as is, if we replace it anyhow.

zazabe:
I think it makes sense to create URLs based on the `CM_Frontend_Environment` vs `CM_Model_Language` + `$baseUrl`...
It will definitely simplify usages of those URLs in CM.

zazabe:
@njam please check again

zazabe:
@njam please hold, still doing some changes.. according to CM usages :)

zazabe:
@njam please review (see also usages in #2535)

zazabe:
@njam should we merge it? it could simplify the review of #2535

njam:
sgtm

Issue #2486, opened 2017-01-05
Class for representing a link target (URL)


njam:
There are differently "complete" representations of a URL:
- A: **Absolute URL**. Can be used directly.
- B: **Relative URL**: e.g. `/example`. Only in combination with a *Render* it can become a full URL. It can be passed to the client, and the client can convert it to a full URL.
- C: **Page Classname**: e.g. `CM_Page_Example`. Only in combination with a *Render* it can become a full URL.

Each of these representations can optionally include *parameters*.

Usages of such URL representations include:
- smarty_function_linkUrl(): param `page` is a *page classname*, param `params` is URL params
- CM_Page_Abstract::getPath(): returns a *relative URL*
- CM_MenuEntry: option `page` Is a *page classname*, option `params` is URL params
- CM_Http_Response_Page::redirect: *page classname* and URL params
- CM_Http_Response_Page::redirectUrl: *absolute URL*

I think it would make sense to wrap the representation of such URLs into an object, so we **can pass it along as one variable**.

I'm not sure all the different use cases should be wrapped into a single class. Case A) and B) are generic use cases (and already covered to some extent in `CM_Http_UrlParser`), while C) is specific to our framework. Nevertheless it would be nice if there was one type that could be passed along that represents either of the three options.

@dlondero @tomaszdurka @zazabe wdyt?

zazabe:
For A and B, we could use something like http://uri.thephpleague.com/4.0/uri/schemes/http i guess, it's [PSR-7 compliant](http://www.php-fig.org/psr/psr-7/#psrhttpmessageuriinterface), and it seems to support relative paths (host/scheme/... don't need to be specified).
Then, we can [probably extend the URI object](http://uri.thephpleague.com/4.0/uri/extension/) to support case C

zazabe:
Draft version with [league/uri](https://packagist.org/packages/league/uri) : https://github.com/cargomedia/cm/pull/2491

zazabe:
as discussed with @njam:

- introduce a new `CM\Url` namespace
- introduce interfaces: `FullUrlInterface`, `RelativeUrlInterface` and `PageUrlInterface`    
- implement those interfaces via classes inheriting from vendor URI object

tomaszdurka:
[...]

PR #2494, opened 2017-01-10
Always use the language in URLs

zazabe:
- make the language always available on env, from the url, the user pref or a default one 
- use the language in all URLs
- support edge case, when there's no language defined in `cm_model_language` (in tests only) -> language abbreviation will not be added to URLs

### DIFF
--- a/library/CM/RenderAdapter/Document.php
+++ b/library/CM/RenderAdapter/Document.php
@@ -99,9 +99,8 @@ class CM_RenderAdapter_Document extends CM_RenderAdapter_Abstract {
         $options['urlCdn'] = $this->getRender()->getSite()->getUrlCdn();
         $options['urlUserContentList'] = $serviceManager->getUserContent()->getUrlList();
         $options['urlServiceWorker'] = $this->getRender()->getUrlServiceWorker();
-        if ($this->getRender()->getLanguageRewrite()) {
-            $options['urlLanguage'] = $this->getRender()->getLanguage();
-        }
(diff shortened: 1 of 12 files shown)


zazabe:
related to https://github.com/cargomedia/cm/issues/2486

zazabe:
@njam should we redirect url without language, like `/foo` => `/en/foo`? seems not so easy, we don't know the page in https://github.com/zazabe/cm/blob/8b06bbda414b9768493d68617ae6bbc25c28329a/library/CM/Http/Response/Page.php#L70...

njam:
We should not redict. The default-language version should still be accessible, be it only for Google.

zazabe:
@njam please review

library/CM/RenderAdapter/Document.php
```diff
@@ -99,9 +99,8 @@ private function _getOptions() {
         $options['urlCdn'] = $this->getRender()->getSite()->getUrlCdn();
         $options['urlUserContentList'] = $serviceManager->getUserContent()->getUrlList();
         $options['urlServiceWorker'] = $this->getRender()->getUrlServiceWorker();
-        if ($this->getRender()->getLanguageRewrite()) {
-            $options['urlLanguage'] = $this->getRender()->getLanguage();
-        }
+        $options['urlLanguage'] = $this->getRender()->getLanguage();
```
njam: Rename this to just `language`?

njam:
lgtm

PR #2497, opened 2017-01-12
Setup output stream for log handler.

fvovan:
currently it's not aware of handlers we redefine in `CM_Cli_CommandManager` when `--quiet` is passed

### DIFF
--- a/library/CM/Cli/CommandManager.php
+++ b/library/CM/Cli/CommandManager.php
@@ -61,6 +61,7 @@ class CM_Cli_CommandManager {
         if ($quiet) {
             $this->_setStreamOutput(new CM_OutputStream_Null());
             $this->_setStreamError(new CM_OutputStream_Null());
+            CM_Bootloader::getInstance()->getExceptionHandler()->setStderr(new CM_OutputStream_Null());
         }
         if ($quietWarnings) {
             CM_Bootloader::getInstance()->getExceptionHandler()->setPrintSeverityMin(CM_Exception::ERROR);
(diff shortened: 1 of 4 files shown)


fvovan:
@njam @tomaszdurka wdyt?
i understand it's probably meaningless to have `stderr` for `CM_ExceptionHandling_Handler_Http` but it's more clear from code perspective now. otherwise we need find the way to setup specific exception handler only.

@njam we discussed that it's possible to remove `--quiet` option. but does it mean that it will be used by default then?

njam:
Discussed it with Tomasz: looks good, but let's rename the "stderr" to "output" on the exception handler. Then it makes sense for both handlers.

fvovan:
please review

fvovan:
>Http handler should also use the getOutput() function

done.

>How about setting the default outputs for both handlers in their constructors? (so getOutput() d
oesn't return null while actually there is an output)

I'm not sure about constructors. as long as we modify output via the setter may be select it in the  corresponding getter?

njam:
It's ok, but now we print to stdout instead of stderr in the CLI handler. There should be a different default for the two implementations.

fvovan:
>It's ok, but now we print to stdout instead of stderr in the CLI handler. There should be a different default for the two implementations.

actually it works just the same. I guess STDERR is redirected to STDOUT by default.
But agree let's not rely on it.

fvovan:
>You mean it looks the same? Most terminals display stdout and stderr the same way. But tools and scripts usually treat it differently.

I meant only http handler behaviour from the end user's perspective.

njam:
Ah! So php-fpm also captures stderr, insteresting.

PR #2501, opened 2017-01-17
Proper close of OpenerDropdown

vogdb:
When page switching happens by click inside OpenerDropdown, OpenerDropdown isn't closed. Memory leak is produced.

### DIFF
--- a/client-vendor/after-body/jquery.opener/jquery.opener.js
+++ b/client-vendor/after-body/jquery.opener/jquery.opener.js
@@ -6,26 +6,26 @@
   var selector = '.openerDropdown';
 
   var OpenerDropdown = function($element) {
-    this.element = $element;
+    this.$element = $element;
   };
 
   OpenerDropdown.prototype = {
     constructor: OpenerDropdown,
 
     activate: function() {
-      this.element.on('click' + selector, this.toggle);
+      this.$element.on('click' + selector, this.toggle);
     },
 
     toggle: function() {
       var self = this;
-      this.element.find('> .openerDropdown-window').toggleModal(function() {
+      this.$element.find('> .openerDropdown-window').toggleModal(function() {
         $(this).toggle();
-        self.element.toggleClass('open');
+        self.$element.toggleClass('open');
       });
     },
 
     close: function() {
-      this.element.find('> .openerDropdown-window').toggleModal('hide');
+      this.$element.find('> .openerDropdown-window').toggleModal('hide');
     }
   };
 
--- a/client-vendor/after-body/jquery.opener/jquery.opener.js
+++ b/client-vendor/after-body/jquery.opener/jquery.opener.js
@@ -15,7 +15,7 @@
     element: null,
 
     activate: function() {
-      this.element.on('click.openerDropdown', this.toggle);
+      this.element.on('click' + selector, this.toggle);
     },
 
     toggle: function() {
@@ -33,7 +33,7 @@
 
   $.fn.opener = function(action) {
     return this.each(function() {
-      var $this = $(this).closest('.openerDropdown');
+      var $this = $(this).closest(selector);
       var data = $this.data('openerDropdown');
       if (!data) {
         $this.data('openerDropdown', (data = new OpenerDropdown($this)))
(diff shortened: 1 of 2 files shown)


vogdb:
@njam FYI
@zazabe please review

client-vendor/after-body/jquery.opener/jquery.opener.js
```diff
@@ -6,36 +6,37 @@
...
-    this.element = $element;
+    this.$element = $element;
   };
 
   OpenerDropdown.prototype = {
     constructor: OpenerDropdown,
 
-    element: null,
-
     activate: function() {
```
zazabe: can't find any usages, droppable?

client-vendor/after-body/jquery.opener/jquery.opener.js
```diff
@@ -45,8 +46,8 @@
   };
 
   $(function() {
-    $('body').on('click.openerDropdown', selector + ' .openerDropdown-panel', function() {
-      $(this).closest('.openerDropdown').opener('toggle');
+    $('body').on('click' + selector, selector + ' .openerDropdown-panel', function() {
```
zazabe: potential memory leak here? maybe acceptable on `body`... not sure

client-vendor/after-body/jquery.opener/jquery.opener.js
```diff
@@ -6,36 +6,37 @@
   var selector = '.openerDropdown';
```
zazabe: `namespace` instead maybe? 
btw, is it about reducing the build size?

client-vendor/after-body/jquery.opener/jquery.opener.js
zazabe: also a bit strange to have this usage here... shouldn't it be outside, in `CM_App` for example?

client-vendor/after-body/jquery.opener/jquery.opener.js
vogdb: I guess so, cause it's a usual purpose of such variable. It was presented before and I decided to improve its coverage.

client-vendor/after-body/jquery.opener/jquery.opener.js
vogdb: > potential memory leak here?

Why leak? It's executed in the global scope once. So, no leak. Just a permanent listener on `body` ;-)

vogdb: > also a bit strange to have this usage here... shouldn't it be outside, in `CM_App` for example?

We can't setup this listener in `cm.dom.setup` cause this place might be executed many times for the same DOM cause of component nesting. Consider the case when ComponentParent contains ComponentChild. Both execute `cm.dom.setup` for their `this.$el` hence DOM of ComponentChild is being set up twice. Once for its own `cm.dom.setup`. Secondly for its parent `cm.dom.setup`. This way listener becomes unreliable cause it's being executed as many times as many components are ancestors of its belonging component.

vogdb:
@zazabe please look

client-vendor/after-body/jquery.opener/jquery.opener.js
zazabe: what about `cm.dom.ready`?

vogdb:
@zazabe please look

client-vendor/after-body/jquery.opener/jquery.opener.js
```diff
@@ -3,51 +3,43 @@
...
+        }
+        $this.data(namespace, (data = new OpenerDropdown($this)))
       }
       if ('string' === typeof action) {
         data[action]()
       }
     })
   };
 
-  $(function() {
```
njam: I don't think it's wrong to have this here. This binding is required for this UI widget to work correctly. Having everything related to "openerDropdown" in one file makes it more portable.

Compare for example with bootstrap, they also define event handlers per widget:
[dropdown.js](https://github.com/twbs/bootstrap/blob/v4-dev/js/src/dropdown.js#L277)

@zazabe @vogdb wdyt?

vogdb: Makes sense but destroying of Dropdown still has to be done in `cm.dom.teardown` for the case when the link inside OpenerDropdown loads a new page.
@zazabe ?

zazabe: thought it's more like a plugin so we should not have usages here, but it doesn't matter so much i guess...
btw, should we use `$(document)` vs `$(body')` ;) ?

zazabe: maybe we could introduce this `activate` and probably a `deactivate` method on `OpenerDropdown`, then we can call them from outside when it's appropriate?

vogdb:
@zazabe please review

PR #2506, opened 2017-01-20
Remove runkit. Add timecop.

dlondero:


njam:
btw we're removing SVM in https://github.com/cargomedia/cm/pull/2532

dlondero:
> btw we're removing SVM in #2532

That's a great news!

njam:
SVM is removed

PR #2507, opened 2017-01-20
Move invalid encoding logging

fvovan:
In order to check if `parse_url()` works well.
https://github.com/cargomedia/cm/issues/2482


njam:
Michel is working on introducing the library League/Uri for parsing and processing URLs in CM (#2486). They have their own sanitization routines.
I would suggest to put this on hold, as it might be solved by using League/Uri.
wdyt?

fvovan:
sure

njam:
lgtm, let's release it!

PR #2508, opened 2017-01-20
Update jquery-timeago to 1.5.4

vogdb:


vogdb:
@zazabe please review
@njam FYI. I removed couple of `cm.data` functions. Didn't find their usages. Do we use them? In Denkmal?

njam:
I can see you removed `cm.date.$timeago()`. There are not usages of it in Denkmal. I think it was used in SK in the past. Removing it lgtm.

PR #2509, opened 2017-01-20
Sync _valueLast when setting value programmatically in FormField_Text

vogdb:


vogdb:
@zazabe please review this super easy one-liner!

PR #2510, opened 2017-01-20
Remove language fallback logic in Environment

njam:
We have the same logic in Request::getLanguage()

@zazabe please review

### DIFF
--- a/tests/library/CM/Frontend/RenderTest.php
+++ b/tests/library/CM/Frontend/RenderTest.php
@@ -2,13 +2,9 @@
 
 class CM_Frontend_RenderTest extends CMTest_TestCase {
 
-    /** @var CM_Model_Language */
-    private $_languageDefault;
-
     protected function setUp() {
         CM_Config::get()->CM_Site_Abstract->url = 'http://www.default.dev';
         CM_Config::get()->CM_Site_Abstract->urlCdn = 'http://cdn.default.dev';
-        $this->_languageDefault = CMTest_TH::createLanguage('en');
     }
 
     public function tearDown() {
--- a/tests/library/CM/Frontend/RenderTest.php
+++ b/tests/library/CM/Frontend/RenderTest.php
@@ -31,13 +31,13 @@ class CM_Frontend_RenderTest extends CMTest_TestCase {
         $render = new CM_Frontend_Render();
         $page = $this->getMockForAbstractClass('CM_Page_Abstract', array(), 'CM_Page_Foo_Bar_FooBar', false);
 
-        $this->assertSame('http://www.default.dev/en/foo/bar/foo-bar',
+        $this->assertSame('http://www.default.dev/foo/bar/foo-bar',
             $render->getUrlPage('CM_Page_Foo_Bar_FooBar'));
-        $this->assertSame('http://www.default.dev/en/foo/bar/foo-bar',
+        $this->assertSame('http://www.default.dev/foo/bar/foo-bar',
             $render->getUrlPage($page));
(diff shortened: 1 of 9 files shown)


tests/library/CM/View/DocumentTest.php
```diff
@@ -100,7 +100,7 @@ public function testLanguageAlternatives() {
         $renderAdapter = new CM_RenderAdapter_Document($render, $page);
         $html = $renderAdapter->fetch();
 
-        $this->assertContains('<link rel="alternate" href="http://www.example.com/en/example" hreflang="x-default">', $html);
```
njam: @zazabe this is actually a bug. The "x-default" variant should link to the version that does auto-detection.

zazabe:
wondering why there was some logic to detect the language in `CM_Frontend_Render` in the past, here:
https://github.com/cargomedia/cm/blob/483b702d8a500b54df8713812080c32bdd3fb68b/library/CM/Frontend/Render.php#L349

tests/library/CM/Frontend/RenderTest.php
```diff
@@ -156,36 +155,6 @@ public function testGetUrlStaticDifferentSite() {
         $this->assertSame('http://cdn.other.com/static/foo.jpg?' . $deployVersion, $render->getUrlStatic('/foo.jpg', $site));
     }
 
-    public function testGetLanguage() {
```
zazabe: `$this->_languageDefault` could be removed from `setUp`

njam:
As discussed `Render::getTranslation()` was accessing the default language like that. Instead we'll make sure the Environment always has the default language if needed.

@zazabe please review

tests/library/CM/Mail/MailableTest.php
```diff
@@ -46,9 +46,10 @@ public function testRenderTranslated() {
         $site = $this->getMockSite(null, null, [
             'url' => 'http://www.foo.com',
         ]);
+        $language = CM_Model_Language::create('Test language', 'foo', true);
```
zazabe: should we add a test with no language for the user?

tests/library/CM/Model/UserTest.php
```diff
@@ -144,4 +144,14 @@ public function testOfflineOld() {
         $this->assertFalse($user1->getOnline());
         $this->assertTrue($user2->getOnline());
     }
+
+    public function testGetEnvironment() {
+        $user = CMTest_TH::createUser();
+        $this->assertNull($user->getEnvironment()->getLanguage());
+
+        $defaultLanguage = CMTest_TH::createLanguage('en');
+        $user = CMTest_TH::createUser();
```
zazabe: maybe add a test when a user has no language set, but he get the default one on his env?

njam: It's covered 2 lines below I think

njam:
@zazabe please review

Could you take over here and release it? We should do it soon to make sure we don't run into problems with Google indexing.

tests/library/CM/Model/UserTest.php
zazabe: not really, it should be ` $this->assertEquals($defaultLanguage, $user->getEnvironment()->getLanguage());`, but without setting `$user->setLanguage($defaultLanguage)`, as below...

zazabe:
yep!

tests/library/CM/Model/UserTest.php
zazabe: like 74a211a

njam: Makes sense

PR #2513, opened 2017-01-23
FB-107 Rework JobDistribution into service

tomaszdurka:


tomaszdurka:
@njam @alexispeter please take a look.

This pull-request is converting job-distribution into a service. Service should be an implementation of jobdistribution queue interface with 2 simple methods
- `publish($job)`
- `consume`
I've moved around gearman code to reuse old-code, but with a use of new interface. I guess implementing this in AMPQ should quite straight forward.

Before adjusting or even rewriting tests, I wanted to get your feedback on it.
Note: With current version I do not support synchronous job execution. I would introduce different interface/service for that (more like RPC? it could reuse gearman client).

library/CM/Jobdistribution/Job/Abstract.php
```diff
@@ -2,201 +2,48 @@
...
+    /** @var  CM_Params */
+    private $_params;
 
     /**
-     * @return CM_Jobdistribution_Priority
+     * @param $params
      */
-    public function getPriority() {
-        return new CM_Jobdistribution_Priority('normal');
+    public function __construct($params) {
```
njam: Add docu and type hinting for $params

library/CM/Jobdistribution/Queue.php
```diff
@@ -0,0 +1,8 @@
+<?php
+
+interface CM_Jobdistribution_Queue {
```
njam: We usually include "Interface" in the name for interfaces (QueueInterface).

library/CM/Jobdistribution/Queue.php
```diff
@@ -0,0 +1,8 @@
...
+
+    public function publish(CM_Jobdistribution_Job_Abstract $job);
```
njam: Docu

library/CM/Jobdistribution/Serializer.php
```diff
@@ -0,0 +1,58 @@
...
+     * @return CM_Jobdistribution_Job_Abstract
+     * @throws CM_Exception_Nonexistent
+     */
+    public function unserialize($data) {
+        try {
+            $workloadParams = $this->_serializer->unserialize($data);
+        } catch (CM_Exception_Nonexistent $ex) {
+            throw new CM_Exception_Nonexistent("Cannot decode workload `{$ex->getMessage()}`, workload: `${data}'", null, null, CM_Exception::WARN);
+        }
+        $jobClassName = $workloadParams['jobClassName'];
```
njam: Check the class that it is really a `CM_Jobdistribution_Job_Abstract`?

library/CM/Serializer/ArrayConvertible.php
```diff
@@ -0,0 +1,13 @@
+<?php
+
+class CM_Serializer_ArrayConvertible implements CM_Serializer_SerializerInterface {
```
njam: Nice!

library/CM/Serializer/SerializerInterface.php
```diff
@@ -0,0 +1,17 @@
+<?php
+
+interface CM_Serializer_SerializerInterface {
+
+    /**
+     * @param mixed $data
+     * @return mixed
+     */
+    public function serialize($data);
```
njam: Should it return string instead of mixed?

library/CM/Jobdistribution/Serializer.php
```diff
@@ -0,0 +1,58 @@
+<?php
+
+class CM_Jobdistribution_Serializer {
```
njam: Should maybe be called JobSerializer?

library/CM/Gearman/JobQueue.php
```diff
@@ -0,0 +1,27 @@
...
+
+class CM_Gearman_JobQueue implements CM_Jobdistribution_Queue {
+    
+    /** @var CM_Gearman_Publisher */
+    private $_publisher;
+    
+    /** @var CM_Gearman_JobWorker */
+    private $_worker;
+
+    public function __construct(CM_Gearman_Publisher $publisher, CM_Gearman_JobWorker $worker) {
```
njam: Docu

tomaszdurka:
Agree with comments, regarding JobWorker, it was moreless complete. I didn't want to rewrite all the code. Worker is reasonable name/concept, valid in gearman naming world, so kept it instead of naming it consumer or so.

Of course if we would use gearman for something else then job queue, we could have a gearman service "spawning" client/worker to fulfil the needs of job queue interface.

library/CM/Serializer/SerializerInterface.php
tomaszdurka: True, thought about and that it's obvious, but had some reasons it might be better to not set strict signature. But I guess if want to use it as dependency than it needs some strictness.

library/CM/Jobdistribution/Serializer.php
tomaszdurka: /nod

library/CM/Jobdistribution/Serializer.php
tomaszdurka: Wasn't sure if it will be only for jobs, but maybe for some other things around. Also not sure about the `_verifyParams`. I was to move it to `ArrayConvertible` (and make verification optional for performance). Wdyt?

library/CM/Jobdistribution/Queue.php
tomaszdurka: True, I'll change that. Just pushed the convention a bit, so it feels better when you request the service.
```php
$serviceManager->get(CM_Jobdistribution_Queue::class);
```

library/CM/Jobdistribution/Serializer.php
njam: Moving to `ArrayConvertible` sounds reasonable.
Not sure about disabling for performance. Maybe benchmark, and if not relevant ignore it?

tomaszdurka:
I've added RPCInterface to support synchronous running of the jobs.
Example code from PhotoUpload looks like this:
```php
$jobs = Functional\map($params->getArray('photo'), function (CM_File_UserContent_Temp $fileTmp) use ($user, $viewer) {
            return new SK_Entity_PhotoCreateJob(new CM_Params([
                'actor'     => $viewer,
                'user'      => $user,
                'fileTmpId' => $fileTmp->getUniqid(),
            ]));
        });
        $response->getServiceManager()->get(CM_Jobdistribution_RPCInterface::class)->runMultiple($jobs);
```

@njam @alexispeter wdyt?

I believe splitting those interfaces help keep things simpler.

njam:
sgtm

The method name of `CM_Jobdistribution_QueueInterface` don't look super clear to me, especially next to `CM_Jobdistribution_RPCInterface`'s. How about something like?
- runJob()
- runJobs()
- queueJob()
- consumeQueuedJobs()

tomaszdurka:
I agree that interface might be a bit too indirect. I am not sure about the `consume` at all being part of the interfaces. I think public interface should be publish/queueJob at most. Consume should be somehow run from dynamically created cli..... that may go in strange direction. Somehow like I thought about registering clockwork jobs. So the same way we could register cli processes.

njam:
agree

library/CM/Gearman/Factory.php
```diff
@@ -0,0 +1,79 @@
+<?php
+
+class CM_Gearman_Factory {
+
+    /**
+     * @param array $servers
+     * @param int   $workerJobLimit
+     * @return CM_Gearman_JobService
+     */
+    protected function createJobService(array $servers, $workerJobLimit) {
```
tomaszdurka: |`line 23`|
|---|
should be public

tomaszdurka:
|`library/CM/Gearman/Factory.php:33`|
|---|
should be protected

library/CM/Gearman/Factory.php
```diff
@@ -0,0 +1,79 @@
...
+    }
+
+    /**
+     * @param array                            $servers
+     * @param CM_Jobdistribution_JobSerializer $serializer
+     * @return CM_Gearman_Client
+     */
+    protected function createClient(array $servers, CM_Jobdistribution_JobSerializer $serializer) {
+        $client = $this->createGearmanClient($servers);
+        return new CM_Gearman_Client($client, $serializer);
```
tomaszdurka: |`line 60`|
|---|
should be protected

library/CM/Jobdistribution/QueueInterface.php
```diff
@@ -0,0 +1,8 @@
+<?php
+
+interface CM_Jobdistribution_QueueInterface {
+
+    public function publish(CM_Jobdistribution_Job_Abstract $job);
+
```
tomaszdurka: rename to `queue`?

fvovan:
transition is tested. seems to work

tomaszdurka:
|`resources/config/default.php:310`|
|---|
This somehow pops here, but it's not part of this pull-request, right? :|

fvovan:
hmm.. it's not. basically https://github.com/cargomedia/cm/pull/2513/files#diff-f275ed64b29d4f71f76c5dd27def3927 doesn't show it..
is it upsource bug?

PR #2520, opened 2017-01-24
Avoid migration script execution after an app setup

zazabe:
Part of https://github.com/cargomedia/cm/issues/2403

same than with db updates currently, see https://github.com/cargomedia/cm/blob/master/library/CM/Db/SetupScript.php#L47


zazabe:
@njam please review

njam:
lgtm

PR #2522, opened 2017-01-24
FB-112 Switch from CM_Site_Abstract::factory()

### LINKED ISSUE #2504: Remove default site and CM_Site_Abstract::factory method
Let's try to remove all the `CM_Site_Abstract::factory()` usages and method itself from the codebase. This will also get rid of default site concept.

**Usages without argument**
This way of invoking the method returns the default site. This is used extensively in tests. In those cases let's replace it with call to newly created `CMTest_TH::getDefaultSite`, a method which returns singleton instance of mocked site.
Unfortunately we also use default site in the real code:
- `CM_Http_Request_Abstract::popPathSite`
- `CM_Http_Request_Abstract::popPathSiteByMatch`
The goal is to remove this functionality, but first let's check if there are requests without site (let's log warning if so).

**Usages with argument**
This way of invoking the method searches for site by certain type. 
There are only several usages in the codebase. Let's replace it with `CM_Site_SiteFactory::findById`. This new method would search through the instance's siteList by type and return it (let's keep the name `findById` already).

fvovan:
closes #2504


fvovan:
here is the draft. `CM_Site_Abstract::factory()` is _almostly_ deleted.
so I added 
I have few questions/issues here.
 - it's still presents at `CMTest_TH::createResponsePage()`. can we move this method to `CMTest_TestCase` to get the access for its instance methods?
- isn't `(new CM_Site_SiteFactory())->getSiteById($userInput)` signature a bit cumbersome? I thought about making `CM_Site_SiteFactory` kind of singleton, holding its instance, but in the end decided it just would add unnecessary complexity.
- should i prepare PR for "concrete application" part?
@tomaszdurka , @njam wdyt?

tests/helpers/CMTest/library/CMTest/TestCase.php
```diff
@@ -322,6 +325,48 @@ public function forceInvokeMethod($objectOrClass, $methodName, array $arguments
...
+     * @param CM_Model_User|null        $viewer
+     * @param CM_Model_Language|null    $language
+     * @param DateTimeZone|null         $timeZone
+     * @param boolean|null              $debug
+     * @param CM_Model_Location|null    $location
+     * @param CM_Model_Currency|null    $currency
+     * @param CM_Http_ClientDevice|null $clientDevice
+     * @return CM_Frontend_Environment
+     */
+    public function getDefaultEnvironment(CM_Site_Abstract $site = null,
```
tomaszdurka: As this is test, what about making it a mock?

tests/helpers/CMTest/library/CMTest/TestCase.php
```diff
@@ -322,6 +325,48 @@ public function forceInvokeMethod($objectOrClass, $methodName, array $arguments
...
+     * @return CM_Frontend_Render
+     */
+    public function getDefaultRender() {
+        return new CM_Frontend_Render($this->getDefaultEnvironment());
+    }
+
+    /**
+     * @return CM_Site_Abstract|PHPUnit_Framework_MockObject_MockObject
+     */
+    public function getDefaultSite() {
```
tomaszdurka: Here it probably works as default as it's the same site every time.

tests/helpers/CMTest/library/CMTest/TestCase.php
```diff
@@ -322,6 +325,48 @@ public function forceInvokeMethod($objectOrClass, $methodName, array $arguments
...
+                                          CM_Model_Language $language = null,
+                                          DateTimeZone $timeZone = null,
+                                          $debug = null,
+                                          CM_Model_Location $location = null,
+                                          CM_Model_Currency $currency = null,
+                                          CM_Http_ClientDevice $clientDevice = null) {
+        if (null === $site) {
+            $site = $this->getDefaultSite();
+        }
+        return new CM_Frontend_Environment($site, $viewer, $language, $timeZone, $debug, $location, $currency, $clientDevice);
```
tomaszdurka: Mock as well.

My idea for creating mocks would be to create mocks with some factory which resolves dependencies via reflection and create mocks for those too.

tests/helpers/CMTest/library/CMTest/TestCase.php
```diff
@@ -382,7 +387,7 @@ protected function _createPage($pageClass, array $params = null) {
      * @return CM_Dom_NodeList
      */
     protected function _renderComponent(CM_Component_Abstract $component, CM_Model_User $viewer = null, CM_Site_Abstract $site = null) {
-        $render = new CM_Frontend_Render(new CM_Frontend_Environment($site, $viewer));
+        $render = new CM_Frontend_Render($this->getDefaultEnvironment($site, $viewer));
```
tomaszdurka: Ping me to discuss factory I've described before. Would be great to push it forward.

library/CM/Site/SiteFactory.php
```diff
@@ -52,4 +52,29 @@ public function getSite(CM_Http_Request_Abstract $request) {
...
+     * @throws CM_Exception_Invalid
+     */
+    public function getDefaultSite() {
+        $serviceManager = CM_Service_Manager::getInstance();
+        $options = $serviceManager->getOptions();
+        $defaultSiteId = $options->get('cm.defaultSiteId');
+        if (null === $defaultSiteId) {
+            throw new CM_Exception_Invalid('Default site is not set');
+        }
+        return $this->findSiteById($defaultSiteId);
```
tomaszdurka: I guess this should also throw if null returned.

library/CM/FormField/Site.php
```diff
@@ -19,6 +19,6 @@ protected function _initialize() {
      */
     public function validate(CM_Frontend_Environment $environment, $userInput) {
         $userInput = parent::validate($environment, $userInput);
-        return CM_Site_Abstract::factory($userInput);
+        return (new CM_Site_SiteFactory())->getSiteById($userInput);
```
tomaszdurka: It's pity we cannot pass this factory as dependency. Passing it to FF doesn't sound right.

tomaszdurka: I guess there is no other-easy choice without huge refactoring.

tests/helpers/CMTest/library/CMTest/TestCase.php
```diff
@@ -397,7 +390,7 @@ protected function _renderComponent(CM_Component_Abstract $component, CM_Model_U
      */
     protected function _renderFormField(CM_FormField_Abstract $formField, array $renderParams = null, CM_Frontend_Render $render = null) {
         if (null === $render) {
-            $render = $this->getDefaultRender();
+            $render = new CM_Frontend_Render();
```
tomaszdurka: Ok so we remove it now :)

tests/helpers/CMTest/library/CMTest/TestCase.php
```diff
@@ -376,7 +376,7 @@ protected function _createPage($pageClass, array $params = null) {
      * @return CM_Dom_NodeList
      */
     protected function _renderComponent(CM_Component_Abstract $component, CM_Model_User $viewer = null, CM_Site_Abstract $site = null) {
-        $render = new CM_Frontend_Render($this->getDefaultEnvironment($site, $viewer));
+        $render = new CM_Frontend_Render(new CM_Frontend_Environment($site, $viewer));
```
tomaszdurka: So we remove this too! :)

library/CM/Site/SiteFactory.php
fvovan: it was changed  with new added `getSiteById()`

tests/helpers/CMTest/library/CMTest/TestCase.php
fvovan: yep, it was in favour of factory calls

tests/helpers/CMTest/library/CMTest/TestCase.php
fvovan: it was removed later

tests/helpers/CMTest/library/CMTest/TestCase.php
fvovan: it was removed

library/CM/Site/Abstract.php
```diff
@@ -267,6 +267,7 @@ public function equals(CM_Comparable $other = null) {
      * @param int|null $type
      * @return CM_Site_Abstract
      * @throws CM_Class_Exception_TypeNotConfiguredException
+     * @deprecated use CM_Site_SiteFactory methods instead
      */
     public static function factory($type = null) {
```
tomaszdurka: Shall we remove this method?

fvovan: done

tomaszdurka:
So this is ready? Lgtm, right?

PR #2527, opened 2017-01-25
Allow to pass metaInfo to client exception

dlondero:
This is in order to be able to pass params to the client using metaInfo when needed. Currently results in empty object.


njam:
Do the same for `CM_Http_Response_RPC`?

PR #2530, opened 2017-01-26
Follow-up: #2527. Allow to pass metaInfo to client exception.

dlondero:
Using `getClientData()` in `CM_Http_Response_RPC` as well.


njam:
lgtm

njam:
cc @tomaszdurka

Issue #2525, opened 2017-01-25
Don't fork() in Clockwork


alexispeter:
Currently Clockwork events are running in a forked process.
We would like to drop forking (https://github.com/cargomedia/cm/issues/2524).

Instead we could run Clockwork events in job-distribution (Gearman).

Usages:
- **Elasticsearch CLI**: Move it to Clockwork CLI
- **Maintenance CLI Local**: Remove it!
  - SVM: Remove it completely
  - Command Manager process synchronization: No need to call it in Clockwork. We already call it before *checking* a synchronized process, which is sufficient
- **Maintenance CLI**: Run each event/task in a new Gearman job, which would need to communicate with the manager process to let it know about when it's finished:
```php
class CM_Clockwork_RunEventJob extends CM_Jobdistribution_Job_Abstract {

    protected function _execute(CM_Params $params) {
        $maintenanceService = $serviceManager->getMaintenanceService('maintenance');
        try {
            $maintenanceService->runClockworkEvent('event-identifier');
        } finally {
            $maintenanceService->setClockworkEventResult();
        }
    }

}
```
Move `CM_Clockwork_Storage_FileSystem` to MongoDB.

Part 1: https://github.com/cargomedia/cm/pull/2532
Part 1: https://github.com/cargomedia/cm/pull/2567

PR #2532, opened 2017-01-26
Consolidate clockwork usage and remove SVM module

alexispeter:
part of https://github.com/cargomedia/cm/issues/2525

- **Elasticsearch CLI**: Move it to Clockwork CLI
- **Maintenance CLI Local**: Remove it!
  - SVM: Remove it completely
  - Command Manager process synchronization: No need to call it in Clockwork. We already call it before *checking* a synchronized process, which is sufficient


alexispeter:
About the storage-files for maintenance. Since they are on the local filesystem, they have to be deleted manually.
- `data/clockwork/083c9a298e60b51c071188cd483736b7.json`
- `data/clockwork/f4fd7caa5b437ad5161bca1284a0720d.json`

The same applies to training-files for svm if they still exist. They would be in `data/svm`

njam:
@dlondero could you review this?

dlondero:
@alexispeter lgtm. I guess you can solve conflict and merge.

PR #2535, opened 2017-01-27
Breaking: CM Usages of CM\Url\*Url

zazabe:
- `CM_Site_Abstract` returns `CM\Url\BaseUrl` for all `getUrl*()` methods
- `CM_Frontend_Render::getUrl*()` use `CM\Url\*Url` objects to build URLs
- remove `CM_Util:link()`
- remove `CM_Http_UrlParser()`
- use `CM_Page_Abstract:: getUrlComponents()` to customize page URL VS`CM_Page_Abstract::getUrl()`
- test fixes


library/CM/Frontend/Render.php
```diff
@@ -240,49 +223,26 @@ public function getUrlPage($pageClassName, array $params = null, CM_Site_Abstrac
...
      */
     public function getUrlResource($type = null, $path = null, array $options = null, CM_Site_Abstract $site = null) {
-        $options = array_merge([
-            'sameOrigin' => false,
-        ], (array) $options);
-        if (null === $site) {
-            $site = $this->getSite();
+        // TODO: strange smell here, verify usages...
+        if (null === $type || null === $path) {
+            return (string) $this->getEnvironment()->getSite()->getUrlCdn();
```
njam: Instead of accessing the CDN URL directly here, how about creating a ResourceUrl with an empty path? This way we avoid repeating the logic to pick the base URL here.

library/CM/Site/Abstract.php
```diff
@@ -111,28 +113,25 @@ public function getThemes() {
...
+     * @return BaseUrl
      */
     public function getUrl() {
-        return (string) self::_getConfig()->url;
+        return BaseUrl::create(self::_getConfig()->url);
     }
 
     /**
      * @return CM_Http_UrlParser
+     * @deprecated
```
njam: Should we replace it right away?

library/CM/Site/Abstract.php
```diff
@@ -148,25 +147,27 @@ public function getWebFontLoaderConfig() {
...
 
     /**
-     * @return string
+     * @return BaseUrl
      * @throws CM_Exception
      */
     public function getUrlBase() {
-        $urlParser = $this->getUrlParser();
-        return $urlParser->getScheme() . '://' . $urlParser->getHost();
+        // TODO: URL - check usages...
```
njam: This looks good to me

zazabe:
> It looks to me like the "sameOrigin" option is not used anymore now

right, used in tests only I think, I'll adjust  #2491  accordingly 

 
> Did you test this in the browse already?

I was waiting for https://github.com/cargomedia/puppet-cargomedia/pull/1778

>  Maybe we could compare the HTML of the version created from master, with what's generated with this code?

Yes, i'll do more acceptance tests now that https://github.com/cargomedia/puppet-cargomedia/pull/1778 is merged

library/CM/Site/Abstract.php
zazabe: Used here `$this->setHeader('Access-Control-Allow-Origin', $this->getSite()->getUrlBase());`, but could be replaced by `$this->getSite()->getUrl()->withoutPrefix()`.

Also used in `CM_Http_Response_View_Abstract::loadPage()`, but i'm not sure to understand why `$url` has to be generated without the prefix...

njam: Because the client will send the *full* path (including the fragment) on route changes.
The reason is that the client cannot know if a path segment is a site-path-prefix or a page-path, because it doesn't know what pages exist…

library/CM/Frontend/Render.php
zazabe: Not sure, the `$type` is required to get a resource. 
Also, there's only test usages of `getUrlResource()` without params, and it seems to be just a way to retrieve the base resource url, not really a Url to a resource...

Better off making `$path` and `$type` required!

library/CM/Frontend/Render.php
```diff
@@ -199,90 +204,44 @@ public function getUrl($path = null, CM_Site_Abstract $site = null) {
...
      * @param CM_Site_Abstract|null $site
      * @return string
      */
-    public function getUrlResource($type = null, $path = null, array $options = null, CM_Site_Abstract $site = null) {
-        $options = array_merge([
-            'sameOrigin' => false,
-        ], (array) $options);
-        if (null === $site) {
-            $site = $this->getSite();
+    public function getUrlResource($type, $path, array $options = null, CM_Site_Abstract $site = null) {
```
njam: I think we can remove `$options` completely now, all the way down

library/CM/Http/Response/Page.php
```diff
@@ -69,9 +72,8 @@ protected function _process() {
 
     protected function _processContentOrRedirect() {
         $render = $this->getRender();
```
njam: $render not needed anymore

library/CM/Http/Response/Resource/Abstract.php
```diff
@@ -3,7 +3,7 @@
 abstract class CM_Http_Response_Resource_Abstract extends CM_Http_Response_Abstract {
 
     protected function _setContent($content) {
-        $this->setHeader('Access-Control-Allow-Origin', $this->getSite()->getUrlBase());
+        $this->setHeader('Access-Control-Allow-Origin', $this->getSite()->getUrlBase()->getUriBaseComponents());
```
njam: Why not just string-cast the base URL?

library/CM/Page/Abstract.php
```diff
@@ -63,27 +63,15 @@ public function prepare(CM_Frontend_Environment $environment, CM_Frontend_ViewRe
      * @return string
      */
     public static function getPath(array $params = null) {
```
njam: Should we maybe mark this deprecated? Or you wanted to keep it as a helper function?

library/CM/Site/Abstract.php
```diff
@@ -144,25 +139,26 @@ public function getWebFontLoaderConfig() {
...
-        return $this->getUrlParser()->getPath();
+        return $this->getUrl()->getUriRelativeComponents();
     }
 
     /**
-     * @return string
+     * @return BaseUrl
      * @throws CM_Exception
      */
     public function getUrlBase() {
```
njam: I'm wondering if we should remove this function altogether and just call `->getUrl()->withoutPrefix()`. wdyt?

library/CM/Page/Abstract.php
zazabe: It is used in many places, and using the `CM_Page_UrlFactory` seems less intuitive... I would keep it for now

library/CM/Http/Response/Resource/Abstract.php
zazabe: because a trailing `/` is added with string-casting

zazabe:
@njam please review

njam:
- CM/App.js: The "options" argument can be removed I think

Rest lgtm!

zazabe:
@njam approved?

zazabe:
guess yes... let's merge it ;)

njam:
- CM_App.getUrlResource(): Remove the "options" argument completely?
- CM_App.getUrlStatic(): Remove "urlBase" case

zazabe:
@njam fixed in https://github.com/cargomedia/cm/pull/2560

Issue #2536, opened 2017-01-29
Introduce primary key on cm_action


njam:
We need a primary key on the table, so that we can use "pt-online-schema-change" (see https://github.com/cargomedia/puppet-cargomedia/issues/1781).

I think we need to introduce a new column, like "id".
Can we use "pt-online-schema-change" to introduce a primary column?

@alexispeter wdyt? Could you do the change?
cc @fvovan @kris-lab

fvovan:
done

PR #2537, opened 2017-01-30
add primary key to `cm_action` table

fvovan:
pt-online-schema-change does not work with tables without primary or unique keys


fvovan:
closes #2536

fvovan:
I run a benchmark for `SK_User_SignUpProcess::process()`
neither optimizing table nor converting it to InnoDb gave any significant boost.
But anyway it would be good to define `id` at least to be able run percona-tools

njam:
lgtm

as discussed let's remove the "return" because we need the application to be down.

PR #2542, opened 2017-02-01
Update SocketRedis client to 3.0.0

vogdb:


vogdb:
@tomaszdurka this one is important and not so difficult, please review.

vogdb:
well, I would ask @zazabe. Can I use cm-bundler for this one?

PR #2543, opened 2017-02-01
handle fluentD errors

fvovan:
It appeared that the module does not throw at `post()` but rather returns `true/false` codes.


njam:
It would be nice to have an error *message*, otherwise it's difficult to debug.
I can see their "FluentLogger" has a function `registerErrorHandler()`, could we use it?

fvovan:
>I can see their "FluentLogger" has a function registerErrorHandler(), could we use it?

Agree, sometimes it can be useful.  Unfortunately it's quite hard to cover this case with unit tests not touching the module itself, but i tested it manually.

fvovan:
that's really weird but using `registerErrorHandler()` leads to segfault during tests.
I suppose the reason could be they pass `Fluent\Logger\FluentLogger` as a first argument and it somehow leads to circular links and errors at GC.
Possible solution is unregistering handlers write after `post()` method call. 
 but currently they don't have any kind of unregistering. So I just hacked dependency locally and it worked.
Shall I make a PR? it's only 5 lines of code.
@njam wdyt?

fvovan:
https://github.com/fluent/fluent-logger-php/pull/47

fvovan:
PR is merged and tagged, but it hasn't appeared at packagist yet as it seems they don't have github webhook configured.
so just awaiting a week.
>Existing packages without auto-updating (GitHub/BitBucket hook) will be crawled **once a week** for updates.

fvovan:
at last  fluendD lib has been updated.
so the only way to avoid segfault is to reset error handler right after `post()` execution.
hope that will not affect the performance much.
@njam please review

njam:

alexispeter:
lgtm, I take it the segfault wasn't reproducable locally?

fvovan:
it was not consistently reproducible on local, 
sometimes it failed just within test case file. Sometimes only when run together with the rest of tests.

PR #2545, opened 2017-02-02
Introduce CM_Type_Enum client side version

zazabe:
Useful to expose `CM_Type_Enum` on the client slide


zazabe:
@vogdb please review

PR #2546, opened 2017-02-02
Allow to put JS file in service library

zazabe:
Useful for example if a service define some models that should be exposed on the client side.

@njam wdyt?


zazabe:
@njam please review

njam:
I'm wondering if this should not be default for `rglobLibraries()` to include "XService" files? There's no real difference between "X" and "XService" from this point of view, it's just to structure the files.

zazabe:
so `CM_Util::getModulePath($moduleName) . 'library/*/';` maybe?

zazabe:
@njam please check again

njam:
lgtm

PR #2550, opened 2017-02-02
Trim git output for migration name

njam:
```
$ bin/cm migration add
Migration script name [master
]:
```

@zazabe please review


njam:
I have it on OSX

PR #2557, opened 2017-02-06
Check params for 'captcha' request

### LINKED ISSUE #2549: Captcha Response: Undefined index: id
> ErrorException: E_NOTICE: Undefined index: id in CM/Http/Response/Captcha.php on line 7

Should be a warning because it's due to bad user input.

fvovan:
And throw a warning if it's wrong

closes #2549


fvovan:
there is another option - to add config entry  with `$config->CM_Http_Response_Captcha->exceptionsToCatch` and use `_runWithCaching()` but it could be an overkill and  I'm not sure what to do in its error callback.
@tomaszdurka wdyt?

tomaszdurka:
I think this is fine. lgtm

PR #2559, opened 2017-02-06
BREAKING: Pass custom variables to Revive iframe ads

fauvel:


fauvel:
@njam please review!

I might have an error [here](https://github.com/cargomedia/cm/pull/2559/files#diff-5ae7dd9d1f5666fe42d2f6d337f23d94R14), please advise: Is the endpoint `/delivery/ajs-proxy.php` for ads loaded via JavaScript only, while `/delivery/afr.php` is for iframe ads only?

library/CM/Util.php
```diff
@@ -204,6 +204,17 @@ public static function htmlspecialchars($string, $quote_style = ENT_COMPAT, $cha
     }
 
     /**
+     * @param array  $query_data
+     * @param string $numeric_prefix
+     * @param string $arg_separator
+     * @param int    $enc_type
+     * @return string
+     */
+    public static function http_build_query($query_data, $numeric_prefix = '', $arg_separator = '&', $enc_type = PHP_QUERY_RFC3986) {
```
njam: Instead use `CM\Url\Url::createWithParams()` which is introduced in https://github.com/cargomedia/cm/pull/2535.
@zazabe wdyt? could you merge the changes to the URL library to master?

tests/library/CM/AdproviderAdapter/Revive/IframeTest.php
```diff
@@ -0,0 +1,22 @@
...
+            'zoneId' => 1,
+            'host'   => 'example.com',
+            'width'  => '100%',
+            'height' => 100,
+        ], [
+            'foo' => 'bar',
+        ]);
+
+        $this->assertSame(
+            '<iframe src="//example.com/delivery/ajs-proxy.php?foo=bar&amp;zoneId=1&amp;cb=963932192" width="100%" height="100" class="advertisement-hasContent" data-variables="{}" frameborder="0" scrolling="no"></iframe>'
```
njam: I think the `data-variables` attribute can be removed completely. I don't think it's ever useful, because the document in the iframe cannot access the attribute.

njam:
> Is the endpoint `/delivery/ajs-proxy.php` for ads loaded via JavaScript only, while `/delivery/afr.php` is for iframe ads only?

Yes, the endpoints are different for AJAX and iframe.

library/CM/Util.php
fauvel: Using PSR7? Sgtm! @zazabe let me know if you want to pair on that!

njam: CM\Url will take a bit longer.
Let's go without this for now.
But can we use `CM_Util::link()` in `CM_AdproviderAdapter_Revive_Iframe`, to build the full URL?

library/CM/AdproviderAdapter/Iframe.php
```diff
@@ -2,20 +2,22 @@
...
-            'frameborder'    => 0,
-            'scrolling'      => 'no',
+            'src'         => $src,
+            'width'       => $width,
+            'height'      => $height,
+            'class'       => 'advertisement-hasContent',
+            'frameborder' => 0,
+            'scrolling'   => 'no',
         ];
+        if (null !== $variables) {
```
njam: Let's either always or never include it in the DOM. Otherwise parsing it from JS will be difficult.

fauvel:
Ok, please review a last time!

fauvel:
Tagged 1.235.0

PR #2558, opened 2017-02-07
Introduce simple mechanism for application transactions

tomaszdurka:
### Small introduction

As noticed we have a lot actions which can fail only partly done. In case of a failure we throw exception, but do not revert changes made. This results in invalid, corrupt data. 

Sounds like transactions are needed. We could use builtin transactions for mysql, but with our mixed databases it wouldn't work so great. 

### Current solutions
Usually in those cases we use try-catch blocks with undoing in the catch block. Example:
```php
$user = CM_Model_User::createStatic(array('site' => $site, 'language' => $language, 'currency' => $currency));
// 20 lines of code ...
try {
    CM_Db_Db::insert('sk_user', $values);
} catch (CM_Exception $e) {
    CM_Db_Db::delete('cm_user', array('userId' => $userId));
    throw $e;
}
```
This results in cleaning up `cm_user` in case of `sk_user` insert failure. 
It works, but is not so clean, and the code related with cleaning (rollback) is far away from messing part.

### Different solution
Small, simple solution is to keep above kind of pattern with addition for object to store all those "rollbacks"

Example:
```php
$transaction = new Transaction();
try {
    $user = CM_Model_User::createStatic(array('site' => $site, 'language' => $language, 'currency' => $currency));
    $transaction->addRollback(function() use ($user) {
        CM_Db_Db::delete('cm_user', array('userId' => $user->getId()));
        // or rather
        $user->delete();
    });
    // 20 lines of code ...
    CM_Db_Db::insert('sk_user', $values);
} catch (CM_Exception $e) {
    $transaction->rollback();
    throw $e;
}
```

@alexispeter @zazabe wdyt?


tomaszdurka:
@zazabe please review

zazabe:
https://en.wikipedia.org/wiki/Two-phase_commit_protocol but it seems not matching with the title term: "simple mechanism" ;)

tomaszdurka:
- Removed Rollback class.
- Regarding usages we will not pass transaction along wth atomic rollbacks, but rather use it per level.

tomaszdurka:
I guess you agree!?

zazabe:
oui

PR #2567, opened 2017-02-09
FB-98 Run Maintenance jobs in gearman

alexispeter:
Part of https://github.com/cargomedia/cm/issues/2525

- **Maintenance CLI**: Run each event/task in a new Gearman job, which would need to communicate with the manager process to let it know about when it's finished:
```php
class CM_Clockwork_RunEventJob extends CM_Jobdistribution_Job_Abstract {

    protected function _execute(CM_Params $params) {
        $maintenanceService = $serviceManager->getMaintenanceService('maintenance');
        try {
            $maintenanceService->runClockworkEvent('event-identifier');
        } finally {
            $maintenanceService->setClockworkEventResult();
        }
    }

}
```
Move `CM_Clockwork_Storage_FileSystem` to MongoDB.


alexispeter:
please review @tomaszdurka 

Notable changes:
- Decoupled Maintenance and Clockwork, Clockwork now uses an eventHandler to trigger registered events.
- Clockwork persistence for maintenance moved to mongodb due to distributed event-execution (gearman).
- Event definition moved from CM_Maintenance_Cli to CM_Maintenance_ServiceFactory.

Pitfalls:
- The "isRunning"-property of an event's status relies on the event to finish properly. So if a jobWorker crashes while running an event, it's status will remain the same which could prevent it from ever running again. To guard against that to some extend at least, it is always set to false when loading an event for the first time. So restarting the maintenance-daemon should clear such blockades.

Further thoughts:
- Loading/Caching: The current implementation of clockwork-storage causes the scheduling process to do 1 read per registered event per second with only 2-3 writes per event per execution, which could be considered suboptimal. However, I don't think it's necessary to try and improve this.

njam:
@fvovan would you be up to review this?
Also thorough testing would be good.

alexispeter:
Fetching event-status should only happen once per cycle for all events in bulk. Saving it still needs to be done one by one.

fvovan:
started on Friday

library/CM/Clockwork/Manager.php
```diff
@@ -14,25 +14,21 @@ class CM_Clockwork_Manager {
...
 
-    public function __construct() {
+    /** @var CM_Clockwork_Event_Status[] */
+    private $_statusList = [];
+
+    /**
+     * @param CM_EventHandler_EventHandlerInterface $eventHandler
+     */
+    public function __construct(CM_EventHandler_EventHandlerInterface $eventHandler) {
         $this->_events = array();
```
fvovan: I'm just wondering why we store events here and `CM_Maintenance_Service::_eventsList` in plain non-indexed array? It requires us to make, well not really tricky but a bit bloated checkings, while it could be just `array_key_exists()`

library/CM/Maintenance/Service.php
```diff
@@ -0,0 +1,125 @@
...
+        // TODO: remove debug-code
+        $timeEnd = (new DateTime())->format('H:i:s');
+        (new CM_File(DIR_ROOT . '/log.txt'))->appendLine($timeEnd . ' stopping: ' . $eventName . ' started: ' . $timeStart);
+    }
+
+    /**
+     * @return CM_Clockwork_Manager
+     */
+    private function _getClockworkManager() {
+        if (!$this->_clockworkManager) {
```
alexispeter: `runEvents()` is called repeatedly in CM_Mtaintenance_Cli::start(). Reinitialising clockwork every time makes no sense.

library/CM/Clockwork/Storage/MongoDB.php
```diff
@@ -0,0 +1,95 @@
+<?php
+
+class CM_Clockwork_Storage_MongoDB extends CM_Clockwork_Storage_Abstract {
+
```
alexispeter: It's all the same to me.

library/CM/Maintenance/Service.php
fvovan: basically I didn't mean reinitialising each time but rather just moving initializing code (with same memoization) into `runEvents`.

library/CM/Clockwork/Manager.php
alexispeter: That's how it was done before and I just adopted it. But yes, we could use a map instead

library/CM/Clockwork/Manager.php
```diff
@@ -178,54 +153,87 @@ protected function _isIntervalEvent(CM_Clockwork_Event $event) {
...
+        return $duplicateEventName;
+    }
+
+    /**
+     * @param CM_Clockwork_Event $event
+     * @return CM_Clockwork_Event_Status
+     */
+    protected function _getStatus(CM_Clockwork_Event $event) {
+        $status = $this->_storage->getStatus($event);
+        if (!array_key_exists($event->getName(), $this->_statusList)) {
```
alexispeter: It's what I talked about in a previous comment, that, whereas previously `isRunning` was stored in the maintenance-process' memory and therefor reset when the daemon starts, now it is stored in mongo and therefor persistent. Since there is no way to check if a certain job runs with certain parameters, there is no way to check if a maintenance-task is really running or not, which can lead to "zombie" tasks, which are not truly running, but thought to be running by clockwork.
So every task is marked as not running when the daemon starts.

library/CM/Maintenance/Service.php
alexispeter: it's used in more than one place though.

library/CM/Clockwork/Storage/Abstract.php
```diff
@@ -5,50 +5,25 @@
...
-    /**
-     * @return array
-     */
-    abstract protected function _load();
-
-    /**
-     * @param array $data
-     */
-    abstract protected function _save(array $data);
+    abstract public function fetchData();
```
alexispeter: This would require`runEvents()` to force-read only the first event in the loop and read the rest normally, since we only want to fetch data from db once, which would not make things any less complicated.

library/CM/Maintenance/Service.php
```diff
@@ -0,0 +1,125 @@
...
+     * @param string  $dateTimeString
+     * @param Closure $callback
+     * @throws CM_Exception_Invalid
+     */
+    public function registerEvent($name, $dateTimeString, Closure $callback) {
+        $name = (string) $name;
+        $dateTimeString = (string) $dateTimeString;
+
+        $duplicateEventName = \Functional\some($this->_eventList, function (CM_Maintenance_Event $event) use ($name) {
+            return $event->getName() == $name;
```
alexispeter: Shouldn't be necessary, since both are strings.

library/CM/Clockwork/Storage/MongoDB.php
alexispeter: It wouldn't make any sense storing it in mongodb then.

fvovan: it's up to you. As for me storing in mongo db still has advantages over mysql because of flexible event attributes schema (status, success) etc ie its extensible. 
And my point for such not so obvious things (i mean not only this remark, but some others as well) is just usually it takes more time to understand the code. But once somebody did that, it does not realy matter.

alexispeter: I just think that since we always read the data of all events in one go, we should put it all in one document and take advantage of mongodb's subdocument capabilities.

fvovan: Yep i see it's more elegant from the storage's perspective.
btw I forgot to mention another concern regarding it. Is running of several `setStatus()` theoretically possible? If yes - there is a probabilty that https://github.com/cargomedia/cm/pull/2567/files/f53620cc1e18199b3452fee3e09e97f36171d63f#diff-c09eb2f3776d30476b9110a015703cc8R48 will be triggered twice.

alexispeter: You mean that race-conditions would cause two status to be pushed to the array? Even if, for some reason, clockwork were to trigger the same event twice, the probability of them writing their status would be very very low. Furthermore, this clause should only be entered the very first time an event is run, so the probability should be even lower.
Should it still happen, then an event would have multiple status-entries, which would all be updated as one, but clockwork would only ever read the first one. I think we can live with that.

We could make the second update fail by adding `{'events.name': {'$ne': <eventName>}}` to the criteria, but this would make the error-message very cryptic, because the unmatched update-criteria would cause the upsert-part of the query to attempt to create a second document with the same context, leading to a duplicate-key exception on `context`.

fvovan: agree, let's keep it

library/CM/Clockwork/Manager.php
```diff
@@ -178,54 +153,87 @@ protected function _isIntervalEvent(CM_Clockwork_Event $event) {
...
+            throw new CM_Exception_Invalid('Cannot stop event. Event is not running.', null, ['eventName' => $event->getName()]);
         }
-        $this->_eventsRunning[$event->getName()] = ['event' => $event, 'identifier' => $identifier, 'startTime' => $startTime];
+        $status->setRunning(false);
+        $this->_storage->setStatus($event, $status);
     }
 
     /**
      * @param CM_Clockwork_Event $event
      * @throws CM_Exception_Invalid
```
alexispeter: I think this is a mixup because of method-rearranging. The docu you're referring to belonged to `_markStopped()` which is now `setStopped()` and still throws, while `setCompleted()`, like `_markCompleted()` in the past, doe not throw. 
Do you agree?

fvovan: agree, but shouldn't it be deleted from `setCompleted()` docu?

alexispeter: Actually, I think I left it there because it can still come from `_checkEventExists()` which is called in there. 
We normally don't propagate thrown exceptions docu to calling methods, but I think it's a good practice.

fvovan: ahh good. actually I like the idea of exeptions propagation.

alexispeter:
I double-checked to be sure, just had to manually spawn some additional workers since there is only one active worker on dev. 
@njam any chance we can change that to reflect production settings?

fvovan:
>I double-checked to be sure, just had to manually spawn some additional workers since there is only one active worker on the dev-environment.

ahh it's just running one more `./bin/cm --quiet job-distribution start-worker`.

njam:
Sounds reasonable.
Could you change it here?:
https://github.com/cargomedia/sk/blob/master/Vagrantfile#L84
(`foreman-systemd` accepts a `-f` argument, which is passed to *foremans*'s `--formation` argument: https://ddollar.github.io/foreman/)

Issue #2566, opened 2017-02-09
Migrations in cm are run twice


alexispeter:
When running the migration cli command or executing tests in cm, all migration-scripts are handled twice.

The problem lies within CM_Migration_Manager::_getMigrationPaths().
```php
protected function _getMigrationPaths() {
    $paths = [
        self::getMigrationPathByModule(),
    ];
    foreach ($this->_modules as $moduleName) {
        $paths[] = self::getMigrationPathByModule($moduleName);
    }
    return $paths;
}
```
Both `self::getMigrationPathByModule()` and `self::getMigrationPathByModule($moduleName)` return `/home/vagrant/cm//resources/migration` which causes the directory to be scanned twice.

The exception being thrown
```
CM_Db_Exception: Cannot execute SQL statement in /home/vagrant/cm/library/CM/Db/Statement.php on line 47
     0. {main} /home/vagrant/cm/vendor/phpunit/phpunit/phpunit:0
     1. PHPUnit_TextUI_Command->main() /home/vagrant/cm/vendor/phpunit/phpunit/phpunit:47
     2. PHPUnit_TextUI_Command->run([], true) /home/vagrant/cm/vendor/phpunit/phpunit/src/TextUI/Command.php:115
     3. PHPUnit_TextUI_Command->handleArguments([]) /home/vagrant/cm/vendor/phpunit/phpunit/src/TextUI/Command.php:126
     4. PHPUnit_TextUI_Command->handleBootstrap('/home/vagrant/cm/tests/bootstr...') /home/vagrant/cm/vendor/phpunit/phpunit/src/TextUI/Command.php:731
     5. PHPUnit_Util_Fileloader->checkAndLoad('/home/vagrant/cm/tests/bootstr...') /home/vagrant/cm/vendor/phpunit/phpunit/src/TextUI/Command.php:902
     6. PHPUnit_Util_Fileloader->load('/home/vagrant/cm/tests/bootstr...') /home/vagrant/cm/vendor/phpunit/phpunit/src/Util/Fileloader.php:38
     7. include_once('/home/vagrant/cm/tests/bootstr...') /home/vagrant/cm/vendor/phpunit/phpunit/src/Util/Fileloader.php:56
     8. CMTest_TestSuite->bootstrap() /home/vagrant/cm/tests/bootstrap.php:14
     9. CMTest_TH->init() /home/vagrant/cm/tests/helpers/CMTest/library/CMTest/TestSuite.php:14
[...]

PR #2568, opened 2017-02-09
Ensure no duplicated migration paths

zazabe:
fix for https://github.com/cargomedia/cm/issues/2566


zazabe:
@alexispeter please review

PR #2570, opened 2017-02-09
Use `secret` param at SocketRedis fetch status operation

vogdb:


vogdb:
@tomaszdurka please review

PR #2572, opened 2017-02-13
Use different client/server side component reload methods

zazabe:
- with the current `reloadComponent`, there's no simple method to get the reloaded component after a `reload()` on the client side
- Allows to redefine some properties after a component reload, ie:
```js
var CM_Component_Foo = CM_Component_Abstract.extend({
  _bar: null,
  // ...
  reload: function() {
    return this
      .try(function() {
        return CM_Component_Abstract.prototype.reload.apply(this, arguments);
      })
      .then(function(newComponent) {
        newComponent._bar = this._bar;
        return newComponent;
      });
  }
});
```

### DIFF
--- a/library/CM/Component/Abstract.php
+++ b/library/CM/Component/Abstract.php
@@ -15,4 +15,8 @@ abstract class CM_Component_Abstract extends CM_View_Abstract implements CM_View
     protected function _checkViewer(CM_Frontend_Environment $environment) {
         $environment->getViewer(true);
     }
+
+    public function ajax_reloadComponent(CM_Params $params, CM_Frontend_JavascriptContainer_View $handler, CM_Http_Response_View_Ajax $response) {
+        return $response->loadComponent(static::class, $this->getParams()->merge($params));
+    }
 }
--- a/library/CM/Component/Abstract.php
+++ b/library/CM/Component/Abstract.php
@@ -17,8 +17,11 @@ abstract class CM_Component_Abstract extends CM_View_Abstract implements CM_View
     }
 
     public function ajax_reloadComponent(CM_Params $params, CM_Frontend_JavascriptContainer_View $handler, CM_Http_Response_View_Ajax $response) {
-        foreach ($this->getParams()->getParamsDecoded() as $key => $value) {
-            $params->set($key, $value);
+        $componentParams = $this->getParams();
+        foreach ($componentParams->getParamNames() as $key) {
+            if (!$params->has($key)) {
+                $params->set($key, $componentParams->get($key));
+            }
         }
         return $response->loadComponent(static::class, $params);
     }
--- a/library/CM/Component/Abstract.php
+++ b/library/CM/Component/Abstract.php
@@ -17,6 +17,9 @@ abstract class CM_Component_Abstract extends CM_View_Abstract implements CM_View
     }
 
     public function ajax_reloadComponent(CM_Params $params, CM_Frontend_JavascriptContainer_View $handler, CM_Http_Response_View_Ajax $response) {
-        $response->reloadComponent($params->getParamsDecoded());
+        foreach ($this->getParams()->getParamsDecoded() as $key => $value) {
+            $params->set($key, $value);
+        }
+        return $response->loadComponent(static::class, $params);
     }
 }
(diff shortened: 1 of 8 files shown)


zazabe:
wdyt about that @njam?

library/CM/Component/Abstract.php
```diff
@@ -17,6 +17,12 @@ protected function _checkViewer(CM_Frontend_Environment $environment) {
     }
 
     public function ajax_reloadComponent(CM_Params $params, CM_Frontend_JavascriptContainer_View $handler, CM_Http_Response_View_Ajax $response) {
-        $response->reloadComponent($params->getParamsDecoded());
+        $componentParams = $this->getParams();
+        foreach ($componentParams->getParamNames() as $key) {
```
njam: If you do it this way, only parameters already present on $componentParams can be overwritten.
I think it shoud be the other way around - take $componenParams as the base, and overwrite with everything from $params - like it s done already in `reloadComponent()`.

library/CM/Component/Abstract.php
```diff
@@ -17,6 +17,12 @@ protected function _checkViewer(CM_Frontend_Environment $environment) {
...
+            if (!$params->has($key)) {
+                $params->set($key, $componentParams->get($key));
+            }
+        }
+        return $response->loadComponent(static::class, $params);
```
njam: Nice, we don't use `reloadComponent` anymore!

library/CM/View/Abstract.js
```diff
@@ -290,9 +297,29 @@ var CM_View_Abstract = Backbone.View.extend({
   },
 
   /**
+   * @param {Object} [params]
+   * @param {Object} [options]
+   * @return Promise
+   */
+  reloadComponent: function(params, options) {
```
njam: Shouldn't this be on CM_Component_Abstract - the ajax function is also only availabe there?

tests/library/CM/Http/Response/View/AbstractTest.php
```diff
@@ -282,21 +282,18 @@ public function testReloadComponent() {
...
-cm.window.appendHidden("<div id=\\"$newAutoId\\" class=\\"CM_Component_Notfound CM_Component_Abstract CM_View_Abstract\\">Sorry, this page was not found. It has been removed or never existed.\\n<\/div>");
-cm.views["$newAutoId"] = new CM_Component_Notfound({el:$("#$newAutoId").get(0),params:{"foo":"bar"}});
-cm.views["$oldAutoId"].getParent().registerChild(cm.views["$newAutoId"]);
-cm.views["$oldAutoId"].replaceWithHtml(cm.views["$newAutoId"].\$el);
-cm.views["$newAutoId"]._ready();
-EOL;
-        $this->assertSame($expected, $frontend->getJs());
+        $jsonResponse = CM_Util::jsonDecode($response->getContent());
+
+        // can't get the autoId from the globalResponse because of https://github.com/cargomedia/cm/blob/master/library/CM/Http/Response/View/Abstract.php#L135
```
njam: This line is likely to move, and the link will point somewhere else. Shall we just write, "because Response::_getComponentRendering() creates a new Render instance"?

library/CM/View/Abstract.js
zazabe: what about moving `ajax_reloadComponent` in `CM_View_Abstract` instead?

njam: Only components can be reloaded. Why do you want to have the function on the View?

tests/library/CM/Http/Response/View/AbstractTest.php
zazabe: thought it will help you to review :)

library/CM/Component/Abstract.php
zazabe: Introduced `CM_Params::merge()` to merge `decoded` params values

library/CM/View/Abstract.js
zazabe: because we have `ajax_loadComponent` in `CM_View_Abstract` already... thought we could reload a Form for example...

njam: "loadComponent()" loads a *new* component and makes it a child view of the current view. It can be done on any view. But "reload**Component()**" reloads the current component, so it has to be a component. There are some differences between the views that prevent us currently from reloading any view iirc.

zazabe: ok, get it, will move all back to `CM_Component_Abstract`

zazabe:
@njam please check again

PR #2573, opened 2017-02-15
Don't fully decode parameters when queueing delayed jobs

alexispeter:
fixes https://github.com/cargomedia/sk/issues/4547

Instead of fully decoding and thereby implicitly instantiating ArrayConvertibles, they should just be jsonDecoded and passed to the job in their array-form.
Exceptions will be thrown at job-execution inside the worker as it is with non-delayed jobs.

### DIFF
--- a/library/CM/Elasticsearch/Type/Abstract.php
+++ b/library/CM/Elasticsearch/Type/Abstract.php
@@ -284,10 +284,10 @@ abstract class CM_Elasticsearch_Type_Abstract extends CM_Class_Abstract implemen
             return;
         }
         $job = new CM_Elasticsearch_UpdateDocumentJob();
-        $job->queue(array(
+        $job->queue(CM_Params::factory([
             'indexClassName' => get_called_class(),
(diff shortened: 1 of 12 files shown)


njam:
I'm not sure if this is a good idea.
Currently `Job::queue($params)` will always be passed an array of *decoded* parameters, of key-value pairs. This changeset changes that - the function will have to accept also *encoded* parameters in the future.
This works I guess because `Params::encode()` which is used in `queue()` is idempotent.
Isn't it a bit risky? Or should at least be documented for `queue()`?
Alternatively how about decoding it, but handling exceptions more gracefully?

njam:
As discussed: let's pass `CM_Params` instead of `array` to queue() - so it can be encoded or decoded.

alexispeter:
please review @njam

njam:
To keep it simple, let's try/catch the decoding of the params in `queueOutstanding()`, and log a warning if it fails, but continue otherwise.

alexispeter:
Closed in favour of #2581

PR #2575, opened 2017-02-15
Make sure events "foatbox-open" and "-close" don't bubble

njam:
Addressing https://github.com/cargomedia/sk/issues/4534 

jQuery's `trigger()` triggers a native event on the DOM that bubbles up through the element hierarchy. We usually bind on specific elements about these events. So if a floatbox *inside* such an element is opened/closed we wrongly assume the outer floatbox got opened/closed.

The function `triggerHandler()` only triggers jQuery-internal handlers, and thus doesn't bubble.

@christopheschwyzer please review/test

### DIFF
--- a/client-vendor/after-body/jquery.floatbox/jquery.floatbox.js
+++ b/client-vendor/after-body/jquery.floatbox/jquery.floatbox.js
@@ -96,7 +96,7 @@
       this.$floatbox.trap();
 
       this.$layer.data('floatbox', this);
-      $element.trigger('floatbox-open');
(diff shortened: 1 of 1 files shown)


NicolasSchmutz:
btw this also happens in cam shows when closing the payment info floatbox of a "live" show.

njam:
@NicolasSchmutz I deleted your comment as it's CM and a pull request. Could you note it down in https://github.com/cargomedia/sk/issues/4544 ?

PR #2577, opened 2017-02-16
Provide more info via CM_Http_ClientDevice

zazabe:
Rename `CM_Http_ClientDevice` into `CM_Http_ClientInfo` and retrieve more info about the client (remote IP address for now).
Pretty useful to have access to such info via the `CM_Frontend_Environment`, ie. from a component.

cc @njam

### DIFF
--- a/library/CM/Frontend/Environment.php
+++ b/library/CM/Frontend/Environment.php
@@ -23,18 +23,18 @@ class CM_Frontend_Environment extends CM_Class_Abstract {
     /** @var CM_Model_Currency|null */
     protected $_currency;
 
-    /** @var CM_Http_ClientDevice|null */
-    protected $_clientDevice;
+    /** @var CM_Http_ClientInfo|null */
+    protected $_clientInfo;
 
     /**
-     * @param CM_Site_Abstract|null     $site
-     * @param CM_Model_User|null        $viewer
-     * @param CM_Model_Language|null    $language
-     * @param DateTimeZone|null         $timeZone
-     * @param bool|null                 $debug
-     * @param CM_Model_Location|null    $location
-     * @param CM_Model_Currency|null    $currency
-     * @param CM_Http_ClientDevice|null $clientDevice
+     * @param CM_Site_Abstract|null   $site
+     * @param CM_Model_User|null      $viewer
+     * @param CM_Model_Language|null  $language
+     * @param DateTimeZone|null       $timeZone
+     * @param bool|null               $debug
+     * @param CM_Model_Location|null  $location
+     * @param CM_Model_Currency|null  $currency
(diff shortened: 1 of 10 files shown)


zazabe:
@fvovan @njam wdyt about this change?

library/CM/Http/ClientInfo.php
```diff
@@ -1,27 +1,38 @@
...
+        $this->_parser = new Agent($headerList);
         $this->_headerList = $headerList;
+        $this->_server = $request->getServer();
+    }
+
+    /**
+     * @return string|null
+     */
+    public function getIp() {
+        return isset($this->_server['remote_addr']) ? $this->_server['remote_addr'] : null;
```
njam: Use `Request::getIp()` instead?

library/CM/Frontend/Environment.php
```diff
@@ -200,16 +200,16 @@ public function getCurrency() {
...
      */
-    public function setClientDevice(CM_Http_ClientDevice $clientDevice = null) {
-        $this->_clientDevice = $clientDevice;
+    public function setClientInfo(CM_Http_ClientInfo $clientInfo = null) {
+        $this->_clientInfo = $clientInfo;
     }
 
     /**
-     * @return CM_Http_ClientDevice
+     * @return CM_Http_ClientInfo
```
njam: Docu: can be null

fvovan:
btw it seems it's breaking.
It could be good idea to check all usages in the app.

njam:
Ah, AC :‑O

I don't like the name "client info" too much, it's very generic.
In the context of "environment" I think "client device" would still be clearer. The IP address is an attribute of the client's device. wdyt?

library/CM/Http/ClientInfo.php
zazabe: but it's a `ip2long()` representation, not something like `x.x.x.x`

zazabe:
ok, let's keep the previous naming, would avoid conflict in the app too ;)

zazabe:
@fvovan please check again

library/CM/Http/ClientInfo.php
njam: I just meant to keep the extraction from SERVER in a single place (Request).
We could add another function to Request that returns the IP address with dots? It's anyway a bit strange that `getIp()` returns a number..

zazabe:
@fvovan sorry, i did some changes to use `Request::getIp()` as @njam suggested. it's ready of a review now :)

library/CM/Http/ClientInfo.php
zazabe: I did it with `Request::getIp($dotNotation = null)` to not break the app, is it acceptable?

PR #2578, opened 2017-02-16
Introduce CM_Jobdistribution_DelayedQueue::countJob()

zazabe:
Useful to check if a job is already scheduled

cc @njam

### DIFF
--- a/library/CM/Jobdistribution/DelayedQueue.php
+++ b/library/CM/Jobdistribution/DelayedQueue.php
@@ -43,7 +43,7 @@ class CM_Jobdistribution_DelayedQueue implements CM_Service_ManagerAwareInterfac
      * @return int
      */
     public function countJob(CM_Jobdistribution_Job_Abstract $job, array $params) {
-        $where = sprintf("`className` = '%s' AND `params` = '%s' AND executeAt > %s", get_class($job), CM_Params::encode($params, true), time());
+        $where = sprintf("`className` = '%s' AND `params` = '%s' AND executeAt >= %s", get_class($job), CM_Params::encode($params, true), time());
         return CM_Db_Db::count('cm_jobdistribution_delayedqueue', $where);
     }
 
--- a/library/CM/Jobdistribution/DelayedQueue.php
+++ b/library/CM/Jobdistribution/DelayedQueue.php
@@ -43,8 +43,10 @@ class CM_Jobdistribution_DelayedQueue implements CM_Service_ManagerAwareInterfac
      * @return int
      */
     public function countJob(CM_Jobdistribution_Job_Abstract $job, array $params) {
-        $where = sprintf("`className` = '%s' AND `params` = '%s' AND executeAt >= %s", get_class($job), CM_Params::encode($params, true), time());
-        return CM_Db_Db::count('cm_jobdistribution_delayedqueue', $where);
+        return CM_Db_Db::count('cm_jobdistribution_delayedqueue', [
+            'className' => get_class($job),
+            'params'    => CM_Params::encode($params, true),
+        ]);
     }
 
     public function queueOutstanding() {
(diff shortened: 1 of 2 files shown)


library/CM/Jobdistribution/DelayedQueue.php
```diff
@@ -37,6 +37,16 @@ public function cancelJob(CM_Jobdistribution_Job_Abstract $job, array $params) {
         ]);
     }
 
+    /**
+     * @param CM_Jobdistribution_Job_Abstract $job
+     * @param array                           $params
+     * @return int
+     */
+    public function countJob(CM_Jobdistribution_Job_Abstract $job, array $params) {
+        $where = sprintf("`className` = '%s' AND `params` = '%s' AND executeAt >= %s", get_class($job), CM_Params::encode($params, true), time());
```
fvovan: I'm not a big fan of  constructing queries  this way. `className` and `executeAt` are more or less safe (btw `executeAt` can be `%d` as it's a number), but `params` looks potentially dangerous.
On the other hand it's impossible to prepare this query using `array` argument as a where clause because of `>=` operator.
@tomaszdurka, @njam - if we could pass some basic sql operators ( `>`, `<`, `<=`, `>=` and probably `like`) via array in "mongoDB fashion" ie 
```
[
  'className' => get_class($job),
  'params'       => CM_Params::encode($params, true),
  'executeAt'  => [
      '>=' => time(),
  ]
]
``` 
wdyt?

tomaszdurka: Well you could always:
```php
$mysqlClient = CM_Service_Manager::getInstance()->getDatabases()->getRead();
$count = $mysqlClient->createStatement('SELECT COUNT(*) FROM cm_jobdistribution_delayedqueue WHERE className = ?')->execute(['className'])->fetchColumn();
```

fvovan: but  it's simpler to use `CM_Db_Db::exec()`  then

zazabe:
wondering if this PR makes sense...  this simple implementation based on `executeAt` is not working, it really depends on when `queueOutstanding()` is called... so `executeAt >= time()` is not necessarily retrieving _all not executed jobs_...

Maybe we should store the job status in DB also, but it means tracking the gearman `job_handle` id and use [`jobStatus`](http://php.net/manual/en/gearmanclient.jobstatus.php) I guess... wdyt @fvovan @tomaszdurka ?

zazabe:
well... as discussed with @tomaszdurka, let me remove this `executeAt` condition... (thought we are keeping jobs in db, but we are actually removing them!!!)...

let me fix that, and use `?` for query params :)

fvovan:
>let me fix that, and use ? for query params :)

you can just pass the array. 
```
[
  'className' => get_class($job),
  'params'       => CM_Params::encode($params, true),
]
```

zazabe:
@fvovan  👍

zazabe:
BTW, would be nice to support something like 
```
new CM_Db_Query($client, 'table', '`col1` > ? AND `cal2` LIKE ?', [123, 'bar']);
```

zazabe:
@fvovan please check again ;)

njam:
> BTW, would be nice to support something like 
```
new CM_Db_Query($client, 'table', '`col1` > ? AND `cal2` LIKE ?', [123, 'bar']);
```

To create a query one needs an existing database connection. I guess some escaping rules depend on the server version etc.  `DB::createStatement()`  uses `PDO::prepare()`.

tomaszdurka:
I would prefer using raw pdos for these instead of introducing new things to cm-db? Or something already written - like Doctrine. DIsagree? :)

fvovan:
doctrine is tooooo heavyweight.

tomaszdurka:
I agree even DBAL which I though of has lot of features we don't need, like caching and some mapping capabilities. Also ORM.

PR #2580, opened 2017-02-17
Make socket-redis secret param optional

tomaszdurka:


tomaszdurka:
@fvovan please review

Socket-redis server have secret param as optional, so I think client should be aligned.

tomaszdurka:
@fvovan please check again

Small refactoring.

fvovan:
lgtm

PR #2581, opened 2017-02-17
Catch and log exceptions occurring when decoding a delayed job's parameters

alexispeter:
fixes https://github.com/cargomedia/sk/issues/4547

### DIFF
--- a/library/CM/Jobdistribution/DelayedQueue.php
+++ b/library/CM/Jobdistribution/DelayedQueue.php
@@ -60,7 +60,7 @@ class CM_Jobdistribution_DelayedQueue implements CM_Service_ManagerAwareInterfac
                 $paramsEncoded = $row['params'];
                 try {
                     $params = CM_Params::decode($paramsEncoded, true);
-                } catch (CM_Exception $ex) {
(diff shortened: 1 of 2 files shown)


library/CM/Jobdistribution/DelayedQueue.php
```diff
@@ -53,9 +53,21 @@ public function queueOutstanding() {
...
-            $job = $this->_instantiateJob($row['className']);
+            $jobName = $row['className'];
+            $job = $this->_instantiateJob($jobName);
             if ($job) {
-                $job->queue(CM_Params::decode($row['params'], true));
+                $params = null;
+                $paramsEncoded = $row['params'];
+                try {
+                    $params = CM_Params::decode($paramsEncoded, true);
+                } catch (CM_Exception $ex) {
```
njam: Shouldn't we catch *any* exceptions here? So that processing continues no matter what?

alexispeter: True.

njam:
@fvovan could you review this?

fvovan:
lgtm

Issue #2524, opened 2017-01-25
Don't fork php process


alexispeter:
To avoid problems with resource sharing (e.g. MongoDB socket https://github.com/cargomedia/sk/issues/4424).

There are two usages:
- CLI Manager. Can be replaced by letting systemd run multiple processes. https://github.com/cargomedia/puppet-cargomedia/issues/1772
- Clockwork Manager. Could be replaced by running events in job-distribution. https://github.com/cargomedia/cm/issues/2525

fauvel:
As a best practice, we should also close these resources just before forking (especially MySQL and MongoDB connections), otherwise the fastest process would close them unexpectedly for the other ones when it finishes.

njam:
We call `CM_Service_Manager::getInstance()->resetServiceInstances();` in `_fork()` to make sure all connections to databases etc are re-established. But in the future we could even remove `_fork()`.

PR #2584, opened 2017-02-21
Catch CLI exception and return exit code 1

zazabe:


library/CM/Cli/CommandManager.php
```diff
@@ -299,8 +299,14 @@ protected function _getProcessWorkload($transactionName, CM_Cli_Command $command
                 } else {
                     $this->_outputError($this->getHelp());
                 }
+                exit(1);
```
njam: This function should not exit() I think, only `bin/cm` should.
Have a look at `CommandManager::run()` which returns the exit code to use.

zazabe: not sure to get it, the cmd is running in a fork, i think the only way to change the exit code of the CLI is via `CM_Process_Result::$_returnCode`, which is set in `CM_Process::_wait`, based on the child exit code (`pcntl_wexitstatus`)...
or you mean invalid arguments should not use an exit code = 1 ?

njam:
We anyway want to remove forking capability from CLI-Manager (see https://github.com/cargomedia/cm/issues/2524). It can be done today - no CLI command requires forking anymore. Let's do that, it will make this much simpler.
Do you want to do it here? Or should we open a separate ticket?

zazabe:
hum... just spent some time making tests working with forked cmd... about removing the forking, i think it's better to create a new PR

njam:
Closing in favor of https://github.com/cargomedia/cm/issues/2585

PR #2586, opened 2017-02-23
Disable foreign keys check in MySQL dump

zazabe:
To avoid error due to tables ordering / constraints

### DIFF
--- a/library/CM/Db/Db.php
+++ b/library/CM/Db/Db.php
@@ -342,7 +342,8 @@ class CM_Db_Db extends CM_Class_Abstract {
             }
         }
 
-        $dump = 'SET SQL_MODE="NO_AUTO_VALUE_ON_ZERO";' . PHP_EOL;
+        $dump = 'SET FOREIGN_KEY_CHECKS=0;' . PHP_EOL;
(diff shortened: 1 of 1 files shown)


zazabe:
@tomaszdurka please review

zazabe:
The main reason to do that is described here: https://bugs.mysql.com/bug.php?id=16181

If there's a mistake in the SQL dump (ie. constraint with a non existing reference table), the dump will still be loaded with no errors (until the table is used)... but i'm not sure if we can really avoid that, see: http://stackoverflow.com/questions/3727905/check-for-referential-integrity-break

Anyway, if `structure.sql` files are always generated via cli, we should not have this problem...

zazabe:
Note that without this change, using `bin/cm db db-to-file` in the app will generate a dump that can't be loaded with `CM_DB_DB::runDump()`

tomaszdurka:
Yeah I get it all. I see that you also need to do some changes outside this method in order to support this "finally" concept i've proposed.

Let's drop the idea. We anyway load this dump when recreating application, meaning if this fails (and keeps foreign check in invalid state) we will run it again and fix it.

zazabe:
@tomaszdurka mergeable?

zazabe:
@tomaszdurka ?

tomaszdurka:
Yes!

PR #2588, opened 2017-02-23
FB-118 Make $encoded param of CM_Params mandatory

### LINKED ISSUE #2574: Make $encoded param of CM_Params mandatory
It's important to think about what the value of this parameter is, so let's make it mandatory.
It can still optional if $params is empty.

alexispeter:
closes #2574

### DIFF
--- a/library/CM/Params.php
+++ b/library/CM/Params.php
@@ -11,11 +11,10 @@ class CM_Params extends CM_Class_Abstract implements CM_Debug_DebugInfoInterface
      * @throws CM_Exception_Invalid
      */
     public function __construct(array $params = null, $encoded = null) {
-        $params = (null !== $params) ? $params : [];
-        if (null === $encoded) {
-            if (!empty($params)) {
-                throw new CM_Exception_Invalid('Params must be declared encoded or decoded');
-            }
+        if (null === $params) {
+            $params = [];
+        } elseif (null === $encoded) {
+            throw new CM_Exception_Invalid('Params must be declared encoded or decoded');
         }
         $encoded = (boolean) $encoded;
         $this->_params = \Functional\map($params, function ($value) use ($encoded) {
--- a/library/CM/Params.php
+++ b/library/CM/Params.php
@@ -11,7 +11,7 @@ class CM_Params extends CM_Class_Abstract implements CM_Debug_DebugInfoInterface
      * @throws CM_Exception_Invalid
      */
     public function __construct(array $params = null, $encoded = null) {
-        $params = $params ?: array();
+        $params = (null !== $params) ? $params : [];
         if (null === $encoded) {
             if (!empty($params)) {
                 throw new CM_Exception_Invalid('Params must be declared encoded or decoded');
--- a/library/CM/Params.php
+++ b/library/CM/Params.php
@@ -2,17 +2,22 @@
 
 class CM_Params extends CM_Class_Abstract implements CM_Debug_DebugInfoInterface {
 
+    /** @var array */
     private $_params;
 
     /**
-     * @param array|null $params
-     * @param bool|null  $encoded Defaults to true
+     * @param array|null   $params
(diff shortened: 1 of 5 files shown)


njam:
@fvovan please review

tests/library/CM/ParamsTest.php
```diff
@@ -41,7 +49,7 @@ public function testMerge() {
     }
 
     public function testHas() {
-        $params = new CM_Params(array('1' => 0, '2' => 'ababa', '3' => new stdClass(), '4' => null, '5' => false));
+        $params = new CM_Params(array('1' => 0, '2' => 'ababa', '3' => new stdClass(), '4' => null, '5' => false), false);
```
fvovan: May be add assertions for encoded Params as well?

tests/library/CM/ParamsTest.php
```diff
@@ -52,14 +60,14 @@ public function testHas() {
...
-        $params = new CM_Params(['foo' => ['_class' => 'Some_Nonexistent_Class', '_id' => 123]]);
+        $params = new CM_Params(['foo' => ['_class' => 'Some_Nonexistent_Class', '_id' => 123]], true);
         $this->assertSame(['_class' => 'Some_Nonexistent_Class', '_id' => 123], $params->get('foo'));
     }
 
     public function testGetString() {
         $text = "Foo Bar, Bar Foo";
         $notText = new stdClass();
-        $params = new CM_Params(array('text1' => CM_Params::encode($text), 'text2' => $text, 'text3' => $notText));
+        $params = new CM_Params(array('text1' => CM_Params::encode($text), 'text2' => $text, 'text3' => $notText), false);
```
fvovan: What if we set it to `true` at least for 1-2 test cases related to getters?
I undestand that it should work for both but also I see `_get()` method involves decoding in this case which probalby should be covered as well

tests/library/CM/ParamsTest.php
```diff
@@ -15,8 +23,8 @@ public function testMerge() {
             'val' => 2,
         ]);
 
-        $params1 = new CM_Params(['foo' => 1, 'bar' => 'text', 'object1' => $object1]);
-        $params2 = new CM_Params(['foz' => 2, 'bar' => 'other', 'object2' => $object2]);
+        $params1 = new CM_Params(['foo' => 1, 'bar' => 'text', 'object1' => $object1], false);
```
fvovan: I would make one of it decoded.

library/CM/Params.php
```diff
@@ -2,17 +2,22 @@
...
-     * @param bool|null  $encoded Defaults to true
+     * @param array|null   $params
+     * @param boolean|null $encoded
+     * @throws CM_Exception_Invalid
      */
     public function __construct(array $params = null, $encoded = null) {
         $params = $params ?: array();
         if (null === $encoded) {
-            $encoded = true;
+            if (!empty($params)) {
```
fvovan: Is the throwing on `$params = []` desired?

alexispeter: Not sure what you mean, an empty array is deemed empty by `empty()` and so would not throw.

fvovan: I just meant that it could be confusing to throw on `[1]` but pass `[]` especcially if the array is passed from outside and we are not sure what's inside. but not too critical

alexispeter: Ok you had me confused there. I thought you meant that the current implementation makes empty arrays throw and asked if this was on purpose or a mistake.

So you think it should also throw when the array is empty? I agree, instantiating a new object with an empty array instead of not passing anything seems a bit weird.

fvovan: >So you think it should also throw when the array is empty?

yes.

alexispeter:
please review last commit @fvovan

fvovan:
lgtm

Issue #2564, opened 2017-02-07
Rethink how we generate Urls


zazabe:
Maybe remove League\Url and use GuzzleHttp instead because of performance issues

related to #2535, https://github.com/cargomedia/cm/pull/2563 (revert) and https://github.com/cargomedia/cm/pull/2562

zazabe:
Guzzle\Psr7\Uri  seems not an option, mostly everything is `private`, would be hard to extend it...
see https://github.com/guzzle/psr7/issues/37 and https://github.com/guzzle/psr7/pull/72#issue-141914805

njam:
I see, so we copy it?

PR #2591, opened 2017-02-24
FB-105 Use Guzzle\Uri like class to build CM Urls

zazabe:
related to https://github.com/cargomedia/cm/issues/2564

cc @njam


zazabe:
retest this please

zazabe:
before ee3d5cd
- testSiteListSorted sorting duration: `19.42ms`

after ee3d5cd
- testSiteListSorted sorting duration: `3.77ms`

library/CM/Url/Vendor/Uri.php
```diff
@@ -0,0 +1,568 @@
+<?php
+namespace CM\Url\Vendor;
+
+use Psr\Http\Message\UriInterface;
+
+/**
+ * Modified version from GuzzleHttp, with protected Uri properties
```
njam: Upgrade this to the newest GuzzleHttp version 1.4.1?

library/CM/Url/Vendor/Uri.php
```diff
@@ -0,0 +1,568 @@
...
+ * @see https://github.com/guzzle/psr7
```
njam: I would suggest to link directly to the source file, at the version we have it from.
E.g. https://github.com/guzzle/psr7/blob/1.4.1/src/Uri.php

This way it's easier in the future to understand where the file is coming from and how to update it.

njam:
|`library/CM/Frontend/Render.php:191`|
|---|
@{f4355e77-07da-4ca6-b033-d106f96ec3bb,Michel Meyer} Maybe a `Environment::withSite($site)` would be good :) Next time!

njam:
|`library/CM/Frontend/Render.php:239`|
|---|
@{f4355e77-07da-4ca6-b033-d106f96ec3bb,Michel Meyer} I believe this should always be "serviceworker", and could be inlined (should match what's hardcoded in `CM_Http_Response_Resource_Javascript_ServiceWorker`)

library/CM/Http/Response/Abstract.php
```diff
@@ -83,6 +85,17 @@ public function getSite() {
     }
```
njam: |`line 90`|
|---|
Shouldn't this function rather be on the *Request*?

njam:
@{f4355e77-07da-4ca6-b033-d106f96ec3bb,Michel Meyer} could be nice to pass a *Url* instance here, and extrcat host+path inside, wdyt?

library/CM/Janus/Server.php
```diff
@@ -38,8 +41,8 @@ public function __construct($serverId, $key, $httpAddress, $webSocketAddress, ar
         }
```
njam: |`line 85`|
|---|
Use `withParams()` instead?

library/CM/Site/Abstract.php
```diff
@@ -111,24 +122,33 @@ public function getThemes() {
...
-     * @return string
+     * @return BaseUrl
      */
     public function getUrl() {
-        return (string) self::_getConfig()->url;
+        if (null === $this->_url) {
+            $this->_url = BaseUrl::create($this->getConfig()->url);
+        }
+        return $this->_url;
     }
```
njam: As discussed, let's make it `getUrlString()` instead for reusability

njam:
|`library/CM/Url/AbstractUrl.php:54`|
|---|
This seems a bit unsafe. The field can be NULL when neither `withTrailingSlash()` nor `withoutTrailingSlash()` was called.
Also it needs to be set whenever the path is set.
How about calculating it here based on the path, but cache it, and have a single function to invalidate the cache?

library/CM/Url/RouteUrl.php
```diff
@@ -3,53 +3,18 @@
 namespace CM\Url;
```
njam: This class is only used by `getUrlEmailTracking()` now.
It will become fully obsolete with https://youtrack.cargomedia.ch/issue/FB-115

How about inlining this code into `getUrlEmailTracking()`, or rename it to `EmailTrackingUrl`?

njam:
|`library/CM/Url/ServiceWorkerUrl.php:18`|
|---|
Could be hardcoded I think

njam:
|`library/CM/Url/UrlInterface.php:10`|
|---|
What's the prupose of this?
We don't use the Interface anywhere (e.g. type hinting) - should we?

tests/library/CM/Site/SiteFactoryTest.php
```diff
@@ -6,6 +6,24 @@ public function tearDown() {
         CMTest_TH::clearEnv();
```
njam: |`line 10`|
|---|
Could we use http://php.net/manual/en/function.shuffle.php ?

zazabe:
well, `setSize()` was implemented already

tests/library/CM/Site/SiteFactoryTest.php
zazabe: I wanted to keep item in the exact same "disordering", to test the constructor performance. 
Should i change it now?

zazabe:
I think it's nice to clearly describe CM specific methods added to the PSR-7 interface

But it's not strictly needed, right. 

Should I remove it?

library/CM/Url/RouteUrl.php
zazabe: btw, https://youtrack.cargomedia.ch/issue/FB-115 will also make ResourceUrl obsolete I guess...

zazabe: well... it's used in `CMTest_TestCase::createRequestFormAction()/createRequestAjax()` also... 
and if i rename `RouteUrl`, i would have to introduce a new `somethingUrl` to generate the path with the prefix and the language at the correct position.

I would suggest to keep it like that for now, and  to do it in https://youtrack.cargomedia.ch/issue/FB-115, when we will have to merge `RouteUrl` and `ResourceUrl` anyway

wdyt?

zazabe:
I don't think we really need to cache the result...

zazabe:
Agree, but it implies much more changes to retrieve a Url instance all the time (eg. `CM_Http_Response_Page::getRedirectUrl()`, `CM_Http_Request_Abstract::getUri()`, `CM_Frontend_Render::getUrl*()`), i think we should do this "deeper Url integration" in another PR.

zazabe:
see https://upsource.cargomedia.ch/cm/review/CM-CR-24?commentId=65e0a2e6-c2cd-4ee8-a030-f8f5a656d4d9

library/CM/Site/Abstract.php
```diff
@@ -111,24 +119,37 @@ public function getThemes() {
...
 
     /**
-     * @return string
+     * @return BaseUrl
```
njam: Add string casting

library/CM/Site/Abstract.php
```diff
@@ -111,24 +119,37 @@ public function getThemes() {
...
      */
     public function getUrl() {
-        return (string) self::_getConfig()->url;
+        if (null === $this->_url) {
+            $this->_url = BaseUrl::create($this->getUrlString());
+        }
+        return $this->_url;
```
njam: Add string casting

njam:
I would rather use the Interface in method signatures where possible (instead of the class). We can leave it as-is for now in my opinion.

njam:
A response doesn't inherently have a URL. We can retrieve a response's request, and from that request we can retrieve the URL.

So subsequently the original `getUrl()` function should be `Request::getUrl()`.
We can then still proxy it through a `Response:.getUrl()`.

This seems to be a very easy to accomplish change, we just move the function to the request?

library/CM/Http/Response/Abstract.php
njam: A response doesn't inherently have a URL. We can retrieve a response's request, and from that request we can retrieve the URL.

So subsequently the original `getUrl()` function should be `Request::getUrl()`.
We can then still proxy it through a `Response:.getUrl()`.

This seems to be a very easy to accomplish change, we just move the function to the request?

tests/library/CM/Site/SiteFactoryTest.php
njam: I see. We can leave it.

library/CM/Http/Response/Abstract.php
zazabe: I meant that adding `Request::getUrl()::Url` alongside with Request::getUri()/getHost()/getPath()/getQuery()` seems strange, all those methods should be removed, but it implies much more changes then...
But if we accept to keep them living altogether for now, it is very easy.

njam:
As discussed, let's revert df5823b10f0b5e381e3e6f2ad3eb86d2c142f06d and then it's good

PR #2594, opened 2017-03-09
FB-122 Update favicon/touch icons

christopheschwyzer:


layout/default/resource/browserconfig.xml.smarty
```diff
@@ -2,10 +2,7 @@
 <browserconfig>
   <msapplication>
     <tile>
-      <square70x70logo src="{resourceUrl path='img/meta/tile-small-128x128-transparent.png' type='layout'}"/>
```
njam: Do you have some reference about this?
Why are we delivering a file of 270x270, for a slot called "square 150x150logo"? For Retina?

Official docu: https://msdn.microsoft.com/en-us/library/dn320426(v=vs.85).aspx

christopheschwyzer: They actually recommend larger images. https://msdn.microsoft.com/library/dn455106(v=vs.85).aspx I guess for retina, etc.

> Note  To cover a wide range of devices, use an image 1.8 times the standard tile size so the image can be scaled up or down as needed.

Interesting factor (1.8)

njam: ok let's leave it.
but should we have the square310x310logo version?

christopheschwyzer: You mean additionally or instead? I understood that it uses 150x150 as default. Only difference is that you can't enlarge the home tile beyond default on metro. I think it's fine as it will be hardly used anyways?!
screen shot 2017-03-11 at 10 32 18

njam: I was just testing it recently on a friend's Windows Phone (using Denkmal), probably that's why I'm asking. It worked well.
If the default size still works, lgtm!

PR #2595, opened 2017-03-14
Fixed MultipartUploader usage

fauvel:
See http://docs.aws.amazon.com/aws-sdk-php/v3/guide/service/s3-multipart-upload.html#the-multipartuploader-object


alexispeter:
Adding a unit-test for writeStream() would probably be a good idea. If only to make sure the Uploader-object is built correctly.

fauvel:
Tagged 1.240.1

fauvel:
Mocking the upload builder is really complicated unfortunately...

Issue #2505, opened 2017-01-18
Storing preferences per site in persistence


tomaszdurka:
As discussed the implementations of the SiteSettings we decided to simplify approach and extend site with the model.
During discussion with @fvovan we came to some conclusions, also on the code level.

We will keep the persistence in mongo (for flexibility), but remove the flexibility from code-schema and define all the fields within the Site's `CM_Model_SchemaDefinition`.

Code snippets below:
```php
<?php

abstract class CM_Site_Abstract extends CM_Model_Abstract implements CM_ArrayConvertible, CM_Typed, CM_Comparable {

    protected function _getSchema() {
        return new CM_Model_Schema_Definition([
            'type' => 'int',
        ]);
    }

}

abstract class SK_Site_Abstract extends CM_Site_Abstract  {

    protected function _getSchema() {
        return new CM_Model_Schema_Definition([
            'type' => 'int',
            'emailSupport' => 'string'
        ]);
    }

}

class CM_Site_Dummy extends CM_Site_Abstract {}
class FB_Site_Admin extends SK_Site_Abstract {}

$config->CM_Site_Abstract->types[10] = CM_Site_Dummy::class;
$config->CM_Site_Abstract->types[100] = FB_Site_Admin::class;
```

MongoDB collection:
```json

[
    {
        "id": "mongo-id",
        "siteId": 1,
        "type": 10,
        "host": "www.example.org",
    },
    {
        "id": "mongo-id",
        "siteId": 2,
        "type": 100,
        "host": "admin.example.org",
        "emailSupport": "value"
    }
]
```

``` php

$site = new FB_Site_Admin(2);

$site = $sitesFactory->findSiteById(1);
$site = $sitesFactory->findSiteById(2);
$site = $sitesFactory->findSiteByHost('www.example.org');

PR #2596, opened 2017-03-14
FB-214: Site as a model

fvovan:
It's the second option of https://github.com/cargomedia/cm/issues/2505 solution


library/CM/FormField/Site.php
```diff
@@ -2,6 +2,7 @@
 
 class CM_FormField_Site extends CM_FormField_Set_Select {
 
+    //TODO fix method
```
njam: todo

fvovan:
It's more or less draft again (i mean going further with sk changes most likely will require to adjust something here)
first part of SK changes is here https://upsource.cargomedia.ch/sk/review/SK-CR-383 (not even close to be finished)

As  for me, maybe i would prefer separate models for site settings as i did in https://upsource.cargomedia.ch/cm/review/CM-CR-11. Site models concept itself is not really clear   since (as we  discussed) it will be anyway impossible to create new site without any application release.
and besides there are some inconveniences.

1) We'll lose the ability to refer to site objects with their constructor calls. So each `new FB_Site_Fuckbook_Cam()` will be changed to something like `(new CM_Site_Factory())->getSiteByType(FB_Site_Fuckbook_Cam::getTypeStatic())`. I.e. we need either siteID (and it's not clear how to obtain it in general case), or `siteType`.
2) we'll have to migrate 9 mysql tables and 1 mongo collection to contain site id. Though it could be `siteType` as it's now (since we agreed `type` will be unique)

And it the end we get back only database schemas which probably can be implemented in separate from `CM_Site_Abstract` classes.

fvovan:
As discussed with @{c7e8ab9c-90bf-4796-bb88-9aa7c8150044,Reto Kaiser},
1) use `FB_Site_Fuckbook_Cam::factory()` instead of constructors
2) keep db for now
3) implement admin part of settings later
4) for sk part move to db only basic settings (`name` and `email`)
5) look into example data scripts later when sk part is near to be finished

library/CM/Model/Schema/Definition.php
```diff
@@ -124,7 +124,12 @@ public function decodeField($key, $value) {
                                 throw new CM_Model_Exception_Validation('Field type is not a valid class/interface', null, ['type' => $type]);
                             }
                             $className = $type;
-                            if (is_a($className, 'CM_Model_Abstract', true)) {
+                            if (is_a($className, CM_Site_Abstract::class, true)) {
+                            /** @var CM_Site_Abstract $type */
+                                $id = $value;//TODO check if we need JSON encoded values
+                                $value = CM_Site_Abstract::factoryFromId($id);
+                            }
+                            elseif (is_a($className, 'CM_Model_Abstract', true)) {
```
njam: Formatting

library/CM/Model/Schema/Definition.php
```diff
@@ -124,7 +124,12 @@ public function decodeField($key, $value) {
                                 throw new CM_Model_Exception_Validation('Field type is not a valid class/interface', null, ['type' => $type]);
                             }
                             $className = $type;
-                            if (is_a($className, 'CM_Model_Abstract', true)) {
+                            if (is_a($className, CM_Site_Abstract::class, true)) {
+                            /** @var CM_Site_Abstract $type */
+                                $id = $value;//TODO check if we need JSON encoded values
```
njam: what about this todo?

library/CM/Site/Abstract.php
```diff
@@ -2,7 +2,7 @@
```
njam: |`line 357`|
|---|
Should we change it to `cm_site`?

njam:
|`library/CM/Site/Abstract.php:119`|
|---|
In this case I think it's better to invalidate all relevant caches in `setDefault()`. In the end we just need to invalidate the cache of thee *previous* default site, and of the *new* default site. Or if it's easier we can invalidate the caches of *all* sites.

njam:
|`library/CM/Site/Abstract.php:132`|
|---|
Alternatively how about getting the *previous* default site instance, ans setting `setDefault(false)` on it?
On the other hand, it would be nice if this was an atomic operation. Can it be done in a single mongodb query?

njam:
|`library/CM/Site/Abstract.php:119`|
|---|
In this case I think it's better to invalidate all relevant caches in `setDefault()`. In the end we just need to invalidate the cache of thee *previous* default site, and of the *new* default site. Or if it's easier we can invalidate the caches of *all* sites.

njam:
|`library/CM/Site/Abstract.php:281`|
|---|
Use getUrlString() for better performance

njam:
|`library/CM/Site/Abstract.php:337`|
|---|
Should we add caching for this?

njam:
|`library/CM/Site/Abstract.php:372`|
|---|
Let's replace it with a proper `public static function create($proper, $signature)`?

fvovan:
Not sure it's needed. We are not going to call `CM_Class_Abstract::create()` anyway. 
`_createStatic()` is for `SK_SIte_Abstract::create()` (and its descendants usages)

njam:
agree

fvovan:
may be but retrieval by ID should be very fast, as it stores indexes in RAM. but i need to check how it works with our new mongo

njam:
Still it's magnitudes slower than doing a lookup in memory.
Since this function is used when decoding models, we really need to assume it's used a lot - because we don't know ''when'' it's used.
I assume the "_type" of a site cannot be changed? If so then cache invalidation is not an issue, and we can simply cache this in "CacheLocal" - easy and impactful.

njam:
Still it's magnitudes slower than doing a lookup in memory.
Since this function is used when decoding models, we really need to assume it's used a lot - because we don't know ''when'' it's used.
I assume the "_type" of a site cannot be changed? If so then cache invalidation is not an issue, and we can simply cache this in "CacheLocal" - easy and impactful.

fvovan:
okay, but why not use `$id` as cache key?

njam:
Yes $id as cache key sounds good.

fvovan:
Currently it's not possible to *unset* the field (we need this to allow additional index level guard, setting value to `null` doesn't work btw).
MongoDB doesn't support atomicity across several documents update. How about just resetting the cache?

njam:
sgtm
btw, please don't "resolve" my comments - I'd like to use it for checking it once more, thanks

fvovan:
ah. sorry. I thought it's a part of the upsource workflow

njam:
|`library/CM/Site/Abstract.php:120`|
|---|
How about just making this not optional? Seems it's mandatory.

njam:
|`library/CM/Site/Abstract.php:128`|
|---|
If we pass FALSE we can end up with no default anymore, which will be a problem. How about some sanity checks?

fvovan:
It can't be mandatory. The thing it's indexed as *unique* key in the collection, so `true` value may exist only once. The same is for `false` and `null`

fvovan:
I removed method parameter.

PR #2598, opened 2017-03-16
FB-173 Added GTM client and tracked split tests using the data layer

fauvel:


library/CMService/GoogleTagManager/Client.php
```diff
@@ -0,0 +1,69 @@
...
+
+    public function trackAction(CM_Action_Abstract $action) {
+    }
+
+    public function trackPageView(CM_Frontend_Environment $environment, $path) {
+        if ($viewer = $environment->getViewer()) {
+            $fixture = new CM_Splittest_Fixture($viewer);
+        } elseif (CM_Http_Request_Abstract::hasInstance()) {
+            $request = CM_Http_Request_Abstract::getInstance();
+            if ($request->hasClientId() || (int) $request->getCookie('clientId')) {
```
fauvel: @njam any idea how to make this better? I need to retrieve the client id if there is one, and I want to avoid creating one if not.

njam:

library/CMService/GoogleTagManager/Client.php
fauvel: The goal here is to retrieve all existing split test variations assigned either to the viewer or to the request client. If there's no clientId yet, there cannot be any variation assigned to the request client. We don't want to create a clientId on-the-fly here: We need to take into account performance considerations, because we're going to run this code for every single page view. We've already decided a long time ago not to create blindly one clientId for every request client, but only to do so when required.

library/CM/Http/ClientDevice.php
```diff
@@ -37,6 +37,13 @@ public function getIp($dotNotation = null) {
     }
 
     /**
+     * @return CM_Http_Request_Abstract
+     */
+    public function getRequest() {
```
njam: I think we should avoid exposing the Request here for two reasons:
- We expose too much, which adds unnecessary dependencies
- What we want to access should not be on a Request reference in the first place (but maybe on a ClientDevice)?

How about we proxy what we need like we're doing it already with `getIp()`?

fauvel: So let's introduce `$clientDevice->getId($createIfNew = null)` instead? And proxy to `$this->_request->findClientId()` resp. `$this->_request->getClientId()` then? Wdyt?

fauvel:
Tagged 1.243.3

PR #2603, opened 2017-03-17
Trigger "navigate:*" events on any location change

njam:
Goal: trigger the `navigate:*` events not only on ajax navigation, but also on location changes when only the client-side state changes.


njam:
@alexispeter please review

alexispeter:
lgtm

PR #2605, opened 2017-03-21
FB-105 Use only the Site in CM_Frontend_Render::getUrl()

zazabe:


zazabe:
it produces an issue with the captcha, because the url is generated with the language.

otherwise we could change `CM_Http_Response_Captcha::createFromRequest()` to support the language, and maybe layout/default/resource/manifest.json.smarty also?

zazabe:
But the idea here is to keep the same behavior than before, anyway, those `CM_Frontend_Render::getUrl*()` will be removed soon i think...

PR #2606, opened 2017-03-22
FB-326 Streamline CM_Janus_Server Url getters

zazabe:


njam:
A unit test would be nice?

PR #2610, opened 2017-03-23
FB-173 Send custom Page View event to the GTM

fauvel:


fauvel:
We need to send a custom Page View event to the Google Tag Manager in order to fire tags upon ajax navigation.

fauvel:
Tagged 1.243.4

PR #2611, opened 2017-03-24
Fix line separators

alexispeter:


alexispeter:
Ran an inspection and found "a few more" issues.

PR #2612, opened 2017-03-24
FB-333 Add FileReader form field

christopheschwyzer:


christopheschwyzer:
@{7d7c0e0b-25a2-49ba-bfe9-94768ea36b50,Tomasz Durka} maybe we should discuss a few things.
- General question. Is it enough to do client-side form field validation? I guess not :) It's just we don't want to upload the file until the form is submitted, but there could/should be some validation beforehand (I already do this now). 
- In `FileReader.php` with `createFile()` function we have to create a file, but `new CM_File()` required a path, that we obviously don't have. Any idea how we could handle this?
- I separated `FileReader` and `ImageReader`. Imo `FileReader` should be able to handle mixed files (images and other files) that's why I didn't separate the template and validation. Wdyt?
- Like I mentioned in the task description. FileReader is limited by the client's memory, same goes for the ability to resize images on the client. I'm a bit worried b/c e.g. new phones create huge image files. (I noticed this on the MBP!!)

christopheschwyzer:
Compress data before uploading? https://github.com/devongovett/brotli.js

christopheschwyzer:
This has been idle for some time and I'm not sure how we should prioritise this. This came up while working on an image cropper solution (See https://upsource.cargomedia.ch/sk/review/SK-CR-149). I don't like how these two things are coupled now. 

Like mentioned above I'm a bit worried about possible implications. Mostly about "FileReader is limited by the client's memory".  
At this point I can't really make a decision on how to move forward. @{7d7c0e0b-25a2-49ba-bfe9-94768ea36b50,Tomasz Durka} @{c7e8ab9c-90bf-4796-bb88-9aa7c8150044,Reto Kaiser} advise or kill it. :)

njam:
Rejecting for now, as we'll most likely create a completely new frontend.

PR #2613, opened 2017-03-27
FB-325 Improve CM/Url perf with cache in CM_Frontend_Render

zazabe:


zazabe:
see https://youtrack.cargomedia.ch/issue/FB-325#comment=93-992

PR #2614, opened 2017-03-27
FB-267 Add global event handler and trigger events on cud-operations on models

alexispeter:


tomaszdurka:
Wanted to discuss whole concept again a bit.

I see you've added all those triggers in model methods directly. Instantiating instance in method and triggering a method with event-handler as argument.
- Isn't this a bit overcomplicated and complicated to test?

Additionally this brings this kind of functionality only to the models - which probably ain't bad. Wdyt about delegating crud operations to single instance only - retrieved using service-manager (at least for now).
Meaning that we would have a model - doing operation and delegating it to some kind of action processing "service". This action processing could then persist data, trigger events etc. Difference would be using Dependency Injection and injecting all those "event-handelrs" to the action-processing-service (I always feel bad in using singleton in models, so would prefer it only once).

Let me know if you want to discuss it heads-up :)

alexispeter:
I'm not sure what the difference would be. This action-processing service sounds to me like it is the same as "CM_ModelEvents_Trigger", triggering events on an injected eventHandler, the only difference being that this action-processor is registered on the serviceManager, which would "save" some object instantiations I guess?
if I misunderstood you, could you show me with pseudo-code how triggering a model-event would look like with your solution?

PR #2615, opened 2017-03-28
Reset config first in TH

njam:


njam:
This is to prevent that some config changes in tests will affect the cleanup of other parts like database.
Config is one of the most fundamental modules, so should be cleaned up first.

PR #2617, opened 2017-03-28
FB-360 Make sure the "variables"-field is set when creating a languageKey without variables

alexispeter:


library/CM/Model/LanguageKey.php
```diff
@@ -132,6 +132,7 @@ public static function create($name, array $variables = null) {
             'updateCount'             => 0,
             'updateCountResetVersion' => 0,
             'javascript'              => false,
+            'variables'               => null,
```
tomaszdurka: Shouldn't we rather make variables non-optional and set them to empty array instead?
Then we could drop the getter check and always expect array. 

I think it's cleaner but requires update script to change or nulls to arrays.

@{889dab99-4493-437d-87fa-92d5ed832ec7,Alexis Peter} wdyt?

alexispeter: Alright with me.

PR #2618, opened 2017-03-30
FB-369 set "executedAt" inside CM_Migration_Model::create()

alexispeter:


tests/library/CM/Migration/ModelTest.php
```diff
@@ -21,7 +20,7 @@ public function testCreate() {
         $this->assertSame([
             'tries'                    => 0,
             'originalExceptionMessage' => "SQLSTATE[23000]: Integrity constraint violation: 1062 Duplicate entry 'foo' for key 'name'",
-            'query'                    => 'INSERT INTO `cm_migration` (`name`) VALUES (?)',
+            'query'                    => 'INSERT INTO `cm_migration` (`name`,`executedAt`) VALUES (?,NULL)',
```
tomaszdurka: I thought it's a default value so we don't need it, but looks good.

library/CM/Migration/Model.php
```diff
@@ -55,6 +55,7 @@ protected function _getSchema() {
     public static function create($name) {
```
tomaszdurka: |`line 47`|
|---|
Again should this be optional if we expect this to be set aka _get to always work?

alexispeter: It should be optional because it can be null. Optional does not mean that it can be completely absent.

tomaszdurka:
Ok.
Hmm bit confused now. I am not recalling how exactly this works, but from what I understand:
- making field optional is only way to set it to null, when other type is defined
- _get method returns null if it has been set to null, but throw if not set at all?
- _has method is needed to distinguish 2 above?
- you setExecuteStamp on start to not need to _has as mentioned above?

Sorry for the confusion, i just wanted to understand it better.

alexispeter:
* Yes (but I would say it is the intended and the only way. optional means that this field may either be of the specified type or null).
* Yes.
* Yes and no, a field not being set is an illegal state for the object if you ask me.
* The reason I setExecuteStamp on start is so that the databag is in a valid state.

The point of the storage-adapter is to have an orm-style schema that is always adhered to. This means that data should be fetchable and storable by doing nothing more than _get(propertyName) and _set(propertyName, propertyValue) respectively. This also means that getters and setters are not permitted to do additional stuff like return null if the property is not set.
If doing that is a requirement, then a storageAdapter should not be used.

alexispeter:
* Yes (but I would say it is the intended and the only way. optional means that this field may either be of the specified type or null).
* Yes.
* Yes and no, a field not being set is an illegal state for the object if you ask me.
* The reason I setExecuteStamp on start is so that the databag is in a valid state.

The point of the storage-adapter is to have an orm-style schema that is always adhered to. This means that data should be fetchable and storable by doing nothing more than _get(propertyName) and _set(propertyName, propertyValue) respectively. By extension, this means that getters are not permitted to do additional stuff like return null if the property is not set.
If doing that is a requirement, then a schema should not be used.

I guess in the end the question is, should the absence of data be handled as if that data was null?

PR #2622, opened 2017-04-04
FB-89 Upgrade mongodb client

alexispeter:


tomaszdurka:
I've checked commits after the revert commit. Looks good.

PR #2627, opened 2017-04-21
FB-89 Added regression test for bug introduced in CM 1.246.0

alexispeter:


tests/library/CM/Clockwork/Storage/MongoDBTest.php
```diff
@@ -33,8 +36,25 @@ public function testDataStorage() {
         date_default_timezone_set($defaultTimeZoneBackup);
     }
 
+    public function testMongoClientUpdateRegression() {
+        $storage = new CM_Clockwork_Storage_MongoDB('example');
+        $event = new CM_Clockwork_Event('foo', '1sec');
+        $this->_assertSameStatus(new CM_Clockwork_Event_Status(), $storage->getStatus($event));
+        $status1 = (new CM_Clockwork_Event_Status())->setLastStartTime(DateTime::createFromFormat('U', time()));
```
tomaszdurka: For readabilty it could be nice to have this time as variable and then add 1000s to it using dateinterval.

PR #2629, opened 2017-05-02
FB-506 Set caret to end of contenteditable on focus

christopheschwyzer:


christopheschwyzer:
For a quick fix I've set the caret to the end of the contenteditable on focus. Unfortunately this is not the exact behavior of a native textarea. For textareas the caret position gets "stored": When you blur a textarea and re-focus via js it will position the caret at the previous position. Here is a fiddle: https://jsfiddle.net/Lat4hucd/ Should we try to mimic this behavior?

njam:
Why doesn't it happen when adding a status (in the "share") component? Ping me when you're back?

christopheschwyzer:
"Status" uses an `<input>` field.

christopheschwyzer:
Unfortunately I don't recall why I added `window.getSelection().removeAllRanges();`. I removed it again and it seems fine. 

I noticed a new issue surfaced with latest Edge. When you write some text and delete it again you end up with the placeholder on a newline.
file

christopheschwyzer:
I will go on and merge this, but we have to keep an eye on the "contenteditable" :(

PR #2632, opened 2017-05-08
FB-372 Added accountAge to the dataLayer

fauvel:


fauvel:
We're sending the account age to the dataLayer of the Google Tag Manager in order to be able to trigger Inspectlet for new accounts only (first day), see https://youtrack.cargomedia.ch/issue/FB-372

fauvel:
Tagged 1.246.6

PR #2635, opened 2017-05-11
FB-115 Change App Url format

zazabe:
- Extract site / language / version from Url path parts
- Use AppUrl in CM_Request
- Map CM_Response with AppUrl


zazabe:
@{c7e8ab9c-90bf-4796-bb88-9aa7c8150044,Reto Kaiser} here we are... this is the challenging PR i told you, about https://youtrack.cargomedia.ch/issue/FB-115...

- about Url (AppUrl, PageUrl, etc...): 7bb842c  to 2c20991
- about CM_Request / CM_Response: 2c20991 to 44a6739

njam:
There are some problems with `setUrlFromString()`, which needs to instantiate a specific type of URL. This is needed so that later the `Request::getUrl()` can be used to generate correct URLs.
Unfortunately there are way to many dependencies in this area of code (Request, Response, URL, etc), and it's uncertain if it makes sense / is possible to refactor it.

One idea would be to cut the dependencies at `CM_Http_Response_Abstract`, by ensuring it doesn't receive the `Request`.

njam:
Idea: 
- Keep single type of Url on Request (Url\Url)
- Move parsing of URL to different place (out of Url) - will provide functionality like `Request::popUrlEnvironment()`
- Move generating of URLs to different place (out of Url) - will provide functionality like `Render::getUrlLayout()`

PR #2637, opened 2017-05-15
FB-565 Throw FATAL exceptions when failing to persist maintenance-event results

alexispeter:


library/CM/Maintenance/RunEventJob.php
```diff
@@ -21,7 +21,13 @@ protected function _execute(CM_Params $params) {
             throw $e;
         } finally {
             $newRelic->endTransaction();
-            $maintenance->handleClockworkEventResult($eventName, $result);
+            try {
+                $maintenance->handleClockworkEventResult($eventName, $result);
+            } catch (CM_Exception_Invalid $e) {
```
njam: As discussed let's remove this empty catch. This situation (even name is unknown) can be fatal (or not), but we cannot know it - so better throw fatal.

PR #2639, opened 2017-05-24
FB-636 Use name-hashing to make languageKeys unique on the db-level

alexispeter:


njam:
|`library/CM/Model/LanguageKey.php:54`|
|---|
This should not be public API imo. There's no use case to set the hash, only the name?

njam:
|`library/CM/Model/LanguageKey.php:183`|
|---|
Would be good to have the original exception's info here, in case of debugging. How about just rethrowing?

alexispeter:
There is no use case to set the name only either. Or do you mean setName() should change the hash as well?

njam:
Yes, my thought was that setName() implicitely sets the hash (and fails if there's a duplicate). I think the hash mechanism can be completely private to this class and any callers of setName() don't need to know about it.

library/CM/Model/LanguageKey.php
```diff
@@ -92,17 +106,18 @@ protected function _getDeployVersion() {
     }
 
     protected function _getSchema() {
-        return new CM_Model_Schema_Definition(array(
-            'name'                    => array('type' => 'string'),
-            'variables'               => array('type' => 'string'),
-            'updateCountResetVersion' => array('type' => 'int', 'optional' => true),
-            'updateCount'             => array('type' => 'int'),
-            'javascript'              => array('type' => 'bool'),
```
njam: Use atomic `set()` for both "name" and "nameHash"

library/CM/Model/LanguageKey.php
```diff
@@ -92,17 +106,18 @@ protected function _getDeployVersion() {
...
-        ));
+        return new CM_Model_Schema_Definition([
+            'name'                    => ['type' => 'string'],
+            'variables'               => ['type' => 'string'],
+            'updateCountResetVersion' => ['type' => 'int', 'optional' => true],
+            'updateCount'             => ['type' => 'int'],
+            'javascript'              => ['type' => 'bool'],
+            'nameHash'                => ['type' => 'string'],
```
njam: |`line 32`|
|---|
Remove public getter.

library/CM/Model/LanguageKey.php
```diff
@@ -12,10 +12,19 @@ public function getName() {
     }
 
     /**
-     * @return string $name
+     * @param string $name
+     * @throws CM_Exception_Invalid
      */
     public function setName($name) {
-        $this->_set('name', $name);
+        $hash = self::calculateHash($name);
```
njam: The message is not necessarily correct. What if the application cannot connect to MySQL? We should avoid this to prevent misinterpretation later when debugging problems.

Either we need a more specific check of the exception, or just let the original exception throw?  Probably the latter is fine?

alexispeter: I would prefer if the exception was specific enough to be able to confirm it's exact cause, but yes, not catching in the first place is probable fine.

PR #2642, opened 2017-06-05
pass null to prevent returning

fvovan:
I noticed that sometimes index update sctipt doesn't work like expected
Empty array is passed to `CM_Elasticsearch_Type_Abstract::_updateIndex()`


fvovan:
as it turned out, it's expected behavior

PR #2650, opened 2017-06-27
FB-838 Install timecop for Travis

dlondero:


zazabe:
I guess we want to have it on our vagrant box, only in dev, so we need to add it to https://github.com/cargomedia/puppet-packages/blob/master/modules/cm/manifests/application.pp#L32

zazabe:
ho!
there's https://upsource.cargomedia.ch/puppet-packages/review/PP-CR-57 already ;)

PR #2653, opened 2017-06-29
FB-837 Update test helper to use timecop

dlondero:


dlondero:
@to

composer.json
```diff
@@ -30,8 +30,9 @@
     "mongodb/mongodb": "~1.0.0"
   },
   "require-dev": {
-    "phpunit/phpunit": "~5.5.0",
-    "tomaszdurka/mocka": "~0.12.3"
```
tomaszdurka: Why adding this now?

tests/library/CM/UtilTest.php
```diff
@@ -6,6 +6,8 @@ public function testBenchmark() {
         $this->assertSame(0.0, (float) CM_Util::benchmark());
         $this->assertSame(0.0, (float) CM_Util::benchmark('CM'));
 
+        CMTest_TH::timeForward(1);
```
tomaszdurka: Why this is needed now?

composer.json
dlondero: Right, not needed as phpunit has it `^3.2`.

tests/library/CM/UtilTest.php
dlondero: This is because `CM_Util::benchmark()` is based on `microtime` which is now frozen by timecop, while it wasn't when using runkit (only redefining `time` and `gettimeofday`).

PR #2654, opened 2017-06-29
FB-931 Add new site-config "robots-noindex" option

alexispeter:


alexispeter:
I saw you added new properties as optional and made it so they don't even need to exist in the document. I didn't really see why, so I added this one normally. I hope this doesn't break something that I'm not aware of.

fvovan:
Please don't forget to include it to `SK_Site_Abstract::_getschema()` as well. I thought about correct composition of these schemas to avoid the duplication but then we postponed it for now.

fvovan:
It was needed to instantiate models in migration. Otherwise I would have had to use mongoDB directly and it would have increased the probability of errors. But in general we plan to discuss the optionality of all fields. I tend to make them all mandatory.

PR #2656, opened 2017-06-30
FB-961 Improve performance of CM_Site_Model

fvovan:


fvovan:
It's the simplest thing that we may do now.
Idea is we don't need sorting of  30 sites every time.  If it doesn't help we can aggressively cache `$_siteList` and `$_siteListSorted` for 5-10 minutes (to not care of invalidation). @{c7e8ab9c-90bf-4796-bb88-9aa7c8150044,Reto Kaiser} wdyt?

njam:
Add 2 lazy functions:
- _getSiteListSorted()
- _getSiteList() <- Try caching the full array in Memcache

fvovan:
I would suggest trying to deploy it without caching for now. Got stuck on serialization. For some reason it loses `_modules` and `_themes` attributes when unserialized, and we can't quickly define `CM_Site_Abstract::unserialize()` (parent's method is final). @{c7e8ab9c-90bf-4796-bb88-9aa7c8150044,Reto Kaiser} wdyt?

library/CM/Site/SiteFactory.php
```diff
@@ -5,23 +5,13 @@ class CM_Site_SiteFactory {
     /** @var CM_Site_Abstract[] */
```
njam: |`line 5`|
|---|
Docu: can be null

njam:
sgtm

njam:
Small docu, otherwise lgtm.
Feel free to merge and release when adjusted

PR #2664, opened 2017-07-04
CM-971: invalidate site's cache on update

fvovan:


fvovan:
my only concern is if setting `_autocommit` to true in case of passed data can hurt.
Maybe it would be better to set it directly in `_factoryFromModelData()`. wdyt?

library/CM/Model/Abstract.php
```diff
@@ -39,6 +39,8 @@ public function __construct($id = null) {
         if (null === $id && null === $data) {
             $data = array();
             $this->_autoCommit = false;
+        } else {
+            $this->_autoCommit = true;
```
tomaszdurka: So this is in cases we use `_construct` manually after normal constructor (with _construct without id).
Again same in different words: I understand we first run the _construct and set autoCommit to false and then when we load data, we should treat it as existing and autoCommit.

Is that the reason?

tomaszdurka:
Again I think it's proper. We wanted autoCommit=false only for models which are newly instantiated without persistence storage.

library/CM/Model/Abstract.php
fvovan: Yes ` $model = new $className();` sets it to `false` and following call of `_construct` should set it to `true`. Otherwise all setters invoked from the results of `CM_Paging_Site_All` will not take effect without `commit()`

PR #2665, opened 2017-07-05
FB-982 Remove model multiget

alexispeter:


alexispeter:
This removes CM_Model::factoryGenericMultiple(), all implementations of multi-loading on storage adapters as well as some multi-data retrieval mechanisms implemented for that specific purpose.

PR #2667, opened 2017-07-07
FB-969 Don't rely on DNS check in unit tests

dlondero:


library/CM/Service/NetworkTools.php
```diff
@@ -0,0 +1,12 @@
+<?php
+
+class CM_Service_NetworkTools {
+
+    /**
+     * @param string $domain
+     * @return bool
+     */
+    public function getMXRecords($domain) {
```
fauvel: It is rather `hasMXRecords` if you only return a boolean and dismiss the host list...

fauvel:
Otherwise lgtm!

dlondero:
Fixed naming, updated tests.

fauvel:
Lgtm!

PR #2669, opened 2017-07-10
FB-988 remove trailing slashes in getPath()

fvovan:


fvovan:
It's a bit straightforward solution. but the thing is `parse_url()` doesn't work with `'//' ` string passed as a parameter  and we need to handle this case apart of other invalid urls.. @{c7e8ab9c-90bf-4796-bb88-9aa7c8150044,Reto Kaiser} wdyt?

PR #2671, opened 2017-07-12
FB-1019 Fix clear cache checkboxes in the debug component

fauvel:


fauvel:
The bug actually came from the refactoring we've done when introducing the root view `CM_View_Document` on Sep 28th 2016...

fauvel:
Tagged 1.257.0

PR #2672, opened 2017-07-13
FB-1012 Removed timezone functionality from CM_FormField_Birthdate

alexispeter:


library/CM/FormField/Birthdate.php
```diff
@@ -1,28 +1,67 @@
...
     }
 
     public function validate(CM_Frontend_Environment $environment, $userInput) {
-        $userInput = parent::validate($environment, $userInput);
-        $age = $userInput->diff(new DateTime())->y;
+        $dd = (int) trim($userInput['day']);
+        $mm = (int) trim($userInput['month']);
+        $yy = (int) trim($userInput['year']);
+
+        $birthdate = new DateTime($yy . '-' . $mm . '-' . $dd);
```
fvovan: we need to use `checkdate()` or call `DateTime::getLastErrors()` afterwards to validate these values. Otherwise is will pass invalid date like '1934-02-31'  and save it as 3rd March.

alexispeter: That's waaaaaaaaaaaaaay outside of the scope of this ticket. 
It would probably be a whole story of its own to make a user-friendly date-input field that does not allow invalid dates.

fvovan:
okay, if we need to create a separate ticket to add one line of code to fix it, let it be so

alexispeter:
Depending on how this check is performed (e.g. throw an exception), frontend aspects need to be considered. For one, we need to change the field in a way that makes invalid dates impossible, requiring js shenanigans to change the day-dropdown to match the month etc.
I simply do not want to include an unrelated "fix" that may or may not require additional work.

PR #2676, opened 2017-07-17
FB-1038 Move stat logging to ES client, log index names and types

tomaszdurka:


dlondero:
So this is just a preparation refactoring?

tomaszdurka:
This is all. Now I moved the logging intot the client and there I encode all params including index name, type and data.

dlondero:
Ok, then now you can finish the review I'm waiting for ^.^

PR #2678, opened 2017-07-18
FB-1038 Move stat logging to ES client, log index names and types

tomaszdurka:
Reverts cargomedia/cm#2677

### DIFF
--- a/library/CM/Elasticsearch/Client.php
+++ b/library/CM/Elasticsearch/Client.php
@@ -5,11 +5,16 @@ class CM_Elasticsearch_Client {
     /** @var Elasticsearch\Client */
     protected $_client;
 
+    /** @var CM_Debug|null */
+    private $_debug;
+
     /**
      * @param \Elasticsearch\Client $client
+     * @param CM_Debug|null         $debug
      */
-    public function __construct(Elasticsearch\Client $client) {
+    public function __construct(Elasticsearch\Client $client, CM_Debug $debug = null) {
         $this->_client = $client;
+        $this->_debug = $debug;
     }
 
     /**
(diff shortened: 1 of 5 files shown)


library/CM/Elasticsearch/Factory.php
```diff
@@ -0,0 +1,41 @@
...
+        $client = $this->createClient($servers, $debug);
+        return new CM_Elasticsearch_Cluster($client, $disabled);
+    }
+
+    /**
+     * @param array         $servers
+     * @param CM_Debug|null $debug
+     * @return CM_Elasticsearch_Client
+     */
+    public function createClient(array $servers, $debug = null) {
```
fvovan: Please add type hinting for `$debug`

library/CM/Elasticsearch/Client.php
```diff
@@ -264,12 +269,16 @@ public function search($indexNameList, $typeNameList, array $data) {
             'type'  => self::_prepareIndexNameParam($typeNameList),
             'body'  => $data,
         ];
+
+        if ($debug = $this->_getDebug()) {
+            $debug->incStats('search', json_encode($params));
```
fvovan: I think it's better to use CM_Utils::jsonEncode()

fvovan:
rest looks good, the only thing came to my mind is  probably we don't even need `Cluster` class now.
It's not a container anymore and all it does is just proxifying queries to `Client` (which is kind of proxy too). well also it allows to enable/disable cluster..

tomaszdurka:
Yes, but I didn't want to refactor it here. Or maybe.

tomaszdurka:
I will keep it, it has this "enabled" feature which we should keep for now.

PR #2680, opened 2017-07-20
FB-1079 Move email validation service to where we *send* the email

fauvel:


fauvel:
@njam what do you think of keeping the "disable-email-verification" setting on the form field, so that we would disable it explicitly on the sign up form (and use the standard MX record verification as a fallback)? This way we could easily re-enable email verification in the sign up form again if we find a better provider.

library/CM/Mail/SendJob.php
```diff
@@ -3,9 +3,19 @@
...
-        CM_Service_Manager::getInstance()->getMailer()->send($params->getMailMessage('message'));
-        if ($params->has('recipient') && $params->has('mailType')) {
-            $recipient = $params->getUser('recipient');
+        $message = $params->getMailMessage('message');
+        $recipient = $params->has('recipient') ? $params->getUser('recipient') : null;
+        if (!$recipient || !$recipient->getEmailVerified()) {
+            /** @var CM_Service_EmailVerification_ClientInterface $emailVerificationClient */
+            $emailVerificationClient = CM_Service_Manager::getInstance()->get('email-verification', CM_Service_EmailVerification_ClientInterface::class);
+            foreach ($message->getTo() as $address => $name) {
+                if (!$emailVerificationClient->isValid($address)) {
```
njam: How about adding a logger entry here so we know how often it's happening and for debugging? Could be level "info".
Please remember to document it here: https://cargomedia.atlassian.net/wiki/spaces/FB/pages/5144736/Logs

library/CM/Mail/SendJob.php
```diff
@@ -3,9 +3,19 @@
...
+                    return;
+                }
+            }
+        }
+        CM_Service_Manager::getInstance()->getMailer()->send($message);
+        if ($recipient && $params->has('mailType')) {
```
njam: While you're at it, let's move all params extraction to the beginning of the function for easier readability ($mailType)?

njam:
As discussed sgtm, let's keep "disable-email-verification" but make it default to TRUE, or change to "enable-email-verification".

library/CM/Mail/SendJob.php
njam: please remember to document it

PR #2682, opened 2017-07-21
FB-1060 add csv parser() helper

fvovan:


library/CM/File/Csv.php
```diff
@@ -50,6 +50,39 @@ public function mergeHeader(array $header) {
...
+     * @param string|null $lineBreak
+     * @param string|null $delimeter
+     * @param string|null $enclosure
+     * @param string|null $escape
+     * @param int|null $linesToSkip
+     * @return array
+     * @throws CM_Exception_Invalid
+     */
+    public function parse($lineBreak = null, $delimeter = null, $enclosure = null, $escape = null, $linesToSkip = null) {
+        $lineBreak = $lineBreak !== null ? (string) $lineBreak : "\n";
```
fauvel: I would prefer `isset($lineBreak) ? :` for readability

library/CM/File/Csv.php
```diff
@@ -50,6 +50,39 @@ public function mergeHeader(array $header) {
...
+        $delimeter = $delimeter !== null ? (string) $delimeter : ',';
```
fauvel: Typo in `$delimiter`

library/CM/File/Csv.php
```diff
@@ -50,6 +50,39 @@ public function mergeHeader(array $header) {
...
+        $enclosure = $enclosure !== null ? (string) $enclosure : '"';
+        $escape = $escape !== null ? (string) $escape : "\\";
+        $linesToSkip = $linesToSkip !== null ? (int) $linesToSkip : 1;
```
fauvel: This should rather default to 0 imo

library/CM/File/Csv.php
```diff
@@ -50,6 +50,39 @@ public function mergeHeader(array $header) {
...
+
+        $fileContent = $this->read();
+        $lineList = explode($lineBreak, $fileContent);
+        if (empty($lineList)) {
+            throw new CM_Exception_Invalid('Empty or bad CSV content');
```
fauvel: This should only happen if the *delimiter* is an empty string

library/CM/File/Csv.php
```diff
@@ -50,6 +50,39 @@ public function mergeHeader(array $header) {
...
+        }
+        $result = [];
+        $i = 0;
+        foreach ($lineList as $line) {
+            $line = trim($line);
```
fauvel: Not sure about that... Spaces should be included imo.

library/CM/File/Csv.php
fvovan: It's not. please see `str_getcsv()` signature

library/CM/File/Csv.php
fvovan: There are several edge cases in real world CSV files, like last empty line (which ideally should not be included), trailing spaces (usually given by creating a file with excel), usually if somebody is interested in preserving space chars, he uses `enclosure` symbols to emphasize it.

library/CM/File/Csv.php
fauvel: I know, but still it cannot be the expected behavior that you have to state explicitly that you have no header when you haven't... It is pretty much error prone in my opinion.

library/CM/File/Csv.php
fauvel: What do you mean?

library/CM/File/Csv.php
fauvel: Well, okay...

library/CM/File/Csv.php
fvovan: ohh sorry. my bad. thought i copied it right

library/CM/File/Csv.php
fvovan: okay, let it be mandatory then

fauvel: I don't really like it... 0 should be the default imo. I agree it's good to have it as the first argument because in general we will just pass 1.

fvovan: I would stay with the mandatory param. In case we can not agree with the default value it's the best indicator for the meaning it explicitly

fauvel: Okay... so let's call the parameter "headerLinesToSkip" to make sure everyone understands at once what it is about.

library/CM/File/Csv.php
```diff
@@ -50,6 +50,47 @@ public function mergeHeader(array $header) {
...
+    public function parse($headerLinesToSkip, $lineBreak = null, $delimiter = null, $enclosure = null, $escape = null) {
+        $headerLinesToSkip = (int) $headerLinesToSkip;
+        $lineBreak = isset($lineBreak) ? (string) $lineBreak : "\n";
+        $delimiter = isset($delimiter) ? (string) $delimiter : ',';
+        $enclosure = isset($enclosure) ? (string) $enclosure : '"';
+        $escape = isset($escape) ? (string) $escape : "\\";
+
+        if (empty($lineBreak)) {
+            throw new CM_Exception_Invalid('Empty linebreak');
+        }
```
fauvel: |`line 77`|
|---|
Why not return an empty array? This might be a use case, and you have no way to prevent it from happening without reading the file before calling this method, which is odd.

PR #2684, opened 2017-07-26
FB-1114 Lazy loading image blob CM_Image_Image / Imagick

ppp0:


njam:
We *want* to parse the image in the constructor, to ensure it is a fully valid image.

njam:
Closeable?

PR #2685, opened 2017-08-03
FB-1162 Allow string/mongodb ids for getSite as key

tomaszdurka:


library/CM/Params.php
```diff
@@ -403,6 +403,9 @@ public function getLanguage($key, $default = null) {
      */
     public function getSite($key, $default = null) {
         $param = $this->_get($key, $default);
+        if (is_string($param) && CM_MongoDb_Client::isValidObjectId($param)) {
+            return (new CM_Site_SiteFactory())->getSiteById($param);
+        }
         if (ctype_digit($param) || is_int($param)) {
             return (new CM_Site_SiteFactory())->getSiteByType($param);
```
zazabe: is it still used?

PR #2687, opened 2017-08-07
FB-1166 Use a recursive serialization for exception metaInfo

zazabe:


library/CM/ExceptionHandling/SerializableException.php
```diff
@@ -81,16 +82,20 @@ public function getTraceAsString() {
...
         $this->line = $exception->getLine();
         $this->file = $exception->getFile();
         if ($exception instanceof CM_Exception) {
             $variableInspector = new CM_Debug_VariableInspector();
-            $this->metaInfo = Functional\map($exception->getMetaInfo(), function ($value) use ($variableInspector) {
-                return $variableInspector->getDebugInfo($value);
+            $this->metaInfo = Functional\map($exception->getMetaInfo(), function ($value) use ($variableInspector, $recursive) {
+                return $variableInspector->getDebugInfo($value, [
+                    'recursive' => $recursive,
+                ]);
```
tomaszdurka: Lets make it default / only behaviour.

